PR #4, opened 2017-02-11
Split Coupling

elevien:
Implementation of the split coupling between two `JumpProblem`s. Currently this only works for constant rates, but it should be possible to extend it. Note that this is independent of the aggregator, but other couplings may need to be associated with specific aggregators.

### DIFF
--- a/src/coupled_array.jl
+++ b/src/coupled_array.jl
@@ -1,0 +1,47 @@
+type CoupledArray{T,T2} <: AbstractArray{Float64,1}
+  u::T
+  u_control::T2
+  order::Bool
+end
+
+Base.length(A::CoupledArray) = length(A.u) + length(A.u_control)
+Base.size(A::CoupledArray) = (length(A),)
+function Base.getindex(A::CoupledArray,i::Int)
+  if A.order == true
+    i <= length(A.u) ? A.u[i] : A.u_control[i-length(A.u)]
+  else
+    i <= length(A.u) ? A.u_control[i] : A.u[i-length(A.u)]
+  end
+end
+
+function Base.getindex(A::CoupledArray,I...)
+  A[sub2ind(A.u,I...)]
+end
+
+function Base.getindex(A::CoupledArray,I::CartesianIndex{1})
+  A[I[1]]
+end
+
+Base.setindex!(A::CoupledArray,v,I...) = (A[sub2ind(A.u,I...)] = v)
+Base.setindex!(A::CoupledArray,v,I::CartesianIndex{1}) = (A[I[1]] = v)
+function Base.setindex!(A::CoupledArray,v,i::Int)
+  if A.order == true
+    i <= length(A.u) ? (A.u[i] = v) : (A.u_control[i-length(A.u)] = v)
+  else
+    i <= length(A.u) ? (A.u_control[i] = v) : (A.u[i-length(A.u)] = v)
+  end
+end
+
+linearindexing{T<:CoupledArray}(::Type{T}) = Base.LinearFast()
+similar(A::CoupledArray) = deepcopy(A)
+
+
+function recursivecopy!{T<:CoupledArray}(dest::T, src::T)
+  recursivecopy!(dest.u,src.u)
+  recursivecopy!(dest.u_control,src.u_control)
+end
+
+display(A::CoupledArray) = display(A.u)
+show(A::CoupledArray) = show(A.u)
+plot_indices(A::CoupledArray) = eachindex(A)
+flip_u!(A::CoupledArray) = (A.order = !A.order)
--- a/src/coupled_array.jl
+++ b/src/coupled_array.jl
@@ -39,6 +39,7 @@ similar(A::CoupledArray) = deepcopy(A)
 function recursivecopy!{T<:CoupledArray}(dest::T, src::T)
   recursivecopy!(dest.u,src.u)
   recursivecopy!(dest.u_control,src.u_control)
+  dest.order = src.order
 end
 
 display(A::CoupledArray) = display(A.u)
--- a/src/coupling.jl
+++ b/src/coupling.jl
@@ -1,15 +1,62 @@
 SplitCoupledJumpProblem(prob::AbstractJumpProblem,prob_control::AbstractJumpProblem,aggregator::AbstractAggregatorAlgorithm,coupling_map::Vector{Tuple{Int64,Int64}})= JumpProblem(cat_problems(prob,prob_control),aggregator,build_split_jumps(prob,prob_control,coupling_map)...)
 
+type CoupledArray{T,T2} <: AbstractArray{Float64,1}
+  u::T
+  u_control::T2
+  order::Bool
+end
+
+Base.length(A::CoupledArray) = length(A.u) + length(A.u_control)
+Base.size(A::CoupledArray) = (length(A),)
+function Base.getindex(A::CoupledArray,i::Int)
+  if A.order == true
+    i <= length(A.u) ? A.u[i] : A.u_control[i-length(A.u)]
+  else
+    i <= length(A.u) ? A.u_control[i] : A.u[i-length(A.u)]
+  end
+end
+function Base.getindex(A::CoupledArray,I...)
+  A[sub2ind(A.u,I...)]
+end
+function Base.getindex(A::CoupledArray,I::CartesianIndex{1})
+  A[I[1]]
+end
+Base.setindex!(A::CoupledArray,v,I...) = (A[sub2ind(A.u,I...)] = v)
+Base.setindex!(A::CoupledArray,v,I::CartesianIndex{1}) = (A[I[1]] = v)
+function Base.setindex!(A::CoupledArray,v,i::Int)
+  if A.order == true
+    i <= length(A.u) ? (A.u[i] = v) : (A.u_control[i-length(A.u)] = v)
+  else
+    i <= length(A.u) ? (A.u_control[i] = v) : (A.u[i-length(A.u)] = v)
+  end
+end
+
+linearindexing{T<:CoupledArray}(::Type{T}) = Base.LinearFast()
+similar(A::CoupledArray) = deepcopy(A)
+
+
+function recursivecopy!{T<:CoupledArray}(dest::T, src::T)
+  recursivecopy!(dest.u,src.u)
+  recursivecopy!(dest.u_control,src.u_control)
+end
+
+display(A::CoupledArray) = display(A.u)
+show(A::CoupledArray) = show(A.u)
+plot_indices(A::CoupledArray) = eachindex(A.u)
+flip_u!(A::CoupledArray) = (A.order = !A.order)
+
+
 # make new discrete problem by joining initial_data
 function cat_problems(prob::AbstractJumpProblem,prob_control::AbstractJumpProblem)
-  coupled_u0 = cat(1,prob.prob.u0,prob_control.prob.u0)
-  DiscreteProblem(coupled_u0,prob.prob.tspan)
+  u0_coupled = CoupledArray(prob.prob.u0,prob_control.prob.u0,true)
+  DiscreteProblem(u0_coupled,prob.prob.tspan)
 end
 
 function build_split_jumps(prob::AbstractJumpProblem,prob_control::AbstractJumpProblem,coupling_map::Vector{Tuple{Int64,Int64}})
   num_jumps = length(prob.discrete_jump_aggregation.rates)
   d = length(prob.prob.u0)
   jumps = []
+  # overallocates, will fix later
   uncoupled = deleteat!(Vector(1:num_jumps),[c[1] for c in coupling_map ])
   uncoupled_control = deleteat!(Vector(1:num_jumps),[c[2] for c in coupling_map ])
   for c in uncoupled   # make uncoupled jumps in prob
@@ -22,9 +69,9 @@ function build_split_jumps(prob::AbstractJumpProblem,prob_control::AbstractJumpP
     new_rate = (t,u)->rate(t,u[d+1:2*d])
     affect! = prob_control.discrete_jump_aggregation.affects![c]
     new_affect! = function (integrator)
-        flip_u!(integrator)
+        flip_u!(integrator.u)
         affect!(integrator)
-        flip_u!(integrator)
+        flip_u!(integrator.u)
     end
     append!(jumps,[ConstantRateJump(new_rate,new_affect!)])
   end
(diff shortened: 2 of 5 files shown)


src/coupling.jl
```diff
@@ -0,0 +1,68 @@
...
+function cat_problems(prob::AbstractJumpProblem,prob_control::AbstractJumpProblem)
+  coupled_u0 = cat(1,prob.prob.u0,prob_control.prob.u0)
+  DiscreteProblem(coupled_u0,prob.prob.tspan)
+end
+
+function build_split_jumps(prob::AbstractJumpProblem,prob_control::AbstractJumpProblem,coupling_map::Vector{Tuple{Int64,Int64}})
+  num_jumps = length(prob.discrete_jump_aggregation.rates)
+  d = length(prob.prob.u0)
+  jumps = []
+  uncoupled = deleteat!(Vector(1:num_jumps),[c[1] for c in coupling_map ])
```
ChrisRackauckas: You shouldn't have to create the whole vector and deleteat. This is likely much worse than building by pushing (which is automatically amortized), especially since the comprehension forms an extra vector here as well. But there's likely a way to do it directly.

ChrisRackauckas: This probably doesn't need to be max performance since it's just building a problem, so at least adding a comment that it overallocates should be fine unless you want to fix it now.

src/coupling.jl
```diff
@@ -0,0 +1,68 @@
...
+  uncoupled_control = deleteat!(Vector(1:num_jumps),[c[2] for c in coupling_map ])
+  for c in uncoupled   # make uncoupled jumps in prob
+    new_rate = prob.discrete_jump_aggregation.rates[c]
+    new_affect! = prob.discrete_jump_aggregation.affects![c]
+    append!(jumps,[ConstantRateJump(new_rate,new_affect!)])
+  end
+  for c in uncoupled_control  # make uncoupled jumps in prob_control
+    rate = prob_control.discrete_jump_aggregation.rates[c]
+    new_rate = (t,u)->rate(t,u[d+1:2*d])
```
ChrisRackauckas: This should use a view.

src/coupling.jl
```diff
@@ -0,0 +1,68 @@
...
+        affect!(integrator)
+        flip_u!(integrator)
+        affect_control!(integrator)
+        flip_u!(integrator)
+    end
+    new_rate = (t,u)->min(rate(t,u),rate_control(t,u[d+1:2*d]))
+    push!(jumps,ConstantRateJump(new_rate,new_affect!))
+     # only prob
+    new_affect! = affect!
+    new_rate = (t,u)->rate(t,u)-min(rate(t,u),rate_control(t,u[d+1:2*d]))
```
ChrisRackauckas: views here

src/coupling.jl
```diff
@@ -0,0 +1,68 @@
...
+    push!(jumps,ConstantRateJump(new_rate,new_affect!))
+    # only prob_control
+    new_affect! = function (integrator)
+        flip_u!(integrator)
+        affect!(integrator)
+        flip_u!(integrator)
+    end
+    new_rate = (t,u)->rate_control(t,u[d+1:2*d])-min(rate(t,u),rate_control(t,u[d+1:2*d]))
```
ChrisRackauckas: view here

src/coupling.jl
```diff
@@ -0,0 +1,68 @@
...
+  end
+  jumps
+end
+
+# flip positions of u in integrator so that affects can be applied
+# is there a better way to build new affects from old ones?
+flip_u! =  function (integrator)
+  d = length(integrator.u)
+  halfd = Int(d//2)
+  integrator.u[1:halfd],integrator.u[halfd+1:d] = integrator.u[halfd+1:d],integrator.u[1:halfd]
```
ChrisRackauckas: views

src/coupling.jl
```diff
@@ -0,0 +1,68 @@
+SplitCoupledJumpProblem(prob::AbstractJumpProblem,prob_control::AbstractJumpProblem,aggregator::AbstractAggregatorAlgorithm,coupling_map::Vector{Tuple{Int64,Int64}})= JumpProblem(cat_problems(prob,prob_control),aggregator,build_split_jumps(prob,prob_control,coupling_map)...)
+
+# make new discrete problem by joining initial_data
+function cat_problems(prob::AbstractJumpProblem,prob_control::AbstractJumpProblem)
+  coupled_u0 = cat(1,prob.prob.u0,prob_control.prob.u0)
```
ChrisRackauckas: This form of concatenation can change the array type. Also, it can be costly. It's probably better to use a CatView here, or a new array type.

src/coupling.jl
```diff
@@ -0,0 +1,68 @@
...
+  DiscreteProblem(coupled_u0,prob.prob.tspan)
```
ChrisRackauckas: Does this need to be restrained to `DiscreteProblem`?

ChrisRackauckas: This is the paper, right? https://arxiv.org/pdf/1403.3127.pdf

Is there any think that makes it so that way constant rate jumps in an ODE/SDE would not have the correct accuracy? I don't know this well enough.

ChrisRackauckas:
This kind of thing seems like it's very wasteful because it's using a contiguous array and having to repeatedly swap components. Instead, I think you can remove all of the swapping by generating a new `AbstractArray` which is similar to the `ExtendedJumpArray`, i.e. it has a linear index through two fields corresponding to the different problems, except have a separate field for "which is first" that is a flag. "Flipping" the indexing would then just be a switch of this flag. That would probably get rid of all of the slice/view issues in the rate/affect calculations.

elevien:
>  Instead, I think you can remove all of the swapping by generating a new AbstractArray which is similar to the ExtendedJumpArray

Yeah, good idea. I'm looking at your other code and this definitely seems like the way to do it, so I'll give it a try.

src/coupling.jl
elevien: > Does this need to be restrained to DiscreteProblem?

I don't think so, but I haven't experimented with that yet. 

> Is there any think that makes it so that way constant rate jumps in an ODE/SDE would not have the correct accuracy? I don't know this well enough.

If I understand what you are asking, then no. The coupling shouldn't really have a significant impact on the accuracy since the jump rates of the coupled process are comparable to the original process.

src/coupling.jl
elevien: I'll leave a comment and get everything else working first.

elevien:
I still need to implement `SplitCoupledJumpProblem` for ODE/SDE problems and come up with some better tests.  Also, maybe `CoupledArray` shouldn't be implemented here? I expect it can used for SDE/SPDE specific couplings as well.

ChrisRackauckas:
>Also, maybe CoupledArray shouldn't be implemented here? I expect it can used for SDE/SPDE specific couplings as well.

I'll accept it here and likely move it to DiffEqBase later.

ChrisRackauckas:
>I still need to implement SplitCoupledJumpProblem for ODE/SDE problems and come up with some better tests.

One way to handle this is just making a few dispatches.

https://github.com/JuliaDiffEq/DiffEqJump.jl/blob/master/src/problem.jl#L42

elevien:
> I'll accept it here and likely move it to DiffEqBase later.

I put it in a separate file. Hopefully, that will make things easier when you move it.

elevien:
I didn't implement the DDE and DAE couplings since I don't really know much about those, but I could easily add them. At least for DDE problems, I don't see why it wouldn't work out the same.

src/coupled_array.jl
```diff
@@ -0,0 +1,47 @@
...
+end
+
+linearindexing{T<:CoupledArray}(::Type{T}) = Base.LinearFast()
+similar(A::CoupledArray) = deepcopy(A)
+
+
+function recursivecopy!{T<:CoupledArray}(dest::T, src::T)
+  recursivecopy!(dest.u,src.u)
+  recursivecopy!(dest.u_control,src.u_control)
+end
```
ChrisRackauckas: Won't this need to set the order on the destination as well?

src/coupled_array.jl
```diff
@@ -0,0 +1,47 @@
...
+Base.size(A::CoupledArray) = (length(A),)
+function Base.getindex(A::CoupledArray,i::Int)
+  if A.order == true
+    i <= length(A.u) ? A.u[i] : A.u_control[i-length(A.u)]
+  else
+    i <= length(A.u) ? A.u_control[i] : A.u[i-length(A.u)]
+  end
+end
+
+function Base.getindex(A::CoupledArray,I...)
```
ChrisRackauckas: I like this way of handling higher dimensional indices for this.

src/coupled_array.jl
```diff
@@ -0,0 +1,47 @@
...
+  A[sub2ind(A.u,I...)]
+end
+
+function Base.getindex(A::CoupledArray,I::CartesianIndex{1})
```
ChrisRackauckas: Is this needed?

ChrisRackauckas:
>I didn't implement the DDE and DAE couplings since I don't really know much about those, but I could easily add them. At least for DDE problems, I don't see why it wouldn't work out the same.

I'll add them after merging. DDE problems, it'll work just fine. DAE problems, I'm not quite sure, but I'll do it the same way I had jump problems and leave a comment in there again. There's no solver for DAEs with the full callback interface to test it, so it's a little in the dark there right now. But since I never encounter DAEs in mathematical biology...

src/coupled_array.jl
elevien: Not that I'm aware of, but it was implemented for `ExtendedJumpArray` so I put it in there to be safe.

src/coupled_array.jl
elevien: Yeah, I missed that.

src/coupled_array.jl
elevien: Actually, it is needed:
`Got an exception of type LoadError outside of a @test
  LoadError: MethodError: no method matching sub2ind(::Tuple{Base.OneTo{Int64}}, ::CartesianIndex{1})
  Closest candidates are:
    sub2ind(::AbstractArray{T,N}, ::Any...) at abstractarray.jl:1419
    sub2ind(::Tuple{Base.OneTo}, ::Integer) at abstractarray.jl:1439
    sub2ind(::Tuple{Base.OneTo}, ::Integer...) at abstractarray.jl:1438
    ...
   in macro expansion at ./broadcast.jl:127 [inlined]
   in macro expansion at ./simdloop.jl:73 [inlined]
   in macro expansion at ./broadcast.jl:123 [inlined]
   in _broadcast!(::Base.#abs, ::Array{Float64,1}, ::Tuple{Tuple{Bool}}, ::Tuple{Tuple{Int64}}, ::Tuple{DiffEqJump.CoupledArray{Array{Float64,1},Array{Float64,1}}}, ::Type{Val{1}}) at ./broadcast.jl:117
   in broadcast! at ./broadcast.jl:172 [inlined]
   in broadcast_t at ./broadcast.jl:228 [inlined]
   in broadcast at ./broadcast.jl:230 [inlined]
   in ode_determine_initdt`

ChrisRackauckas: Ahh, okay. It's a broadcast thing.

src/coupling.jl
```diff
@@ -0,0 +1,134 @@
...
+  l = length(prob.u0) # add l_c = length(prob_control.u0)
+  new_f = function (t,u,du)
+    prob.f(t,u.u,@view du[1:l])
+    prob_control.f(t,u.u_control,@view du[l+1:2*l])
+  end
+  u0_coupled = CoupledArray(prob.u0,prob_control.u0,true)
+  ODEProblem(new_f,u0_coupled,prob.tspan)
+end
+
+function cat_problems(prob::DiscreteProblem,prob_control::AbstractODEProblem)
```
ChrisRackauckas: Mixing these only makes sense if some non-defaults in the solver are on, otherwise a non-trivial `f` will act differently when changed to an `ODEProblem` (from a function map to an ODE RHS). This may cause problems. We may just want to disallow this.

src/coupling.jl
```diff
@@ -0,0 +1,134 @@
...
+    prob.g(t,u.u,@view du[1:l])
+    for i in l+1:2*l
+      du[i] = 0.
+    end
+  end
+  u0_coupled = CoupledArray(prob.u0,prob_control.u0,true)
+  SDEProblem(new_f,new_g,u0_coupled,prob.tspan)
+end
+
+function cat_problems(prob::AbstractSDEProblem,prob_control::DiscreteProblem)
```
ChrisRackauckas: Same as the Discrete+ODE

src/coupling.jl
ChrisRackauckas: Or warn if the `f` in the `DiscreteProblem` is not trivial.

elevien: At least for me, it will be useful to mix these. What is the best way to check if `f` is trivial? Right now I just discard the `f` from `DiscreteProblem` and make a new one.

ChrisRackauckas: The anonymous functions used for the defaults in DiffEqBase should be set as constants, and you could then type-check against them.

https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/problems/discrete_problems.jl#L25

elevien: Oh, I see. I wasn't sure how to type-check anonymous functions. So you want me to put in a PR in DiffEqBase for this?

ChrisRackauckas: Yeah sure. It should just be like 4 lines. (Sorry I've been away so I haven't been able to help too much!). Every function (including anonymous functions) are its own type, so if you just

```julia
const DISCRETE_INPLACE_DEFAULT = (t,u,du) -> ...
```

then you can just

```julia
if typeof(prob.f) <: typeof(DiffEqBase.DISCRETE_INPLACE_DEFAULT)
  ...
else
  # warn
end
```

You may be able to just check `prob.f === DiffEqBase....` and have that compile away.

src/coupling.jl
```diff
@@ -18,10 +18,11 @@ end
 
 function cat_problems(prob::DiscreteProblem,prob_control::AbstractODEProblem)
   l = length(prob.u0) # add l_c = length(prob_control.u0)
+  if !(typeof(prob.f) <: typeof(DiffEqBase.DISCRETE_INPLACE_DEFAULT))
+    warn("Coupling to DiscreteProblem with nontrivial f.")
```
ChrisRackauckas: Something like `""Coupling to DiscreteProblem with nontrivial f. Note that, unless the `f` was intended to be solved with `scale_by_time=true`, then the meaning of `f` will change when using an ODE/SDE/DDE/DAE solver."

elevien:
Do I need to reconfigure something for it to pass the tests? It looks like it's using DiffEqBase v0.11.0 not master.

ChrisRackauckas:
It's okay if tests fail right now because of that change.

ChrisRackauckas:
Thanks a lot! This looks like really good stuff.

PR #18, opened 2018-04-01
Function wrapped rates

isaacsas:
This is a first attempt to switch from storing rates and affects! in tuples to storing them in vectors of function wrappers (for systems with enough jumps). A bit of testing on a diffusion example suggested that at < 10 jumps tuples are faster, but at 10 jumps they are noticeably slower than using functionwrappers. The new `Direct()` aggregator therefore switches to a function wrapper representation at 10 or more jumps. 

I also included a new test script that runs a system with a bunch of first order reactions (so that the mean is known exactly). It shows that, at least for this example, everything seems to give the right mean.

Note, this PR does not update the recursive `JumpSet` construction code to avoid tuples; at some point this will probably be necessary for performance reasons when considering systems with many many jumps.


src/aggregators/direct.jl
```diff
@@ -55,14 +57,54 @@ end
...
+      new_rate     = rates[i](u, p, t)
+      cur_rates[i] = new_rate + prev_rate
+      prev_rate    = cur_rates[i]
+  end
+  @inbounds sum_rate = cur_rates[end]
+  sum_rate,randexp()/sum_rate
+end
+##################################################################
+
+function getRatesAffectsAsTuples(constant_jumps)
```
ChrisRackauckas: Camel case is against stand Julia style.

src/aggregators/direct.jl
```diff
@@ -55,14 +57,54 @@ end
...
+  rates         = [RateWrapper(c.rate) for c in constant_jumps]
+  AffectWrapper = FunctionWrappers.FunctionWrapper{Void,Tuple{Any}}
+  affects!      = [AffectWrapper(x->(c.affect!(x);nothing)) for c in constant_jumps]
+
+  return rates, affects!
+end
+
+
+# when to switch from tuples for rates/affects to FunctionWrappers
+const TUPLE_TO_FWRAPPER_CUTOFF = 10
```
ChrisRackauckas: It feels weird to try and just make this built in. I think that instead it can just be a new aggregator, but with a lot of common code. Which one to use is problem dependent. @alanderos91 seems to have ideas for generalizing the code re-use.

test/functwrapper_tests.jl
```diff
@@ -0,0 +1,53 @@
+# N stochastic A->B reactions at different rates
+using DiffEqJump, DiffEqBase, OrdinaryDiffEq
+using Base.Test
+
+doPrint = false
+
+function testAtoBMean(N)
```
ChrisRackauckas: camel case

test/functwrapper_tests.jl
```diff
@@ -0,0 +1,53 @@
+# N stochastic A->B reactions at different rates
+using DiffEqJump, DiffEqBase, OrdinaryDiffEq
+using Base.Test
+
+doPrint = false
```
ChrisRackauckas: camel case

src/aggregators/direct.jl
isaacsas: Ok, I'm to remove this PR then. Once @alanderos91 finishes revising the interface I can look into sending in updates to enable larger SSAs.

PR #20, opened 2018-04-04
AbstractSSAJumpAggregator interface

alanderos91:
I think everything we discussed is in place. Please feel free to suggest alternative names for the various functions and fields added.

### DIFF
--- a/src/aggregators/direct.jl
+++ b/src/aggregators/direct.jl
@@ -38,10 +38,10 @@ end
 function aggregate(aggregator::Direct, u, p, t, end_time, constant_jumps, save_positions, rng)
   rates = ((c.rate for c in constant_jumps)...)
   affects! = ((c.affect! for c in constant_jumps)...)
-  cur_rates = Vector{Float64}(length(rates))
-  sum_rate = zero(Float64)
+  cur_rates = Vector{typeof(t)}(length(rates))
+  sum_rate = zero(typeof(t))
   next_jump = 0
-  next_jump_time = typemax(Float64)
+  next_jump_time = typemax(typeof(t))
   DirectJumpAggregation(next_jump, next_jump_time, end_time, cur_rates,
     sum_rate, rates, affects!, save_positions, rng)
 end
--- a/src/aggregators/direct.jl
+++ b/src/aggregators/direct.jl
@@ -73,8 +73,7 @@ end
   cur_rates = p.cur_rates
   rates = p.rates
 
-  @inbounds fill_cur_rates(u, params, t, cur_rates, 1, rates...)
-  @inbounds cur_rates[1] = cur_rates[1]
+  fill_cur_rates(u, params, t, cur_rates, 1, rates...)
   @inbounds for i in 2:length(cur_rates) # normalize for choice, cumsum
     cur_rates[i] = cur_rates[i] + cur_rates[i-1]
   end
--- a/src/aggregators/direct.jl
+++ b/src/aggregators/direct.jl
@@ -1,4 +1,4 @@
-type DirectJumpAggregation{T,F1,F2} <: SSAJumpAggregator
+type DirectJumpAggregation{T,F1,F2,RNG} <: SSAJumpAggregator
   next_jump::Int
   next_jump_time::T
   end_time::T
@@ -7,6 +7,7 @@ type DirectJumpAggregation{T,F1,F2} <: SSAJumpAggregator
   rates::F1
   affects!::F2
   save_positions::Tuple{Bool,Bool}
+  rng::RNG
 end
 
 @inline function (p::DirectJumpAggregation)(u,t,integrator) # condition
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
@isaacsas wanna take a look?

src/problem.jl
```diff
@@ -19,15 +19,16 @@ JumpProblem(prob,aggregator::AbstractAggregatorAlgorithm,jumps::AbstractJump...;
 JumpProblem(prob,jumps::JumpSet;kwargs...) = JumpProblem(prob,NullAggregator(),jumps;kwargs...)
 
 function JumpProblem(prob,aggregator::AbstractAggregatorAlgorithm,jumps::JumpSet;
-                     save_positions = typeof(prob) <: AbstractDiscreteProblem ? (false,true) : (true,true))
+                     save_positions = typeof(prob) <: AbstractDiscreteProblem ? (false,true) : (true,true),
+                     rng = Xorshifts.Xoroshiro128Star(rand(UInt64)))
```
ChrisRackauckas: Let's use Xoroshiro128Plus. See the end of the chat which states it's not the same as the one in the post.

ChrisRackauckas:
Forgot to define the abstract type

src/problem.jl
isaacsas: Just to clarify. The blog post found issues with Xoroshiro128Plus. It recommended as fine XorShift* 128/64. I haven't found any info about Xoroshiro128star.

ChrisRackauckas: Oh really? But then it passed all of the tests of the repo?

src/aggregators/direct.jl
```diff
@@ -1,68 +1,94 @@
...
 
-DiscreteCallback(c::DirectJumpAggregation) = DiscreteCallback(c,c,initialize=c,save_positions=c.save_positions)
+######################## SSA specific helper routines ########################
+
+@fastmath function time_to_next_jump(p::DirectJumpAggregation, u, params, t)
+  cur_rates = p.cur_rates
+  rates = p.rates
+
+  @inbounds fill_cur_rates(u, params, t, cur_rates, 1, rates...)
+  @inbounds cur_rates[1] = cur_rates[1]
```
isaacsas: Is this necessary?

src/aggregators/direct.jl
```diff
@@ -1,68 +1,94 @@
...
-  @fastmath normalizer = 1/sum_rate
-  @inbounds cur_rates[1] = normalizer*cur_rates[1]
-  @inbounds for i in 2:length(cur_rates) # normalize for choice, cumsum
-    cur_rates[i] = normalizer*cur_rates[i] + cur_rates[i-1]
-  end
-  sum_rate,randexp()/sum_rate
+############################# Required Functions #############################
+
+# creating the JumpAggregation structure
+@inline function aggregate(aggregator::Direct, u, p, t, end_time, constant_jumps, save_positions, rng)
```
isaacsas: The inline here also seems unnecessary (it was present before but I don't see what good it can do here).

ChrisRackauckas: Yeah it's probably unnecessary.

src/problem.jl
isaacsas: It failed a different series of tests (newer I think) called PractRand. I'm not an expert on these tests, so don't know the significance. You can see the details on the blog post I linked back in the issue thread.

isaacsas:
It looks good to me. This is the basic interface we discussed in the issue thread. I'll send through the mass action updates once this is in.

src/aggregators/direct.jl
```diff
@@ -1,68 +1,93 @@
...
-    cur_rates[i] = normalizer*cur_rates[i] + cur_rates[i-1]
-  end
-  sum_rate,randexp()/sum_rate
+############################# Required Functions #############################
+
+# creating the JumpAggregation structure
+function aggregate(aggregator::Direct, u, p, t, end_time, constant_jumps, save_positions, rng)
+  rates = ((c.rate for c in constant_jumps)...)
+  affects! = ((c.affect! for c in constant_jumps)...)
+  cur_rates = Vector{Float64}(length(rates))
```
isaacsas: Maybe instead of using Float64's here you should use `typeof(t)`?

alanderos91: Sure. I'll hold off on pushing so I don't trigger so many builds...

src/problem.jl
alanderos91: I don't know enough to comment on this but can change as needed.

src/aggregators/direct.jl
isaacsas: Yeah, sorry for the intermittent stream of comments there. I'm still getting used to looking at PRs on Github... That should be it from me -- I've gone through everything now.

ChrisRackauckas: Don't worry about triggering builds. It'll queue.

ChrisRackauckas: Note you can also add [ci-skip] to make it not queue a build for a commit.

alanderos91:
Fantastic. Many thanks to both you for all the help! 🎉

ChrisRackauckas:
Many thanks to you!

PR #22, opened 2018-04-05
add mass action jump type

isaacsas:
OK, here is a first try adding in a new `MassActionJump`. I've also added a new test script that verifies for a linear reaction system that the SSAs get the right mean using each of
1. tuple-based constant jumps
2. function wrapper-based constant jumps
3. a mass action jump
4. a hybrid of tuples and mass action 
5. a hybrid of function wrappers and mass action

Once this is fixed up and approved I'll work on a couple more test scripts with nonlinear reactions. It would be good to have a database of a few examples before starting on new SSAs algorithms.

EDIT: I called the new function wrapper aggregator algorithm `DirectManyJumps`. I'm happy to change that if someone has a better name...

### DIFF
--- a/src/jumps.jl
+++ b/src/jumps.jl
@@ -35,29 +35,86 @@ end
 RegularJump(rate,c,dc;mark_dist = nothing,constant_c = false) =
             RegularJump(rate,c,dc,mark_dist,constant_c)
 
-struct JumpSet{T1,T2,T3} <: AbstractJump
+
+struct MassActionJump{T,S} <: AbstractJump
+  scaled_rates::T
+  reactant_stoch::S
+  net_stoch::S
+
+  function MassActionJump{T,S}(rates::T, rs::S, ns::S, scale_rates::Bool) where {T,S}
+    sr = copy(rates)
+    if scale_rates && !isempty(sr) 
+      scalerates!(sr, rs)
+    end
+    new(sr, rs, ns)
+  end
+end
+MassActionJump(usr::T, rs::S, ns::S; scale_rates = true ) where {T,S} = MassActionJump{T,S}(usr, rs, ns, scale_rates)
+
+
+struct JumpSet{T1,T2,T3,T4} <: AbstractJump
   variable_jumps::T1
   constant_jumps::T2
   regular_jump::T3
+  massaction_jump::T4
 end
 
-JumpSet(jump::ConstantRateJump) = JumpSet((),(jump,),nothing)
-JumpSet(jump::VariableRateJump) = JumpSet((jump,),(),nothing)
-JumpSet(jump::RegularJump) = JumpSet((),(),jump)
-JumpSet() = JumpSet((),(),nothing)
+JumpSet(jump::ConstantRateJump) = JumpSet((),(jump,),nothing,nothing)
+JumpSet(jump::VariableRateJump) = JumpSet((jump,),(),nothing,nothing)
+JumpSet(jump::RegularJump)      = JumpSet((),(),jump,nothing)
+JumpSet(jump::MassActionJump)   = JumpSet((),(),nothing,jump)
+JumpSet() = JumpSet((),(),nothing,nothing)
 JumpSet(jb::Void) = JumpSet()
 
 # For Varargs, use recursion to make it type-stable
 
-JumpSet(jumps::AbstractJump...) = JumpSet(split_jumps((), (), nothing, jumps...)...)
+JumpSet(jumps::AbstractJump...) = JumpSet(split_jumps((), (), nothing, nothing, jumps...)...)
 
-@inline split_jumps(vj, cj, rj) = vj, cj, rj
-@inline split_jumps(vj, cj, rj, v::VariableRateJump, args...) = split_jumps((vj..., v), cj, rj, args...)
-@inline split_jumps(vj, cj, rj, c::ConstantRateJump, args...) = split_jumps(vj, (cj..., c), rj, args...)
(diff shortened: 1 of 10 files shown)


src/jumps.jl
```diff
@@ -35,29 +35,86 @@ end
...
+JumpSet(jump::VariableRateJump) = JumpSet((jump,),(),nothing,nothing)
+JumpSet(jump::RegularJump)      = JumpSet((),(),jump,nothing)
+JumpSet(jump::MassActionJump)   = JumpSet((),(),nothing,jump)
+JumpSet() = JumpSet((),(),nothing,nothing)
 JumpSet(jb::Void) = JumpSet()
 
 # For Varargs, use recursion to make it type-stable
 
-JumpSet(jumps::AbstractJump...) = JumpSet(split_jumps((), (), nothing, jumps...)...)
+JumpSet(jumps::AbstractJump...) = JumpSet(split_jumps((), (), nothing, nothing, jumps...)...)
```
ChrisRackauckas: I have a feeling that building this with a ton of jumps will be terrible for performance. One of the next things we should look at is allowing vectors of jumps as input.

ChrisRackauckas:
I really like this, even more than I thought I would. Basically it adds a specialization for mass-action terms, which even when you have lots of special regulation terms are bound to be the majority of the rates and thus this should be a huge speed improvement without sacrificing the flexibility. I thought it would be a little bit difficult to combine them, but really you just use the length and it works out pretty simply, so cool!

I am not a fan of the name. While not very descriptive, `DirectFW` would be my choice. 

There must be a way to get rid of the `Vector{Vector{Pair}}` structure. Can the inner vector be a StaticVector? We would have to limit the number of reactants, but could that be a good idea? I'm curious what happens to the indexing performance. The memory usage will definitely be better, but that's not my concern. My concern is really the looped small vector dereference performance. 

>It would be good to have a database of a few examples before starting on new SSAs algorithms.

We should get them added to DiffEqProblemLibrary.jl and start adding some SSA stuff to DiffEqBenchmarks.jl.

src/jumps.jl
isaacsas: Yup. Not an issue for the `MassActionJump`, but it would be an issue for `ConstantJumps` or `VariableRateJumps`. It does seem that everything works if one just directly creates a `JumpSet` that stores `ConstantJumps` as a vector, and then passes that into the `JumpProblem`. I do that in the new test script I added.

isaacsas:
The name isn't great.. For some reason nothing good came to mind. I can change it to `DirectFW`. The only reason I hadn't gone with such a name is that it works just fine with no constant jumps, and isn't using function wrappers for mass action jumps. 

I've also tried vectors of matrices with two rows, and vectors of static matrices with two rows. What I liked about pairs is that it offers the opportunity to have different types for the species id and the stochiometric coefficient. Though maybe there aren't any use cases where someone would want different types for the two... When the two components are both the same type, if I'm understanding `sizeof` correctly, it does look like the inner `Vector{Pair}` gets laid out as one contiguous chunk of memory. A named `Pair` seems like it would be even better, but I didn't try using a custom type...

For whatever reason `Vector{SMatrix}`, where the static matrices all had two rows but varying numbers of columns, didn't give good performance and used way more memory. It was on an older version of the code, so maybe I wasn't doing something the right way though... I could check again. At the time I assumed that the `SMatrix` was getting copied over when passed into `evalRxRate!`, and that this was more expensive than looking up the `Vector{Pairs}`. I didn't try passing the whole `Vector{SMatrix}` into `evalRxRate!`, and then indexing the specific reaction `SMatrix` there. Maybe that would avoid any copying issues.

ChrisRackauckas:
>if I'm understanding sizeof correctly, it does look like the inner Vector{Pair} gets laid out as one contiguous chunk of memory. A named Pair seems like it would be even better, but I didn't try using a custom type...

Yes, `Pair` is a `struct` of `isbits` types, so it's `isbits` as well and will pack well. I think using `Pair` is fine here.

>For whatever reason Vector{SMatrix}, where the static matrices all had two rows but varying numbers of columns, didn't give good performance and used way more memory. 

They all have to have the same number of columns for this to be type-stable.

isaacsas:
> They all have to have the same number of columns for this to be type-stable.

That makes sense. I do remember the type signature being confusing to me. I'm slowly getting the hang of recognizing and dealing with type instabilities but that is one place my lack of Julia experience shows... If we wanted to try static matrices again it would probably be fine to specialize to at most two or three species for reactants, and at most 4-6 for net stochiometry.

It would be very easy down the line to swap around what is used in the `MassActionJump`, or add support for other stoichiometry representations. The type is generic, and determined by what the user passes in. The only real requirement I explicitly put on the fields is that the two stochiometry fields have the same type. To change this around just requires modifying the functions in `massaction_rates.jl`.

isaacsas:
@alanderos91 Did you have something in mind for specializing `build_jump_aggregation`?

alanderos91:
@isaacsas Honestly, no. I can't see a clean way to do this given the current function signature. Might be best to hold off until the new SSAs.

isaacsas:
> Am I correct in thinking this is a sparse representation of reaction stoichiometry?

@alanderos91 That is exactly right; it is just a different representation than using a sparse matrix. The `k`th entry in the vector is itself a vector of pairs that encodes the stoichiometry for the reaction. Each pair maps the id of a species to a stoichiometric coefficient.  So it replaces the (row=species, col=rx, value=stoichiometry coefficient) one might have in a sparse matrix representation.

PR #24, opened 2018-04-07
Optimized SSA support and FRM

isaacsas:
1. To enable more general SSA support, kwargs are passed down from `JumpProblem` into `aggregate`. Along with this I made `build_jump_aggregation` a generic routine that should handle setting up all the required universal SSA fields. A `JumpAggregation` outer constructor then accepts kwargs and can use them to require additional SSA specific information (like dependency graphs). This should allow users to make their own SSAs / `JumpAggregations`, possibly requiring additional info, outside of `DiffEqJump` and still have them work through `JumpProblem`.
2. Added first reaction method implementations (FRM and FRMFW). 
3. Renamed `DirectManyJumps` to `DirectFW`.

### DIFF
--- a/src/DiffEqJump.jl
+++ b/src/DiffEqJump.jl
@@ -18,6 +18,7 @@ include("massaction_rates.jl")
 include("aggregators/aggregators.jl")
 include("aggregators/ssajump.jl")
 include("aggregators/direct.jl")
+include("aggregators/frm.jl")
 include("jumps.jl")
 include("problem.jl")
 include("callbacks.jl")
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
Looks great! Can you contribute some docs?

isaacsas:
Sure, but let me get a bit of mass action reaction testing done with nonlinear reactions before we put up any new docs. I don't want to advertise the feature quite yet. I'm hoping to get a nonlinear reaction test up and running this weekend, and then that would be a good point to get this all documented (both for users and for developers that might want to add their own SSAs). I'll make the doc updates the next priority after this testing (before putting through any more SSAs).

Can you point me to which documentation I should update (and where it is located)?

Also, I'll have a slight tweak to put through on this PR (which I guess I'll put through as a second PR). I noticed that the `JumpAggregation` constructor I created with kwargs was missed by the code coverage. I think it needs to have parametric types and forward them to the parametric inner constructor.

ChrisRackauckas:
Sounds good.

isaacsas:
Also, thanks for adding me to the org! I'm happy to contribute.

PR #25, opened 2018-04-07
updating JumpAggregation constructors, adding MassActionJump nonlinear rx tests

isaacsas:
1. Allow kwargs to be sent to the JumpAggregation constructor. Right now it isn't used, but future SSAs could require options to be passed down in this way. Code coverage showed the previous commit's version of the outer constructor wasn't actually being called.
2. Added two tests, a small gene expression network and a system with a mix of bimolecular and trimolecular reactions, to test `MassActionJumps`.

### DIFF
--- a/src/aggregators/direct.jl
+++ b/src/aggregators/direct.jl
@@ -9,8 +9,11 @@ mutable struct DirectJumpAggregation{T,S,F1,F2,RNG} <: AbstractSSAJumpAggregator
   affects!::F2
   save_positions::Tuple{Bool,Bool}
   rng::RNG
+  DirectJumpAggregation{T,S,F1,F2,RNG}(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T, maj::S, rs::F1, affs!::F2, sps::Tuple{Bool,Bool}, rng::RNG) where {T,S,F1,F2,RNG} = 
+    new{T,S,F1,F2,RNG}(nj, njt, et, crs, sr, maj, rs, affs!, sps, rng)
 end
-DirectJumpAggregation(nj, njt, et, crs, sr, maj, rs, affs!, sps, rng; kwargs...) = DirectJumpAggregation(nj, njt, et, crs, sr, maj, rs, affs!, sps, rng)
+DirectJumpAggregation(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T, maj::S, rs::F1, affs!::F2, sps::Tuple{Bool,Bool}, rng::RNG; kwargs...) where {T,S,F1,F2,RNG} = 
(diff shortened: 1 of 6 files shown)


src/aggregators/direct.jl
```diff
@@ -9,8 +9,11 @@ mutable struct DirectJumpAggregation{T,S,F1,F2,RNG} <: AbstractSSAJumpAggregator
   affects!::F2
   save_positions::Tuple{Bool,Bool}
   rng::RNG
+  DirectJumpAggregation{T,S,F1,F2,RNG}(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T, maj::S, rs::F1, affs!::F2, sps::Tuple{Bool,Bool}, rng::RNG) where {T,S,F1,F2,RNG} =
```
ChrisRackauckas: why is this constructor necessary?

src/aggregators/frm.jl
```diff
@@ -9,8 +9,11 @@ mutable struct FRMJumpAggregation{T,S,F1,F2,RNG} <: AbstractSSAJumpAggregator
   affects!::F2
   save_positions::Tuple{Bool,Bool}
   rng::RNG
+  FRMJumpAggregation{T,S,F1,F2,RNG}(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T, maj::S, rs::F1, affs!::F2, sps::Tuple{Bool,Bool}, rng::RNG) where {T,S,F1,F2,RNG} =
```
ChrisRackauckas: why is this inner constructor necessary?

ChrisRackauckas:
Looks great, though I don't know why those inner constructors need to exist.

src/aggregators/direct.jl
isaacsas: This may be my not quite understanding parametric constructors, but when I just had an outer constructor it seemed like the outer constructor was being shadowed by the default inner constructor. If you look at my last PR with the FRM addition, the outer constructor was being missed in the code coverage. I also tried with that code having the outer constructor just print a message, and that indeed showed it wasn't ever getting called. This updated version seems to properly route through the outer constructor when creating the `JumpAggregation`.

isaacsas: Actually I misremembered; it is the exact opposite. Without the inner constructor I get a warning about the default constructor being overwritten.

WARNING: Method definition (::Type{DiffEqJump.DirectJumpAggregation{T, S, F1, F2, RNG} where RNG where F2 where F1 where S where T})(Int64, T, T, Array{T, 1}, T, S, F1, F2, Tuple{Bool, Bool}, RNG) in module DiffEqJump at /Users/isaacsas/.julia/v0.6/DiffEqJump/src/aggregators/direct.jl:2 overwritten at /Users/isaacsas/.julia/v0.6/DiffEqJump/src/aggregators/direct.jl:18.

PR #30, opened 2018-04-13
Expanding MassActionJump input options

isaacsas:
Allows for passing in vectors of `MassActionJump`s in two formats:

1. Each jump stores a scalar rate constant and associated stoichiometry.
2. Each jump stores a vector of rate constants and vectors of stoichiometry. 

Internally these are all converted to a global `MassActionJump` that stores vectors of rates and stoichiometry.

Updated tests to handle more possible input combinations of `MassActionJumps` and `ConstantRateJumps`, both as splatted tuples in the `JumpProblem` and as options passed through a `JumpSet`.

### DIFF
--- a/src/jumps.jl
+++ b/src/jumps.jl
@@ -106,17 +106,17 @@ regular_jump_combine(rj1::RegularJump,rj2::RegularJump) = error("Only one regula
 
 # functionality to merge two mass action jumps together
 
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
@Gaussia I think this is what you needed. I'll tag it.

Gaussia:
Looks useful, I will update the reaction reader to use this.

PR #36, opened 2018-04-24
Cleanup MassActionJump reaction rate function interface

isaacsas:
Tweaks the definition of `evalrate` and `executerx!` so that aggregators don't need to know the internals of `MassActionJump`. Now just the reaction id to execute and the `MassActionJump` are passed in. The routines handle modifying/using the correct reaction. Also modifies the `MassActionJump` parametric representation to no longer require the same type for reaction stoichiometry and net stoichiometry.

This should make it easier to play with different internal `MassActionJump` implementations as aggregators don't need to be updated if the internal fields of `MassActionJump` change. (For example, I'm using this to play around with static arrays, which do seem to offer some performance benefit.)

Going forward, changing the internal representation of a `MassActionJump` should now only require updates in `jumps.jl`and in `massaction_rates.jl`. I've tried to remove any use of internal mass action jump fields outside these two files.

### DIFF
--- a/src/jumps.jl
+++ b/src/jumps.jl
@@ -36,12 +36,12 @@ RegularJump(rate,c,dc;mark_dist = nothing,constant_c = false) =
             RegularJump(rate,c,dc,mark_dist,constant_c)
 
 
-struct MassActionJump{T,S} <: AbstractJump
+struct MassActionJump{T,S,U} <: AbstractJump
   scaled_rates::T
   reactant_stoch::S
-  net_stoch::S
+  net_stoch::U
 
-  function MassActionJump{T,S}(rates::T, rs_in::S, ns::S, scale_rates::Bool) where {T <: AbstractVector,S}
+  function MassActionJump{T,S,U}(rates::T, rs_in::S, ns::U, scale_rates::Bool) where {T <: AbstractVector, S, U}
     sr  = copy(rates)
     rs = copy(rs_in)
     for i in eachindex(rs)
@@ -55,7 +55,7 @@ struct MassActionJump{T,S} <: AbstractJump
     end
     new(sr, rs, ns)
   end
-  function MassActionJump{T,S}(rate::T, rs_in::S, ns::S, scale_rates::Bool) where {T <: Number,S}    
+  function MassActionJump{T,S,U}(rate::T, rs_in::S, ns::U, scale_rates::Bool) where {T <: Number, S, U}    
     rs = rs_in
     if (length(rs) == 1) && (rs[1][1] == 0)
       rs = typeof(rs)()
@@ -65,7 +65,7 @@ struct MassActionJump{T,S} <: AbstractJump
   end
 
 end
-MassActionJump(usr::T, rs::S, ns::S; scale_rates = true) where {T,S} = MassActionJump{T,S}(usr, rs, ns, scale_rates)
+MassActionJump(usr::T, rs::S, ns::U; scale_rates = true) where {T,S,U} = MassActionJump{T,S,U}(usr, rs, ns, scale_rates)
 
 struct JumpSet{T1,T2,T3,T4} <: AbstractJump
   variable_jumps::T1
--- a/src/jumps.jl
+++ b/src/jumps.jl
@@ -67,6 +67,10 @@ struct MassActionJump{T,S,U} <: AbstractJump
 end
 MassActionJump(usr::T, rs::S, ns::U; scale_rates = true) where {T,S,U} = MassActionJump{T,S,U}(usr, rs, ns, scale_rates)
 
+function get_num_majumps(maj)
+  length(maj.scaled_rates)
+end
+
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
Note that WIP is usually used for "Don't Merge - still updating".

src/jumps.jl
```diff
@@ -55,7 +55,7 @@ struct MassActionJump{T,S} <: AbstractJump
     end
     new(sr, rs, ns)
   end
-  function MassActionJump{T,S}(rate::T, rs_in::S, ns::S, scale_rates::Bool) where {T <: Number,S}    
+  function MassActionJump{T,S,U}(rate::T, rs_in::S, ns::U, scale_rates::Bool) where {T <: Number, S, U}
```
ChrisRackauckas: Get rid of the `{T,S,U}`.

src/massaction_rates.jl
```diff
@@ -3,11 +3,10 @@
 # stochiometric coefficient.
 ###############################################################################
 
-@fastmath function evalrxrate(speciesvec::AbstractVector{T}, rateconst,
-                              stochmat::AbstractVector{Pair{S,V}})::typeof(rateconst) where {T,S,V}
+@inline @fastmath function evalrxrate(speciesvec::AbstractVector{T}, rxidx::S,
```
ChrisRackauckas: is the inlining necessary here?

src/massaction_rates.jl
```diff
@@ -16,19 +15,19 @@
         end
     end
 
-     rateconst * val
+    @inbounds return val * majump.scaled_rates[rxidx] 
 end
 
-@inline @fastmath function executerx!(speciesvec::AbstractVector{T},
-                                      net_stoch::AbstractVector{Pair{S,V}}) where {T,S,V}
+@inline @fastmath function executerx!(speciesvec::AbstractVector{T}, rxidx::S,
```
ChrisRackauckas: inlining?

ChrisRackauckas:
Just a cosmetic comment (fixes a deprecation that will happen in v0.7) and a question about inlining. Feel free to merge when answering.

src/jumps.jl
isaacsas: I think those are necessary; without them I get:

```
ERROR: LoadError: LoadError: syntax: too few type parameters specified in "new{...}"
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:576
 [2] include(::String) at /Applications/Julia-0.6.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [3] include_from_node1(::String) at /Applications/Julia-0.6.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [4] include(::String) at /Applications/Julia-0.6.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 [5] anonymous at ./<missing>:2
while loading /Users/isaacsas/.julia/v0.6/DiffEqJump/src/jumps.jl, in expression starting on line 39
```
...

The problem was that I wanted to have an optional named parameter, `scale_rates`, for which I added the outer constructor. We want the outer constructor to always be called, even if the user passes only the three required parameters (which is what the default inner constructor would take). As far as I can tell, the explicit typing in the inner function definitions, and the call to the explicitly typed form from the outer constructor, are the only way to avoid the outer constructor being shadowed. 

If you know an alternative way to set this up I'm happy to change it.

ChrisRackauckas: Oh that's my bad. From the diff I didn't see it was an inner constructor.

src/massaction_rates.jl
isaacsas: There's something going on with the compiler that I can't quite figure out. Without the inlining, the previous version, which indexes into the `MassActionJump` in the aggregator and then passes in the stoichiometry and rate, was noticeably faster (10-20%). Inlining eliminated the performance difference. In contrast, I don't think inlining the old version had much effect on performance. I'm not sure why this is -- I haven't quite mastered reading `@code_warntype` yet.

The conclusion though is that the inlining does make a big difference for this version of the function. Since this routine is often the rate limiting part of any Direct method simulation it seemed worth it. (That's why I'm also playing around with static vectors for the internal jump representation -- every bit of performance gain here will help.)

isaacsas:
So perhaps I'm missing something, but should there be a merge button somewhere? I'm not seeing it on this page.

src/massaction_rates.jl
ChrisRackauckas: Sounds good.

ChrisRackauckas:
I upgraded the privs of the @JuliaDiffEq/bio group to write for this and DiffEqBiological.jl. The PRs should still get reviewed by someone (unless it's some small bugfix), but that'll help reduce some friction.

isaacsas:
Great, that will be useful. (And I certainly agree that having someone review everything but minor bugfixes is key.)

Issue #37, opened 2018-04-25
Keyword argument passing problems in solving a JumpProblem


onoderat:
The following minimal working example shows there is some issues with passing a keyword argument into the solve function. 

The example is basically comprised of code in the documentation, except that EM() which requires "dt" was used instead of SRIW1. 

```
using DifferentialEquations

function f(du,u,p,t)
  du[1] = u[1]
end

function g(du,u,p,t)
  du[1] = 0.1
end

prob = SDEProblem(f,g,[0.2],(0.0,10.0))

rate(u,p,t) = 2
affect!(integrator) = (integrator.u[1] = integrator.u[1]/2)
jump = ConstantRateJump(rate,affect!)

jump_prob = JumpProblem(prob,Direct(),jump)

sol_SRIW1 = solve(jump_prob, SRIW1()); #This works
sol_EM = solve(jump_prob, EM(), dt=1e-4); #Does not work
```
The error message and stack trace is given by
```
Fixed timestep methods require a choice of dt or choosing the tstops

Stacktrace:
[...]

PR #38, opened 2018-04-26
named `dt` parameter in `init` not forwarded

isaacsas:
Fix for [#37](https://github.com/JuliaDiffEq/DiffEqJump.jl/issues/37).

Note, it seems like there could be a similar issue with the next line where the `save_positions` named parameter is set. It also isn't forwarded. Should I remove it too (or forward it)? 

I can update this PR tomorrow to fix that too if it should also be modified.


ChrisRackauckas:
`save_positions` is done in the `JumpProblem` construction, so that's an error too. I think that this evolved but the args weren't deleted on accident. Just `callback` needs to stay so it can be merged with the one from the jump problem.

isaacsas:
OK, will update shortly.

PR #40, opened 2018-04-27
Sorting direct method

isaacsas:
This is pretty much a straight implementation of the sorting direct method as presented in the original manuscript:

McCollum et al, "The sorting direct method for stochastic simulation of biochemical systems with varying reaction execution behavior", Comp. Bio. and Chem., 30, pg. 39-49 (2006). 

Notes:
1. This method requires a dependency graph. I've supplied routines that for **pure** mass action systems calculate the dependency graph internally. Otherwise, for mixed jump types the user must pass a dependency graph as a named option to `JumpProblem` (`dep_graph`). 
2. In the case of mixed jump types, this dependency graph currently assumes all mass action jumps are ordered before constant rate jumps (with each collection ordered in the same order they are ordered within the underlying `JumpSet`). We could consider adding functionality for users to specify the reaction ordering in their dependency graph, and then re-order it within the `JumpAggregator`. For performance and simplicity of code reasons, even if we add that option I think we should pick an assumed internal ordering that is used within the actual core SSA code and translate to it when setting up the `JumpAggregator`.
3. For the types of small systems we have in tests this method doesn't really show any significant speed improvement. However, for the 20 gene example of the paper I am finding an 8x speed up over `Direct()`. If I'm remembering right, I think the multi-state example I added to the problem library gets ~25% speedup.

If there are thoughts on how better to handle dependency graphs for mixed jump type systems, I can make updates to this PR.


src/aggregators/sortingdirect.jl
```diff
@@ -0,0 +1,142 @@
...
+    sum_rate::T
+    ma_jumps::S
+    rates::F1
+    affects!::F2
+    save_positions::Tuple{Bool,Bool}
+    rng::RNG
+    dep_gr::DEPGR
+    jump_search_order::Vector{Int}
+    jump_search_idx::Int
+    SortingDirectJumpAggregation{T,S,F1,F2,RNG}(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T, maj::S,
```
ChrisRackauckas: why as an inner constructor?

src/aggregators/ssajump.jl
```diff
@@ -68,6 +68,54 @@ function build_jump_aggregation(jump_agg_type, u, p, t, end_time, ma_jumps, rate
                 majumps, rates, affects!, save_positions, rng; kwargs...)
 end
 
-DiscreteCallback(c::AbstractSSAJumpAggregator) =DiscreteCallback(c, c, initialize = c, save_positions = c.save_positions)
+# reevaluate all rates and total rate
+function fill_rates_and_sum!(p, u, params, t)
```
ChrisRackauckas: change the name of `p` to make the naming more standard?

src/aggregators/sortingdirect.jl
isaacsas: The previous jump aggregations had needed this setup to avoid the outer constructor getting shadowed. I guess this may not be an issue here. I'll look into this and the `p` notation in ssajump.jl with my next PR (probably next reaction method).

isaacsas:
The method is going to do well when there are timescale separations so some (small number of?) jumps become very frequent. They'll then bubble up to the top of the search order, and I guess even though the access is nonlinear, only searching a few entries gives a big win. With smaller systems, or I'd suspect systems with relatively uniform jump rates, this isn't going to do better. 

This is what BioNetGen uses, and it seems to offer very good performance (at least according to the Mendes review). But perhaps they have some other implementation tricks (for example, they are apparently using a fast but very poor random number generator). It will be interesting to compare across a variety of networks once I get to some systematic benchmarking.

I actually was thinking of trying a variant where the jump rates in `cur_rates` get reordered, so the search is always in order. Then the dependent reaction updating would be out of order and need a map from the reordered index to the original reaction index, but usually only a few reactions get updated anyways... I'd still like to play with that eventually, but figured for the first version I'd just follow the paper.

PR #44, opened 2018-05-02
Extending the diffusion function in extend_problem

onoderat:
See following issue to understand the bug this commit fixes.
DiffEqJump.jl/issues/43


ChrisRackauckas:
Thanks!

PR #45, opened 2018-05-05
Next Reaction Method

isaacsas:
For completeness, this PR provides a reference implementation of the Gibson and Bruck next reaction method. 

This could ultimately use some optimization of the indexed priority queue data structure; right now the DataStructures.jl `mutable_min_heap` is used (which was substantially faster than its `PriorityQueue` implementation). The queue updating is rate limiting, and causes the method to be noticeably slower than the direct method on simple examples. For example, on the negative feedback by dimers gene repression example in the problem library this method is almost a factor of two slower than `Direct`. In contrast, published benchmarks of that network show that NRM should be comparable in performance.

Whether it is worth investing effort in optimizing this more is questionable; there are many better optimized SSAs that would be faster even if we had a faster priority queue implementation. I figured we could merge this for now for completeness, but not recommend it in the docs -- I'll start more thorough benchmarking soon to see how our various methods all compare across different types and sizes of networks. Should more performant Julia PQ implementations become available we can switch to them in the future.


isaacsas:
@alanderos91 If you have any suggestions to improve the performance here (alt PQ implementations) or other modifications please let me know. Thanks!

PR #47, opened 2018-05-22
remove Distributions.jl and use PoissonRandom.jl

ChrisRackauckas:


ChrisRackauckas:
Requires https://github.com/JuliaLang/METADATA.jl/pull/14795

PR #53, opened 2018-10-02
Reenable SSAs using FunctionWrappers

isaacsas:
As FunctionWrappers is now working on 1.0 I've hopefully got these all re-enabled correctly.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,7 +3,7 @@ DiffEqBase 4.27.0
 RecursiveArrayTools
 RandomNumbers
 Compat 0.19.0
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
Since FunctionWrappers is being deprecated, we should really try to get these setup with `@cfunction`. But this is good for now.

isaacsas:
I wasn't aware of `@cfunction`, I'll take a look at it. I'm in grant/teaching season now, so probably won't get much coding done for a month or so, but hopefully at the end of the semester I can make another push.

PR #55, opened 2018-10-24
fixing dependency graph initialization bug in NRM and SortingDirect

isaacsas:
Was passing number of reactions and not species. Usually there are more reactions than species, so bug was not caught...


isaacsas:
@ChrisRackauckas When you have a chance could you tag a release of DiffEqJump with this fix, and DiffEqBiological with the other fix I submitted. (I have students using them I'd like to tell to upgrade.) Thanks!

ChrisRackauckas:
Releasing right now.

isaacsas:
Thanks!

PR #56, opened 2018-11-05
implementation of the Rejection SSA (RSSA)

isaacsas:
See Marchetti, Priami and Thanh - Simulation Algorithms for Computational Systems Biology, for the version I've implemented. Some notes (more for myself):

1. Requires some new functionality (maps from species to dependent jumps, and jumps to dependent species). This has to be passed in as kwargs to the JumpProblem. I'll shortly submit a DiffEqBiological PR that will handle this for systems generated as a reaction_networks. 
2. I'm not seeing performance comparable to what the authors claim on simple examples (of course, the main benefit should be on larger systems than we currently test against). For example, on their own gene network test, which is in the problem library, I get comparable performance to `Direct` (vs a 25% speed-up they observe). Profiling shows that my implementation is slowed primarily by evaluating the mass action reaction rates, and by random number sampling in `generate_jumps!`. Changing the default random number generator can improve performance, but still does not reproduce their relative timings.
3. I'm generating the needed Erlang random variate as a sum of exponentials, in turn generated with `randexp`. An alternative is to use inverse transform sampling, taking the log of a product of uniform random numbers. This was generally slower in my testing.
4. All jump rate functions must be monotone with respect to populations. It is assumed mass action functions are monotone non-decreasing, but constant rate jumps are only assumed monotone.
5. `execute_jumps!` is doing redundant work, as rate bounds are recalculated more than once each for each event (they are recalculated for each species that has gone outside its bracketing interval). This could be fixed by saving a `Set` of marked jumps that need to have their bracketing interval recalculated, then going through them after updating all species. It's not clear if this will actually make a noticeable speed difference for typical networks where only a few species are updated.

Overall, it may just be that our `Direct` implementation is really good for small systems and beats the one by Marchetti et al. Looking at some larger networks would be good to confirm we are still seeing the expected speed-up on large systems.

### DIFF
--- a/src/DiffEqJump.jl
+++ b/src/DiffEqJump.jl
@@ -23,6 +23,8 @@ include("aggregators/direct.jl")
 include("aggregators/frm.jl")
 include("aggregators/sortingdirect.jl")
 include("aggregators/nrm.jl")
+include("aggregators/bracketing.jl")
+include("aggregators/rssa.jl")
 include("problem.jl")
 include("callbacks.jl")
 include("solve.jl")
@@ -39,7 +41,7 @@ export JumpProblem
 
 export SplitCoupledJumpProblem
 
-export Direct, DirectFW, FRM, FRMFW, SortingDirect, NRM, RSSA
+export Direct, DirectFW, FRM, FRMFW, SortingDirect, NRM, BracketData, RSSA
 
 export get_num_majumps, needs_depgraph, needs_vartojumps_map
 
--- a/src/DiffEqJump.jl
+++ b/src/DiffEqJump.jl
@@ -39,7 +39,7 @@ export JumpProblem
 
 export SplitCoupledJumpProblem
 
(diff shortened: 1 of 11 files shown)


isaacsas:
@ChrisRackauckas OK if I merge this?

isaacsas:
@ChrisRackauckas Thanks! If you could tag a release when you have a chance that will unblock a PR I have for DiffEqBiological too.

ChrisRackauckas:
Yup I'll do the release right now. And I bumped you up to admin privileges for this repo.

isaacsas:
OK, great! Going forward, is there a better workflow than me pestering you about PRs/releases? Should I just let a PR sit for a few days and if you don't get to it merge/release it myself? When I submit a bigger PR like this should I request a review by you? 

I appreciate having others take a look, but also don't want to bug you (I know you're probably swamped looking at stuff across all the different DiffEq packages).

PR #60, opened 2019-01-20
Changes in pois_rand

devmotion:
This PR updates to the new order of arguments in `pois_rand` (https://github.com/JuliaDiffEq/PoissonRandom.jl/pull/4).

As mentioned in https://github.com/JuliaDiffEq/PoissonRandom.jl/pull/4#issuecomment-454691343, we probably should upper bound current versions of DiffEqJump and then add a lower bound to a new release of PoissonRandom here?

### DIFF
--- a/src/simple_regular_solve.jl
+++ b/src/simple_regular_solve.jl
@@ -38,7 +38,7 @@ function DiffEqBase.solve(jump_prob::JumpProblem,alg::SimpleTauLeaping;
       tprev = t[i-1]
       rate(rate_cache,uprev,p,tprev)
       rate_cache .*= dt
-      counts .= pois_rand.(rate_cache,(rng,))
+      counts .= pois_rand.(Ref(rng), rate_cache)
       !rj.constant_c && c(dc,uprev,p,tprev,mark)
       mul!(update,dc,counts)
       u[i] = uprev .+ update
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
>As mentioned in JuliaDiffEq/PoissonRandom.jl#4 (comment), we probably should upper bound current versions of DiffEqJump and then add a lower bound to a new release of PoissonRandom here?

I'm scared of the resolver breaking. I might want to just update really fast.

src/simple_regular_solve.jl
```diff
@@ -38,7 +38,7 @@ function DiffEqBase.solve(jump_prob::JumpProblem,alg::SimpleTauLeaping;
       tprev = t[i-1]
       rate(rate_cache,uprev,p,tprev)
       rate_cache .*= dt
-      counts .= pois_rand.(rate_cache,(rng,))
+      counts .= pois_rand.(Ref(rng), rate_cache)
```
ChrisRackauckas: why Ref instead of tuple?

ChrisRackauckas:
Changed the Ref back to a tuple. Since a Ref is mutable, it has to rely on escape analysis to make sure it doesn't allocate, and I'm not sure if that happens here. But I know a tuple won't allocate and we don't need mutability, so it's a nice safe bet.

src/simple_regular_solve.jl
devmotion: Oh I assumed the tuple was used to avoid broadcasting and thought using `Ref` would be the standard way to achieve this.

ChrisRackauckas: Yeah, I've always just used a tuple for it

PR #62, opened 2019-01-20
Update comparisons with nothing

devmotion:


isaacsas:
@devmotion Is this done? (Just wanted to double check before merging.) Thanks!

PR #65, opened 2019-02-12
Direct Composition Rejection SSA

isaacsas:
Implementation of the Direct Composition-Rejection method from

*A constant-time kinetic Monte Carlo algorithm for simulation of large biochemical reaction networks*,
by A. Slepoy, A.P. Thompson and S.J. Plimpton, J. Chem. Phys, 128, 205101 (2008).

incorporating optimizations from 

*Efficient Formulations for Exact Stochastic Simulation of Chemical Systems*,
by S. Mauch and M. Stalzer, ACM Trans. Comp. Biol. and Bioinf., 8, No. 1, 27-35 (2010).

~~Before merging it needs~~
1. ~~Further testing for accuracy for some networks beyond the basic jump tests.~~
2. ~~`@inbounds` decoration throughout.~~
3. ~~handle extinction problems correctly. (check for other SSAs too)~~
4. ~~handle non-power of 2 min priorities in the JumpAggregation (max handled now)~~

OK, this is ready to go; all relevant tests are passing locally. I still want to test `RSSA` and `DirectCR` accuracy more broadly before adding to the docs, but in my testing so far they seem ok.

@ChrisRackauckas Let me know if you want to look over before I merge (I won't merge till the [VariableRate jump issue](https://github.com/JuliaDiffEq/DiffEqJump.jl/issues/64) is fixed so I can make sure tests pass, so it'll be a few days.)


src/aggregators/directcr.jl
```diff
@@ -0,0 +1,178 @@
...
+Composition-Rejection Direct Method (DirectCR), implementation combining
+features from the original article:
+*A constant-time kinetic Monte Carlo algorithm for simulation of large biochemical reaction networks*,
+by A. Slepoy, A.P. Thompson and S.J. Plimpton, J. Chem. Phys, 128, 205101 (2008).
+and
+*Efficient Formulations for Exact Stochastic Simulation of Chemical Systems*,
+by S. Mauch and M. Stalzer, ACM Trans. Comp. Biol. and Bioinf., 8, No. 1, 27-35 (2010).
+"""
+
+const MINJUMPRATE = 2.0^exponent(1e-12)
```
ChrisRackauckas: I don't get why this needs to have a minimum? It needs a binning rate, but why not have the first bin go from 0 to binrate, then exponentiate from there?

isaacsas: binrate is what I'm calling `minrate` (which gets set to `MINJUMPRATE` by default but can be passed into the aggregator as a kwarg). One should never sample from the bin with zero elements, but one could sample from the bin that has elements in (0,`minrate`). If there are lots of zero elements, and one is in a situation where the min non-zero bin is actually getting hit, then I was thinking that could make a performance difference. In practice I'm guessing for many "real-world" networks hitting the min non-zero bin will be exceedingly rare. (In general, one probably does want to pass in a maxpriority as the defaults could add way too many bins on the upper end, which is where you search down from.) 

I made the bin boundaries powers of two to make the bin calculation as simple as possible (i.e. so that the integer exponent gives the bin boundaries). Which now makes me realize I actually haven't checked set up the code to handle non-power of two min/max priorities passed in by the user... So that needs to get looked at before merging.

isaacsas: I guess the argument if you're actually hitting the min non-zero bin a lot would be that you should've given a smaller `minrate` from the start. So I think I'll follow your suggestion and get rid of the (0,minrate) bin.

ChrisRackauckas: Oh I was actually looking for it. I think you need the (0,minrate) bin. It wasn't clear to me from the code/paper it was in there. I think with enough reactions it would bias it to not be included.

isaacsas: Right we're discussing slightly different issues I think.

I had both a zero only bin and then a (0,minrate) bin (open at the ends). One could alternatively have just one [0,minrate) bin. The latter would eliminate an if statement in the mapping function, which might be a win over separating out the zero rates. But maybe that's an optimization to worry about later.

ChrisRackauckas: Yeah, we can take a look at it later. I see what's going on now. Yeah, I was expecting `[0,minrate)` and was slightly confused by the explicit zero handling.

isaacsas:
> wow, I didn't know there was an O(1) SSA

Well, O(number of groups) + O(average number of updates per reaction), but if they're bounded then it is O(1). The defaults I'm using make a fair number of groups (O(100)), so actually calculating an initial `maxpriority` from the rates will generally be a good strategy (otherwise many of the groups one searches down from will be empty).  That's something to play around with optimizing later, along with periodically collapsing down the table by eliminating empty groups at the top. The Mauch paper assumes a 32 bit RNG, and uses 43 bins (allowing for 10 empty bins at the upper end).

What this enables now is an even faster algorithm for big networks, by combining this group'ing idea with the RSSA. That should cut down the update cost by avoiding a decent fraction of them.

isaacsas:
And this table data structure is exactly what I need for my spatial SSAs.

src/aggregators/directcr.jl
isaacsas: I'll add some of these optimization comments to the SSA book keeping thread so they don't get forgotten.

isaacsas: Just to add, I tried to comment the table datastructure code pretty thoroughly and write it a bit more generically since it should be re-usable as a component in other SSAs or by other people (or at least by me for spatial SSAs).

isaacsas:
Is it ok to merge this despite tests failing? The failures aren't related to this method (the relevant tests this would influence all pass locally). (I had been waiting on a fix for [64](https://github.com/JuliaDiffEq/DiffEqJump.jl/issues/64) before merging.)

ChrisRackauckas:
Yeah, let's do it.

isaacsas:
Thanks!

Issue #71, opened 2019-06-14
Jump DDEs error


korsbo:
Hi, 

I tried to run a Jump DDE and, given the supreme coolness of DifferentialEquations.jl, I rather expected this to just work. However, it did not. 

M(not)WE:
```julia
using DifferentialEquations

function dde(du, u, h, p, t)
    du[1] = h(p, t-p[1], idxs=1)
end
h(p, t; kwargs...) = 0
p = [ 5.]
tspan = (0., 15)
u0 = [1.]

prob = DDEProblem(dde, u0, h, tspan, p)

### Add some Gillespie action
prod_rate(u,p,t) = (1 + 9 * u[1] / (5 + u[1]))/10
prod_affect!(i) = (i.u[1] += 1)
prod_jump = ConstantRateJump(prod_rate, prod_affect!)

deg_rate(u,p,t) = u[1]/10
deg_affect!(i) = (i.u[1] -= 1)
deg_jump = ConstantRateJump(deg_rate, deg_affect!)

jump_prob = JumpProblem(prob, Direct(), prod_jump, deg_jump)

sol = solve(jump_prob, MethodOfSteps(Tsit5()))
```

```
[...]

PR #72, opened 2019-06-18
Fix hybrid Jump/DDE and Jump/DAE problems

alanderos91:
This should address #71. The issue is due to (i) typos in `extend_problem` and (ii) `DDEProblem` not using `recompile_flag` in its `__init` function.

Is this an appropriate place to add tests that catch breaking changes to these hybrid problems? I don't have the expertise to make sure the DDE and DAE problems are solved correctly, but we can at least detect when something goes wrong.


isaacsas:
Thanks! I'm going to leave merging for @ChrisRackauckas since I'm not so familar with the DAE/DDE code, but this definitely is a reasonable place to add tests for the hybrid algorithms. Please feel free to add some!

PR #88, opened 2019-11-28
Make cache vector of same type as u0

baggepinnen:


ChrisRackauckas:
Does this fix the autodiff? I assumed it would be something simple.

baggepinnen:
It does not due to 
```
MethodError: no method matching procf(::ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2}, ::Int64, ::ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2})
Closest candidates are:
  procf(::Any, ::Int64, !Matched::Float64) at /home/fredrikb/.julia/packages/PoissonRandom/Ohb1g/src/PoissonRandom.jl:119
ad_rand(::RandomNumbers.Xorshifts.Xoroshiro128Plus, ::ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2}) at PoissonRandom.jl:48
pois_rand(::RandomNumbers.Xorshifts.Xoroshiro128Plus, ::ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2}) at PoissonRandom.jl:145
_broadcast_getindex_evalf at broadcast.jl:630 [inlined]
_broadcast_getindex at broadcast.jl:603 [inlined]
getindex at broadcast.jl:563 [inlined]
macro expansion at broadcast.jl:909 [inlined]
macro expansion at simdloop.jl:77 [inlined]
copyto! at broadcast.jl:908 [inlined]
copyto! at broadcast.jl:863 [inlined]
materialize! at broadcast.jl:822 [inlined]
#solve#136(::Nothing, ::Float64, ::typeof(solve), ::JumpProblem{DiscreteProblem{Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1},Tuple{Float64,Float64},true,Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1},DiscreteFunction{true,DiffEqBase.var"#161#162",Nothing,Nothing},Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}},Direct,CallbackSet{Tuple{},Tuple{}},Nothing,Tuple{},RegularJump{typeof(rate),typeof(c),Array{Float64,2},Nothing},Nothing}, ::SimpleTauLeaping) at simple_regular_solve.jl:41
(::DiffEqBase.var"#kw##solve")(::NamedTuple{(:dt,),Tuple{Float64}}, ::typeof(solve), ::JumpProblem{DiscreteProblem{Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1},Tuple{Float64,Float64},true,Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1},DiscreteFunction{true,DiffEqBase.var"#161#162",Nothing,Nothing},Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}},Direct,CallbackSet{Tuple{},Tuple{}},Nothing,Tuple{},RegularJump{typeof(rate),typeof(c),Array{Float64,2},Nothing},Nothing}, ::SimpleTauLeaping) at none:0
forward_pass(::Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1}) at jumpdiff.jl:45
vector_mode_gradient(::typeof(forward_pass), ::Array{Float64,1}, ::ForwardDiff.GradientConfig{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2,Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1}}) at apiutils.jl:37
gradient(::Function, ::Array{Float64,1}, ::ForwardDiff.GradientConfig{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2,Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1}}, ::Val{true}) at gradient.jl:17
gradient(::Function, ::Array{Float64,1}, ::ForwardDiff.GradientConfig{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2,Array{ForwardDiff.Dual{ForwardDiff.Tag{typeof(forward_pass),Float64},Float64,2},1}}) at gradient.jl:15
gradient(::Function, ::Array{Float64,1}) at gradient.jl:15
(::var"#21#22")(::Array{Float64,1}) at jumpdiff.jl:55
top-level scope at jumpdiff.jl:58

```

ChrisRackauckas:
Oh yes, that would break it... I see. The difference from SDEs to this is that SDEs always use the same way to generate the random numbers (randn()) independently of any state values, and then multiply something to that, so you can treat those as constants. Here the distribution really does need the value of the state, since the rates are dependent on the states and you can't just rescale it by a factor after generating the random numbers. That would make it non-AD differentiable.

ChrisRackauckas:
Well this is necessary anyways. Thanks!

ChrisRackauckas:
I'm not sure how you'd do particle sampling of a Poisson RNG? Something to think about though. Make it generate N random outputs?

baggepinnen:
Yeah it would draw one sample from each poisson distribution represented by the different particles.

ChrisRackauckas:
That might directly require a dispatch on the rand functions.

baggepinnen:
Yep, me and Chad Scherrer have been doing some thinking about that here
https://github.com/baggepinnen/MonteCarloMeasurements.jl/issues/22
So far we have not converged to a systematic way of doing this.

PR #89, opened 2019-12-05
Matrix-Free Tau Leaping [WIP]

arnavs:
Rudimentary implementation. Still want to optimize a bit, but opening mainly for suggestions and review.


ChrisRackauckas:
I think we should just deprecate the old RegularJump form and completely go to this. @isaacsas anything you see here that would really require having that matrix? If it's not needed, let drop it.

arnavs:
@ChrisRackauckas Note that in this PR, I still kept a dummy stoichiometry matrix argument to tell us about the number of jumps. (Mainly for backwards-compatibility, but also we do need to know.)

Of course, there's better ways to communicate that information... just pass in the number of jumps. 

The Travis is failing is because of the 5-argument `mul!` I'm using in one of the tests (that method is only available for Julia 1.3+). I'll either change the test case or bump the `.travis.yml` version, whichever you prefer.

isaacsas:
I’d prefer to fix the test to work on 1.2 as not everyone (including me) has upgrading to 1.3 yet.

Otherwise this looks fine with me, and I don’t see any reason to require a matrix.

arnavs:
Roger that.

arnavs:
OK, @isaacsas I made the change. 

Once we get this out there, I can prepare a PR to update the old interface and switch over to this sort of matrix-free method across the board. 

So, changing the `RegularJump` constructor, deleting the old method with the matrix multiplication,  updating docs, etc.

ChrisRackauckas:
yeah put a depwarn in the old constructor and make a compatible version as well, just to be nice.

arnavs:
@ChrisRackauckas Done.

isaacsas:
Looks good to me. I’ll let @ChrisRackauckas merge if the depwarning is what he was looking for.

src/jumps.jl
```diff
@@ -27,14 +27,18 @@ VariableRateJump(rate,affect!;
 struct RegularJump{R,C,DC,MD}
     rate::R
     c::C
-    dc::DC
+    dc::DC 
     mark_dist::MD
     constant_c::Bool
```
ChrisRackauckas: this isn't needed if it can't be specialized on anymore.

src/jumps.jl
```diff
@@ -27,14 +27,18 @@ VariableRateJump(rate,affect!;
...
 end
 
-RegularJump(rate,c,dc;mark_dist = nothing,constant_c = false) =
-            RegularJump(rate,c,dc,mark_dist,constant_c)
+RegularJump(rate,c,dc::AbstractMatrix; mark_dist = nothing,constant_c = false) =
+          RegularJump(rate,c,dc,mark_dist,constant_c, size(dc, 2))
 
+RegularJump(rate, c, n::Integer; mark_dist = nothing) = RegularJump(rate, c, nothing, mark_dist, nothing, n)
+
+@deprecate RegularJump(rate,c,dc;mark_dist = nothing,constant_c = false) RegularJump(rate, c, n; mark_dist = nothing)
```
ChrisRackauckas: that's not correct. It will error on old code. You should just do the simple conversion: call the function to get `dc` and do the multiplication. Basically, do the conversion for the user.

src/jumps.jl
```diff
@@ -24,21 +24,17 @@ VariableRateJump(rate,affect!;
...
 end
 
-RegularJump(rate,c,dc::AbstractMatrix; mark_dist = nothing,constant_c = false) =
-          RegularJump(rate,c,dc,mark_dist,constant_c, size(dc, 2))
-
-RegularJump(rate, c, n::Integer; mark_dist = nothing) = RegularJump(rate, c, nothing, mark_dist, nothing, n)
-
-@deprecate RegularJump(rate,c,dc;mark_dist = nothing,constant_c = false) RegularJump(rate, c, n; mark_dist = nothing)
+# deprecate old call
+@deprecate RegularJump(rate,c,dc::AbstractMatrix; mark_dist = nothing,constant_c = false) RegularJump(rate, c, mark_dist, constant_c, size(dc, 1), size(dc, 2))
```
ChrisRackauckas: this is still going to fail because it doesn't multiply dc...

arnavs: @ChrisRackauckas Sorry, am resolving things as I check them off locally. Will push these commits up when I'm ready for the review.

PR #93, opened 2020-04-06
ensure SSAStepper reseeds each init

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/585

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -21,7 +21,7 @@ end
 (integrator::SSAIntegrator)(t) = copy(integrator.u)
 (integrator::SSAIntegrator)(out,t) = (out .= integrator.u)
 
-function DiffEqBase.solve(jump_prob::JumpProblem,
+function DiffEqBase.__solve(jump_prob::JumpProblem,
                          alg::SSAStepper;
                          kwargs...)
     integrator = init(jump_prob,alg;kwargs...)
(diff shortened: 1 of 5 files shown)


src/SSA_stepper.jl
```diff
@@ -67,6 +68,9 @@ function DiffEqBase.init(jump_prob::JumpProblem,
     @assert length(jump_prob.jump_callback.discrete_callbacks) == 1
     cb = jump_prob.jump_callback.discrete_callbacks[1]
     prob = jump_prob.prob
+
+    reset_jump_problem!(jump_prob,seed)
```
isaacsas, line 72: @ChrisRackauckas Does this mean that every time `solve` is called `seed` will also be called with a new seed? I.e. if one runs an SSA repeatedly in a serial loop it will keep reseeding instead of just continuing from the current stream?

ChrisRackauckas: Yes. It already did that before, it just wasn't doing it during Ensemble solves.

isaacsas: Maybe this should just continue on the current stream for serial code; it seems undesirable to repeatedly reseed for serial calculations? Does it make sense to only seed if the user explicitly passes one in instead of having a default value? Higher level code, like an ensemble solve, could then pass different seeds for each processor/thread etc.

ChrisRackauckas: that would be something worthwhile.

PR #100, opened 2020-05-11
Allow static arrays in mass rate jumps

ChrisRackauckas:
This allows for static arrays to be used with mrj's for optimal speed on small problems.


isaacsas:
Should we reinvestigate using static arrays for storing the corresponding stoichiometry? I had gone with regular arrays for flexibility in the length of each stoichiometric vector, and type stability of the vector storing each stoichiometry vector. But we could hard-code a fixed length (two or maybe three) and use static arrays with an effective length indicator.

ChrisRackauckas:
It would be interesting to see if that gets the last factor, but I don't think that it'll do much.

Issue #101, opened 2020-05-15
some SSAs giving segmentation faults


isaacsas:
So far I've found RSSA, DirectCR now seem to crash on an inlined `getindex` call. MWE:

```julia
rn = @reaction_network begin
    β, S + I --> 2I
    γ, I --> R
end β γ
u0 = [999,1,0]; p = (0.1/1000,0.01); tspan = (0.,250.)
dprob2 = DiscreteProblem(rn, u0, tspan, p)
jprob2 = JumpProblem(dprob2, ssa, rn)

Nsims = 30000
function getmean(jprob,Nsims)
  m = 0.0
  for i = 1:Nsims
    sol = solve(jprob, SSAStepper())
    m += sol[end,end]
  end
  m/Nsims
end
m2 = getmean(jprob2,Nsims)
```

isaacsas:
Error for DirectCR:
```julia
signal (11): Segmentation fault: 11
in expression starting at /Users/isaacsas/Desktop/tester.jl:62
getindex at ./array.jl:788 [inlined]
sample at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/aggregators/prioritytable.jl:269 [inlined]
sample at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/aggregators/prioritytable.jl:294
calc_next_jump! at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/aggregators/directcr.jl:151
generate_jumps! at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/aggregators/directcr.jl:140 [inlined]
DirectCRJumpAggregation at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/aggregators/directcr.jl:81 [inlined]
step! at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/SSA_stepper.jl:121
solve! at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/SSA_stepper.jl:35
#__solve#128 at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/SSA_stepper.jl:28 [inlined]
__solve at /Users/isaacsas/.julia/packages/DiffEqJump/GPkfY/src/SSA_stepper.jl:27 [inlined]
#442 at /Users/isaacsas/.julia/packages/DiffEqBase/XoVg5/src/solve.jl:49
[...]

PR #102, opened 2020-05-16
Replace eps(x) with eps(typeof(x))

isaacsas:
Fix for https://github.com/SciML/DiffEqJump.jl/issues/101


isaacsas:
Seems to be crashing on a variable rate issue so I'm going to go ahead and merge.

PR #104, opened 2020-05-17
Alias jumps by default in single threaded use

ChrisRackauckas:
```julia
using DiffEqBase
using DiffEqJump
sr = [1.0,2.0,50.]
maj = MassActionJump(sr,[[1 => 1],[1 => 1],[0 => 1]], [[1 => 1], [1 => -1], [1 => 1]])
params = (1.0,2.0,50.)
tspan = (0.,4.)
u0 = [5]
dprob = DiscreteProblem(u0, tspan, params)
jprob = JumpProblem(dprob, Direct(), maj)
solve(EnsembleProblem(jprob), SSAStepper(), EnsembleThreads(); trajectories=10)
solve(EnsembleProblem(jprob;safetycopy=true), SSAStepper(), EnsembleThreads(); trajectories=10)

# From
@time for i in 1:10000 solve(EnsembleProblem(jprob), SSAStepper(), EnsembleThreads(); trajectories=1) end
1.907643 seconds (7.43 M allocations: 2.153 GiB, 11.64% gc time)
1.938158 seconds (7.42 M allocations: 2.153 GiB, 11.02% gc time)
1.908684 seconds (7.43 M allocations: 2.154 GiB, 11.19% gc time)
1.864572 seconds (7.43 M allocations: 2.154 GiB, 11.04% gc time)

# To
@time for i in 1:10000 solve(EnsembleProblem(jprob), SSAStepper(), EnsembleThreads(); trajectories=1) end
1.862159 seconds (7.06 M allocations: 2.128 GiB, 13.37% gc time)
1.723136 seconds (7.07 M allocations: 2.128 GiB, 12.32% gc time)
1.770109 seconds (7.08 M allocations: 2.130 GiB, 11.39% gc time)
1.632181 seconds (7.07 M allocations: 2.129 GiB, 10.78% gc time)
```

### DIFF
--- a/src/DiffEqJump.jl
+++ b/src/DiffEqJump.jl
@@ -11,7 +11,7 @@ import Base: size, getindex, setindex!, length, similar, show
 import DataStructures: length, update!
 
 import RecursiveArrayTools: recursivecopy!
-using StaticArrays
+using StaticArrays, Base.Threads
 
 abstract type AbstractJump end
 abstract type AbstractAggregatorAlgorithm end
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -59,7 +59,7 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
                          alg::SSAStepper;
                          save_start = true,
                          save_end = true,
-                         seed = seed_multiplier()*rand(UInt64),
+                         seed = nothing,
                          alias_jump = Threads.threadid() == 1,
                          saveat = nothing)
     if !(jump_prob.prob isa DiscreteProblem)
@@ -70,12 +70,18 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
 
     if alias_jump
       cb = jump_prob.jump_callback.discrete_callbacks[1]
+      if seed !== nothing
+          Random.seed!(cb.condition.rng,seed)
+      end
     else
       cb = deepcopy(jump_prob.jump_callback.discrete_callbacks[1])
+      if seed === nothing
+          Random.seed!(cb.condition.rng,seed_multiplier()*rand(UInt64))
+      else
+          Random.seed!(cb.condition.rng,seed)
+      end
     end
 
-
-    Random.seed!(cb.condition.rng,seed)
     prob = jump_prob.prob
 
     if save_start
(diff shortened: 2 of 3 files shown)


isaacsas:
Did one of your timing examples use `safetycopy=true`? The code looks the same in your comment.

ChrisRackauckas:
Both were with safetycopy=false, which is the default. These settings are safe on the default. I have some thread hoisting stuff that reduces this to:

```julia
@time for i in 1:10000 solve(EnsembleProblem(jprob), SSAStepper(), EnsembleThreads(); trajectories=1) end
1.253476 seconds (6.92 M allocations: 2.100 GiB, 15.69% gc time)
1.229384 seconds (6.90 M allocations: 2.099 GiB, 16.34% gc time)
1.236830 seconds (6.89 M allocations: 2.098 GiB, 15.04% gc time)
1.301729 seconds (6.90 M allocations: 2.098 GiB, 14.73% gc time)
```

which will be a DiffEqBase.jl PR. All of these are running the same exact test code, it's just internal changes.

ChrisRackauckas:
No longer reseeds by default in serial mode.

isaacsas:
Thanks!

isaacsas:
Would you be interested in having a parallel rng setup through EnsembleProblem using either separate streams per thread/process or a jump mechanism (depending on rng)? I ask as these are the two most common approaches I've seen used, and understand to have been examined statistically. I'm happy to take a crack at adding this into EnsembleProblem to give a more unified approach at some point, if this would be of interest.

ChrisRackauckas:
Yeah, that would be a good idea.

I also think we should redo split coupling as a form of ensembler. Basically, we should make `EnsembleCPUArray` very efficient (which should be done by the update to KernelAbstractions.jl), and then that's a nice way to represent multi-trajectory problems solved in parallel. We should be adding multi-level Monte Carlo methods as dispatches for ensembling. If that's the direction, then RNG control has to move up one level in order to get the couplings.

isaacsas:
Yeah, that would be great, and the ensemble level seems like a good choice of where to put it.

ChrisRackauckas:
This is part of the reason for the big change last year to make ensembles dispatch: there's a lot of different algorithms for how to ensemble which aren't just about parallelism, and I think one of our next avenues in stochastic problems is to start exploiting those in tandem with parallelism.

PR #110, opened 2020-06-01
RSSACR

Vilin97:
Wrote the RSSA-CR method. It passes the current tests, and so far performs a bit worse than RSSA.


test/trajectory_test.jl
```diff
@@ -0,0 +1,101 @@
+using DiffEqBase, DiffEqJump
```
isaacsas: Is this test adding something over the other tests? If so, it should be added to runtests too, if not remove it.

test/geneexpr_test.jl
```diff
@@ -102,13 +102,13 @@ if dotestmean
...
-#     # exact methods
-#     for alg in SSAalgs
-#         println("Solving with method: ", typeof(alg), ", using SSAStepper")
-#         jump_prob = JumpProblem(prob, alg, majumps, vartojumps_map=spec_to_dep_jumps, jumptovars_map=jump_to_dep_specs)
-#         @btime solve($jump_prob, SSAStepper())
-#     end
-#     println()
-# end
+# benchmark performance
+if dobenchmark
```
isaacsas, line 106: Leave these commented out so we don't need the BenchmarkTools dependency.

test/geneexpr_test.jl
```diff
@@ -3,14 +3,14 @@ using Test, Statistics
 
 # using Plots; plotlyjs()
 doplot = false
-# using BenchmarkTools
-# dobenchmark = false
+using BenchmarkTools
```
isaacsas, line 6: comment out

test/extinction_test.jl
```diff
@@ -22,6 +19,9 @@ dprob = DiscreteProblem(u0,(0.,100.),rates)
 
 algs = DiffEqJump.JUMP_AGGREGATORS
 
+u0 = [10]
+dprob = DiscreteProblem(u0,(0.,10.),rates)
```
isaacsas, line 23: Why did you decrease this to 10? We were getting occasional test failures at 10 before (rare events where it hadn't gone extinct by that time).

ChrisRackauckas:
This is looking great!

isaacsas:
@Vilin97 Is this ok to merge? We can work on optimizing the speed on the BCR benchmark as a follow up, but it looks ok accuracy wise now.

Vilin97:
Yes, okay to merge

isaacsas:
Nice first PR!

PR #111, opened 2020-06-04
save last saveat point too

isaacsas:
When putting in the kwarg passing for MT I noticed that if `saveat` ends before `tspan[2]` we weren't actually saving the final `saveat` point.


isaacsas:
@ChrisRackauckas Just requesting you take a look in case the previous behavior was intentional for a reason I'm not aware of.

ChrisRackauckas:
It is intentional since it matches the other packages.

isaacsas:
Hmm, how do you avoid missing the last `saveat` value then if it is less than `tstop`?

ChrisRackauckas:
Oh I see, oops.

PR #115, opened 2020-06-06
Cleaning SSA code

Vilin97:
Factored out more common functionality in SSAs.


src/aggregators/rssa.jl
```diff
@@ -108,7 +108,46 @@ end
 function execute_jumps!(p::RSSAJumpAggregation, integrator, u, params, t)
     # execute jump
     u = update_state!(p, integrator, u)
+    update_rates!(p, u, params, t)
+    nothing
+end
+
+# calculate the next jump / jump time
+@fastmath function generate_jumps!(p::RSSAJumpAggregation, integrator, u, params, t)
```
isaacsas, line 116: Try not to switch the order of functions. It makes it hard to see the diff.

src/aggregators/ssajump.jl
```diff
@@ -119,16 +115,68 @@ end
...
+        idx = next_jump - num_ma_rates
+        @inbounds p.affects![idx](integrator)
     end
     return integrator.u
-end
\ No newline at end of file
+end
+
+"check if the rate is 0 and if it is, make the next jump time Inf"
+@inline function is_total_rate_zero!(p) :: Bool
```
isaacsas, line 134: Maybe call `nomorejumps` and pass current `sum_rate` as input?

src/aggregators/ssajump.jl
```diff
@@ -119,16 +115,68 @@ end
...
+        p.next_jump_time = convert(typeof(sum_rate), Inf)
+        return true
+    end
+    return false
+end
+
+"perform linear search of r over array. Output element j s.t. array[j-1] < r <= array[j]. Will error if no such r exists"
+@inline function linear_search(array, r)
+    jidx = 1
+    parsum = array[jidx]
```
isaacsas, line 147: `@inbounds`

src/aggregators/ssajump.jl
```diff
@@ -119,16 +115,68 @@ end
...
+    while parsum < r
+        jidx   += 1
+        parsum += array[jidx]
+    end
+    return jidx
+end
+
+"perform rejection sampling test"
+@inline function rejectrx(p, u, jidx, params, t)
```
isaacsas, line 156: Does passing local variables like `crhigh` and such give a difference over using the fields of `p` here?

src/aggregators/rssacr.jl
```diff
@@ -128,23 +128,20 @@ end
...
+    @unpack sum_rate, rt, rng, cur_rate_high = p
     rerl = zero(sum_rate)
-    jidx = 0
-    notdone = true
-    @inbounds while notdone
+    
+    jidx    = sample(rt, cur_rate_high, rng)
+    notdone = rejectrx(p, u, jidx, params, t)
+    rerl   += randexp(rng)
+    @inbounds while rejectrx(p, u, jidx, params, t)
```
isaacsas, line 142: So the first rejection test on 140 is ignored?

src/aggregators/ssajump.jl
```diff
@@ -119,16 +115,68 @@ end
...
+        @inbounds crate = calculate_jump_rate(p,u,params,t,jidx)
+        if crate > zero(crate) && r2 <= crate
+            return false
+        end
+    end
+    return true
+end
+
+"update the jump rate, assuming p.rates is a vector of functions"
+@inline function calculate_jump_rate(p, u, params, t, rx)
```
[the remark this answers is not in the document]
Vilin97: They are only known as local variables everywhere because of this calculation. Once it is factored out, ma_jumps will not appear as a local variable.

src/aggregators/ssajump.jl
Vilin97: I think it's a tradeoff. The case, which lead to the fastest implementation is if the whole SSA is done as one big function, and fields of p are only accessed twice: at the start of the simulation and at the end. The other extreme is having a lot of little functions, making code more modular but requiring fields of p to be accessed more times. I believe in this case it is better to just access the fields of p inside the function instead of passing them around as variables, because the whole reason for the existence of p is the ability to do this. After all, wherever we pass in p, we could just pass in all the variables that p stores. This could potentially be a little bit faster but at the cost of making the code less readable and less modular. For the existing SSAs that use rejectrx, there is only a constant number of times (around 2-3) when rejectrx is called, so the difference cannot be that big. I will change it to work with local variables instead but I think it would defy the purpose of the existence of the aggregator p.

isaacsas: I'm against refactoring at the expense of performance. That's why I suggested profiling the effect of this modularization on the BCR model, to see if it matters. We should use `@btime` to see if accessing the fields makes no difference, in which case we can avoid many local variables. If it does make anything more than a couple percent difference I'd say we should stick with unpacking everything before calling hot-loop methods like `rejectrx`.

PR #118, opened 2020-06-09
Rejection direct

Vilin97:
Wrote a new SSA called rejectionDirect. The only difference with Direct is that it uses rejection sampling to choose the next reaction to sample. This takes O(number of reactions) in the worst case, and can be very fast if the propensities are fairly uniform. For example, this is the fastest method for 1D diffusion without reactions.


src/aggregators/rdirect.jl
```diff
@@ -0,0 +1,138 @@
...
+    sum_rate = p.sum_rate
+    if nomorejumps!(p, sum_rate)
+        return nothing
+    end
+    @unpack rng, cur_rates, max_rate = p
+
+    num_rxs = length(cur_rates)
+    rx = trunc(Integer, rand(rng) * num_rxs)+1
+    while cur_rates[rx] < rand(rng) * max_rate
+        # global num_rejections += 1
```
isaacsas, line 106: Remove comment

src/aggregators/rdirect.jl
```diff
@@ -0,0 +1,138 @@
...
+function update_dependent_rates!(p::RDirectJumpAggregation, u, params, t)
+    @inbounds dep_rxs = p.dep_gr[p.next_jump]
+    @unpack ma_jumps, rates, cur_rates, sum_rate = p
+    @inbounds for rx in dep_rxs
+        sum_rate -= cur_rates[rx]
+        @inbounds new_rate = calculate_jump_rate(ma_jumps, rates, u,params,t,rx)
+        sum_rate += new_rate
+        cur_rates[rx] = new_rate
+        if new_rate > p.max_rate
+            p.max_rate = new_rate
```
isaacsas, line 132: Hmm, so the upper bound is never recalculated from scratch, and always just grows? I’d imagine it could get pretty bad then, like if you have 100000 molecules in the spatial example all started in the first voxel.

src/aggregators/rdirect.jl
```diff
@@ -0,0 +1,138 @@
...
+
+    # handle constant jumps using function wrappers
+    rates, affects! = get_jump_info_fwrappers(u, p, t, constant_jumps)
+
+    build_jump_aggregation(RDirectJumpAggregation, u, p, t, end_time, ma_jumps,
+                          rates, affects!, save_positions, rng; num_specs = length(u), kwargs...)
+end
+
+# set up a new simulation and calculate the first jump / jump time
+function initialize!(p::RDirectJumpAggregation, integrator, u, params, t)
```
isaacsas, line 78: Shouldn’t the max rate be recalculated in here?

src/aggregators/rdirect.jl
isaacsas: Maybe you should keep a counter, and periodically regenerate the max rate if it hasn’t increased for a certain number of calls?

src/aggregators/rdirect.jl
```diff
@@ -0,0 +1,136 @@
...
+    p.next_jump_time = t + randexp(p.rng) / sum_rate
+    nothing
+end
+
+
+######################## SSA specific helper routines #########################
+function update_dependent_rates!(p::RDirectJumpAggregation, u, params, t)
+    @inbounds dep_rxs = p.dep_gr[p.next_jump]
+    @unpack ma_jumps, rates, cur_rates, sum_rate = p
+    need_to_reculculate_max_rate = false
```
isaacsas, line 118: `recalculate`

src/aggregators/rdirect.jl
```diff
@@ -0,0 +1,136 @@
...
+    # if no more events possible there is nothing to do
+    sum_rate = p.sum_rate
+    if nomorejumps!(p, sum_rate)
+        return nothing
+    end
+    @unpack rng, cur_rates, max_rate = p
+
+    num_rxs = length(cur_rates)
+    rx = trunc(Integer, rand(rng) * num_rxs)+1
+    while cur_rates[rx] < rand(rng) * max_rate
```
isaacsas, line 103: `@inbounds`

src/aggregators/rdirect.jl
isaacsas, line 118: `need_to_reculculate` -> `need_to_recalculate`

src/aggregators/rdirect.jl
```diff
@@ -0,0 +1,136 @@
...
+    @inbounds dep_rxs = p.dep_gr[p.next_jump]
+    @unpack ma_jumps, rates, cur_rates, sum_rate = p
+    need_to_reculculate_max_rate = false
+    @inbounds for rx in dep_rxs
+        sum_rate -= cur_rates[rx]
+        @inbounds new_rate = calculate_jump_rate(ma_jumps, rates, u,params,t,rx)
+        sum_rate += new_rate
+        if new_rate > p.max_rate
+            p.max_rate = new_rate
+        elseif abs(cur_rates[rx]-p.max_rate) < eps(typeof(p.max_rate))
```
isaacsas, line 125: What is the point of this check? Isn't `need_to_reculculate_max_rate` already `false`? In fact, do you ever set it to `true` in this function?

src/aggregators/rssa.jl
```diff
@@ -119,10 +117,8 @@ end
     if nomorejumps!(p, sum_rate)
         return nothing
     end
-
     # next jump type
     @unpack ma_jumps, rates, cur_rate_high, cur_rate_low, rng = p
-    #rerl        = one(sum_rate)
```
isaacsas, line 125: Please don't remove the commented out lines in the RSSA file here and below. This is something we want to be able to retest periodically, and it is easier to just leave the three commented lines...

src/aggregators/rdirect.jl
Vilin97: It was supposed to be `need_to_recalculate_max_rate = true`. The idea is that if the rate that is equal to the current max_rate decreased, then we need to recalculate max_rate. This keeps max_rate always equal to the current maximum rate.

src/aggregators/rssa.jl
Vilin97: Got it

src/aggregators/rdirect.jl
isaacsas: Can't you just do `==` here, since they really should be equal?

Issue #119, opened 2020-06-12
Aggregated jump callback resetting


ChrisRackauckas:
It would be nice to be able to "reset", i.e. rebuild the rates and take new random numbers. That is the blocker on https://github.com/SciML/DiffEqJump.jl/issues/78 and https://github.com/PumasAI/Pumas.jl/pull/472 . I am envisioning a function like `reset_jumps!(cb)` which does nothing on a non-JumpCallback (so that we can apply it to the full `CallbackSet`) but then regenerates everything, so that if a `DiscreteCallback` or `ContinuousCallback` is used to perturb the system we can `reset_jumps!(integrator.callback)`.

@isaacsas I think this would have to be implemented per-aggregator, but could have a generic fallback?

isaacsas:
An inefficient way would be to just call:
https://github.com/SciML/DiffEqJump.jl/blob/ee225e7f9773b2d87734179c245fed77ba331b92/src/aggregators/direct.jl#L35
I think.

Knowing which state variables have changed would allow us to much more efficiently update some methods.

isaacsas:
We should probably move the time generation call out of that and into `initialize!`, then the latter could be called.

ChrisRackauckas:
I think that function probably works. I can probably whip something up in a bit. Dependency analysis on a callback might be tricky though, but we could make a scheme for that later.

isaacsas:
It would just be looking at `u` vs `hold` and seeing which entries have changed.

isaacsas:
`uold` I mean.

Issue #78, opened 2019-10-19
Add discrete callbacks and tstops to SSAStepper


isaacsas:
Recently requested by a collaborator who wants to be able to simulate cell division in SSA models  at predetermined times.

ChrisRackauckas:
We'd have to time it. Tstops is one of the things it's avoiding so that it doesn't have to check a heap. At least a "fastest version of a stepper that is a bit less restrictive" can be made.

isaacsas:
Why do we need a heap? Is that to allow callbacks to dynamically add additional Tstops during a simulation? If we don't allow injecting additional Tstops, couldn't we just sort the initial Tstop vector and progress through it iteratively? Such a check would be one additional comparison per step, which for anything with a few jumps shouldn't be the dominant cost. I'd think the bigger issue is checking all the callbacks each iteration to see if they fired off (with the current callback setup).

Maybe we want to have a type of callback that is set to (one or more) fixed, user given times times, and a general termination condition call back? This would be a reduced set of functionality, but could cover many use cases I imagine.

isaacsas:
Actually, I’m not really sure why checking the top of a heap would be an issue either. That’s essentially what the NRM does every step, and certainly isn’t the dominant cost for it that makes it generally slower than Direct for simple systems.

ChrisRackauckas:
>Actually, I’m not really sure why checking the top of a heap would be an issue either. That’s essentially what the NRM does every step, and certainly isn’t the dominant cost for it that makes it generally slower than Direct for simple systems.

It used to be a bigger deal, but it doesn't show up in as many profiles these days. It's worth checking. I would start from

https://github.com/JuliaDiffEq/SimpleDiffEq.jl/blob/master/src/functionmap.jl

and add in the pieces for SSA, and see how it compares to SSAStepper. If it's just as fast, then that could be the new backing to the SSAStepper, and then adding callbacks is simple.

isaacsas:
Ok, thanks for the pointer. I have to familiarize myself better with the integrator interface as I haven’t used it before. This is definitely something I’d like to take a look at once I get past those mid-November deadlines I mentioned.

isaacsas:
There's also another issue we really should address longer term -- we need an interface to update the SSAs internally so their state is consistent after a callback runs. With the exception of `Direct`, most have internal state that would need to be reset after changes to `u`.

ChrisRackauckas:
[...]

PR #120, opened 2020-06-12
implement callback resets and DiscreteCallbacks in SSAStepper

ChrisRackauckas:
Right now it requires that the user knows to call `reset_aggregated_jumps!`. This can be fixed by hoisting this call all the way up to DiffEqBase and into `apply_discrete_callback!`, which is a bit of work so I'm punting on that for now but will open an issue.

Fixes https://github.com/SciML/DiffEqJump.jl/issues/119 and mostly fixes https://github.com/SciML/DiffEqJump.jl/issues/78 . For the latter, you cannot add `tstops!` on the fly in a DiscreteCallback, so it doesn't fully conform to the interface, but it's close.

### DIFF
--- a/src/DiffEqJump.jl
+++ b/src/DiffEqJump.jl
@@ -37,6 +37,7 @@ include("aggregators/rssa.jl")
 include("aggregators/prioritytable.jl")
 include("aggregators/directcr.jl")
 include("aggregators/rssacr.jl")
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
Introducing epidemic models with purges.

plot

Issue #124, opened 2020-06-17
saveat not working


isaacsas:
`saveat` seems to be returning wrong values for the given times now. Here is a simple example:

```julia
using DiffEqBiological, DiffEqBase, DiffEqJump
rn = @reaction_network begin
    kp, A + B --> C
end kp
n0 = [1,1,0]
tspan = (0,.2)
dprob = DiscreteProblem(rn, n0, tspan, [10.0])
jprob = JumpProblem(dprob, Direct(), rn, save_positions=(false,false))
ts = collect(0:.002:tspan[2])
Nsims = 1000
NA = zeros(length(ts))
for i = 1:Nsims
    sol = DiffEqJump.solve(jprob, SSAStepper(), saveat=saveat)
    NA .+= sol[1,:]
end
NA ./= Nsims
```
In particular, `NA[2]` is about 0.123 whereas the analytical value should be `exp(-10*ts[2])=0.9801986733067553`
In contrast, saving all event times and using the builtin interpolation seems to work fine:
```julia
jprob = JumpProblem(dprob, Direct(), rn)
NA = zeros(length(ts))
for i = 1:Nsims
    sol = DiffEqJump.solve(jprob, SSAStepper())
    for n = 1:length(ts)
        NA[n] += sol(ts[n])[1]
    end
end
NA ./= Nsims
```
Here, up to sampling error, NA agrees well with the analytical formula.

isaacsas:
@ChrisRackauckas is `savevalues!` here
https://github.com/SciML/DiffEqJump.jl/blob/76451d5f0a0a3626edae44c25398bf25a07feee6/src/SSA_stepper.jl#L163
saving the right value (i.e. the pre-step value)?

ChrisRackauckas:
The two need to swap: the saveat part should come before the save_everystep part, otherwise the array might not be sorted.

ChrisRackauckas:
You're right: https://github.com/SciML/DiffEqJump.jl/blob/76451d5f0a0a3626edae44c25398bf25a07feee6/src/SSA_stepper.jl#L177 should be `integrator.uprev`

isaacsas:
OK, I can try to add a fix when I get some time in the next day or so. Bit busy with getting stuff done for meetings the next few days.

PR #125, opened 2020-06-17
fix SSAStepper saveat

ChrisRackauckas:
Fixes https://github.com/SciML/DiffEqJump.jl/issues/124


isaacsas:
Always good not to make Travis run out of memory. Thanks for looking into this so quick!

Issue #126, opened 2020-07-02
Memory allocation in RSSACR


Vilin97:
The memory allocation in the current implementation of RSSACR is abnormal. On small problems the allocations seem to be quite comparable to RSSA and DirectCR but on bigger problems the allocations of RSSACR skyrocket. For example, we have 
```3799.138189 seconds (9.30 G allocations: 138.623 GiB, 0.39% gc time)```
for RSSACR vs
```12118.556890 seconds (521.16 k allocations: 29.947 MiB)```
for RSSA on 1D CRDME of reaction network A+B <--> C. 
From my investigations, it looks like RSSACR always allocates about 16 bytes per step of the simulation (the ```p(integrator)``` functor), but where exactly it happens is unclear. Specifically, ```@allocated p(integrator)``` gives 0 for DirectCR, and 16 for RSSACR. However, when I look at the allocations of ```execute_jumps!``` and ```generate_jumps!```, they all allocate, for RSSACR and for DirectCR. 
Additionally, ```@code_warntype``` has not revealed any type instabilities for any functions of RSSACR. My investigating also showed that the allocations do not come from the initialization of the data structures. 
In my investigations it came up that ```generate_jumps!``` allocates 32 bytes for RSSACR and 16 bytes for DirectCR using the ```@allocate``` macro. Perhaps, the issue lies there. 

The source of the allocations needs to be found and fixed.

isaacsas:
Yeah, I was seeing this in the CRDME example you mention. This would be good to straighten out. Performance already seems very good, but it would be great if fixing this up improves performance even more.

ChrisRackauckas:
Those two PRs should do a bit.

isaacsas:
@Vilin97 Let's check where allocations are at now with Chris' fixes.

Vilin97:
I will repeat my test tomorrow morning and let you know

Vilin97:
This solved the memory allocations for me. Looks like this was indeed the issue. Thank you, Chris!

PR #127, opened 2020-07-02
Strictly type RSSACRJumpAggregation

ChrisRackauckas:
Should do at least part of https://github.com/SciML/DiffEqJump.jl/issues/126


isaacsas:
Thanks! Vasily and I both missed that when going over his PRs...

PR #133, opened 2020-08-07
Add tprev field to SSAIntegrator

### LINKED ISSUE #123: Optimization of RejectionDirect
Rejection Direct is a method, which replaces the linear search for the next reaction in Direct with rejection sampling. In order to do this, the variable `max_rate` is stored in the aggregator. In order for the method to work correctly, `max_rate` must always be at least as big as the highest propensity over all reactions. At the same time, the smaller `max_rate` is, the fewer rejections there will be, and the faster the method will perform. The challenge is to maintain `max_rate` to have a good balance between these two. It is clear when to increase it: whenever any propensity gets bigger than `max_rate`. The question of when to decrease it is uncertain. 
There are two extremes: if `max_rate` is updated every time a reaction fires (which is close to what is being done now, although not quite), too much time (linear in number of reactions) will be spent on this update. If it is never updated, the method will perform poorly (e.g. in the extinction network, where one species is going extinct). An approach suggested by @isaacsas is to keep a counter and update `max_rate` every `update_frequency` steps, where `update_frequency` is some heuristic parameter (or perhaps is somehow inferred from the network at the initialization step). 
A good strategy needs to be developed and benchmarks on networks with different propensity distributions need to be done.

### LINKED ISSUE #132: Keep track of tprev and uprev in SSAStepper (for use in callbacks)
The [integrator interface](https://diffeq.sciml.ai/dev/basics/integrator/) specifies that an integrator has the fields `tprev` and `uprev` that hold the values of the last time point and last solution value, respectively. I see no reason why `SSAStepper` should not have these fields.

manuel-rhdt:
Fixes #132.

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -7,6 +7,7 @@ mutable struct SSAIntegrator{F,uType,tType,P,S,CB,SA,OPT,TS} <: DiffEqBase.DEInt
     f::F
     u::uType
     t::tType
+    tprev::tType
     p::P
     sol::S
     i::Int
@@ -128,7 +129,7 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
      sizehint!(t,2)
    end
 
-    integrator = SSAIntegrator(prob.f,copy(prob.u0),prob.tspan[1],prob.p,
+    integrator = SSAIntegrator(prob.f,copy(prob.u0),prob.tspan[1],prob.tspan[1],prob.p,
                                sol,1,prob.tspan[1],
                                cb,_saveat,save_everystep,save_end,cur_saveat,
                                opts,tstops,1,false)
@@ -139,6 +140,8 @@ end
 DiffEqBase.add_tstop!(integrator::SSAIntegrator,tstop) = integrator.tstop = tstop
 
 function DiffEqBase.step!(integrator::SSAIntegrator)
+    integrator.tprev = integrator.t
+
     doaffect = false
     if !isempty(integrator.tstops) &&
         integrator.tstops_idx <= length(integrator.tstops) &&


src/SSA_stepper.jl
```diff
@@ -139,6 +141,11 @@ end
 DiffEqBase.add_tstop!(integrator::SSAIntegrator,tstop) = integrator.tstop = tstop
 
 function DiffEqBase.step!(integrator::SSAIntegrator)
+    integrator.tprev = integrator.t
+    if integrator.uprev != integrator.u
+        copy!(integrator.uprev, integrator.u)
```
ChrisRackauckas, line 146: Yeah, I was just worried about the cost of this operation, since it's O(n) and would then effect some O(1) methods.

manuel-rhdt: I see. Do you have a suggestion for what to do about this?

manuel-rhdt: I wonder if we can set `uprev` to be a reference to the most recently saved solution value. The behaviour will only be fully correct if `save_everystep` is `true` however...

isaacsas: Oh yeah, that's definitely an issue we want to avoid. This would have to be updated analogous to `u` being updated within the SSA. I guess it could be done by storing the preceding jump index in the aggregator, and then applying it to `uprev` when `u` is updated with the current jump through `affect!`. I don't think this update is usually rate limiting (maybe for `Direct` if there are very few jumps).

isaacsas: Did you want to use `uprev` for something?

manuel-rhdt: I would like to compute the probability density of the generated trajectory alongside the SSA. Thus in a callback I need to check which reaction just happened. I do not actually need `uprev` but would like to be able to access either the index of the most recently selected reaction or the jump vector `du=u-uprev`.

manuel-rhdt: Maybe a better idea for my use case is to record the index of the most recently performed jump.

isaacsas: We could add a `prev_jump` field to the aggregators to store this. I think you should be able to then access this through the integrator, but I need to double check what everything stores...

isaacsas: This is where the state gets updated and we could then save the old value:

https://github.com/SciML/DiffEqJump.jl/blob/80fdaa6e6abffdfccb1a7624ddd69c83c4cb2b82/src/aggregators/ssajump.jl#L115

isaacsas: (By old value I mean the index of the jump that was just executed.)

manuel-rhdt: That sounds very reasonable. I still would like to be able to access tprev, so I‘ll change the PR to only include that part.

Thank you!

isaacsas: @ChrisRackauckas Can correct me if I've got this wrong, but it appears you can access the underlying JumpAggregation that would store such info through ` integrator.cb.affect!`. I can add a PR for that in the next day or so. We just need to update each of the eight SSAs in the aggregator folder so that their `JumpAggregation` has the `prev_jump` field, and then set that field within `update_state!`.

ChrisRackauckas: yeah that should work

manuel-rhdt:
I changed the PR to only add tprev. This should incur no performance penalty since `t` is only a scalar value.

It will thus be possible to write callbacks that take the length of the most recent tilmestep into account.

isaacsas:
@manuel-rhdt Should have another PR with the `prev_jump` field added shortly.

PR #144, opened 2020-08-18
Revert "CompatHelper: bump compat for "DataStructures" to "0.18""

isaacsas:
Reverts SciML/DiffEqJump.jl#143


isaacsas:
@ChrisRackauckas Having not reverted before I'll let you handle if this is the right way to undo that. FWIW it look like MT is what prevents me from locally updating to it.

ChrisRackauckas:
There's a lot preventing it. This will take like a month to propagate though.

PR #147, opened 2020-08-18
Autogenerate dependency graphs for mass action jumps

isaacsas:
When there are no ConstantRateJumps.


isaacsas:
Yeah, it is helpful for the postdoc working with me on CRDME models of T cell signaling. We're just going to bypass MT (for now) due to the JumpProblem issues, but I'd still like things to be as automatic for him as possible since his interests are primarily in the modeling work.

I'd still like to eventually get some sort of introspection of ConstantRateJumps to try to auto-build dependency graphs for them too. It seems like it should work similar to the tooling for auto-detecting sparsity structures or perhaps like ModelingToolkitize.

Issue #148, opened 2020-08-19
handling extinction for SSAs that use a running and updating sum_rate


isaacsas:
RSSA, RSSACR, DirectCR and SortingDirect all encounter the problem of the rates actually being zero but the current sum of the rates being a bit bigger than zero due to floating point errors in the running sum. This led to crashes for the first three when trying to sample the next reaction in a spatial A+B -> 0 system (SortingDirect ends up just sampling the last reaction, but this shouldn't matter since no further reaction should occur).

I'll put in a PR with a fix, but this is worth investigating more to make sure the fix doesn't hurt performance, and isn't missing any further cases.

Vilin97:
Instead of checking if `sum_rate` is zero we can check that the current state vector is the zero vector. This is, of course, linear in the number of species, but the number of species is usually considerably smaller than the number of reactions. This can be improved by adding a field `num_molecules` to the aggregator to update the total number of molecules left, s.t. the check can be done in constant time. A problem with this approach is that there can be rate functions that are zero while there are some molecules still present.

Another idea would be to compare `sum_rate` with some no-so-small value, e.g. 10^-6, instead of `eps(sum_rate)`, and if `sum_rate < 10^-6`, then actually recalculate `sum_rate`. The hope is that floating point error will never accumulate to the point of exceeding 10^-6 or whatever value is chosen, so if `sum_rate > 10^-6`, then it is definitely non-zero. On the other hand, the event that `sum_rate < 10^-6` is rare enough that recalculating `sum_rate` will happen very infrequently.

isaacsas:
A first fix is in https://github.com/SciML/DiffEqJump.jl/pull/149, see also comments there.

isaacsas:
From Slack: we could try to make nomorejumps! more robust though. perhaps by calculating the mean time to the next reaction (which we can do as 1/sum_rate) and if it is way beyond end time we end. (ie. the probability of a jump with that mean by end time is below some tolerance).

PR #149, opened 2020-08-19
Crashing extinction test

isaacsas:
This seems to fix https://github.com/SciML/DiffEqJump.jl/issues/148.

Note, to try to minimize extra branches in hot loops this assumes that for 
1. SortingDirect and DirectCR the issue can be detected by simply checking if the next reaction time is greater than the simulation end time before calculating the next jump (and in that case the next jump is not calculated).
2. RSSA and RSSACR that if this issue crops up it will appear in the first search for the next reaction, so a check is only needed after it. There I now explicitly set the next reaction time to `Inf` if the search failed and return. (But perhaps we should instead recalculate the sum in this case to make sure there are not other issues, and if it isn't zero try sampling again?)

@Vilin97 mind taking a look? Any thoughts?

I also added some more refactoring I noticed while working on this.

### DIFF
--- a/src/aggregators/rssa.jl
+++ b/src/aggregators/rssa.jl
@@ -98,7 +98,7 @@ function execute_jumps!(p::RSSAJumpAggregation, integrator, u, params, t)
 end
 
 # calculate the next jump / jump time
-@fastmath function generate_jumps!(p::RSSAJumpAggregation, integrator, u, params, t)
+function generate_jumps!(p::RSSAJumpAggregation, integrator, u, params, t)
     sum_rate = p.sum_rate
     # if no more events possible there is nothing to do
     if nomorejumps!(p, sum_rate)
--- a/src/aggregators/rssa.jl
+++ b/src/aggregators/rssa.jl
@@ -107,29 +107,33 @@ end
     # next jump type
     @unpack ma_jumps, rates, cur_rate_high, cur_rate_low, rng = p
     num_majumps = get_num_majumps(ma_jumps)
-    #rerl        = one(sum_rate)
     rerl        = zero(sum_rate)
 
-    r      = rand(rng) * sum_rate
-    jidx   = linear_search(cur_rate_high, r)
-    rerl  += randexp(rng)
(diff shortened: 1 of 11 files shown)


src/aggregators/rssa.jl
```diff
@@ -98,35 +98,43 @@ function execute_jumps!(p::RSSAJumpAggregation, integrator, u, params, t)
...
+    num_majumps = get_num_majumps(ma_jumps)
     rerl        = zero(sum_rate)
 
-    r      = rand(rng) * sum_rate
-    jidx   = linear_search(cur_rate_high, r)
-    rerl  += randexp(rng)
-    @inbounds while rejectrx(ma_jumps, rates, cur_rate_high, cur_rate_low, rng, u, jidx, params, t)
+    r    = rand(rng) * sum_rate
+    jidx = linear_search(cur_rate_high, r)
+    if iszero(jidx)
```
isaacsas, line 114: We might want to update this to recalculate `sum_rate` and if it is zero return as it currently does, and if it is not zero redo the search with an updated `r`. This could mitigate some potential floating point errors when the running sum is larger than the actual total propensity (which could also lead to getting a zero index). I'm going to leave that for a next PR though since I want to think more about how to handle this.

src/aggregators/rssacr.jl
```diff
@@ -130,14 +130,19 @@ function generate_jumps!(p::RSSACRJumpAggregation, integrator, u, params, t)
...
 
     @unpack rt, ma_jumps, rates, cur_rate_high, cur_rate_low, rng = p
+    num_majumps = get_num_majumps(ma_jumps)
     rerl = zero(sum_rate)
 
-    jidx    = sample(rt, cur_rate_high, rng)
-    rerl   += randexp(rng)
-    while rejectrx(ma_jumps, rates, cur_rate_high, cur_rate_low, rng, u, jidx, params, t)
+    jidx = sample(rt, cur_rate_high, rng)
+    if iszero(jidx)
```
isaacsas, line 137: Same comment.

PR #151, opened 2020-08-20
Make dependency graphs more flexible

isaacsas:
This converts dependency graph generation to only use vectors. Since the typical case where we'd expect the construction of such graphs to be slow is very large systems, which are usually sparse, this should perform better (and in benchmarking it is 2.5 to 6 times faster for spatial systems with similar reductions in memory use).

This also gives a template for how the MT dependency graph code should be updated for JumpSystems.


isaacsas:
Hung up till https://github.com/SciML/OrdinaryDiffEq.jl/pull/1250 is finalized.

PR #159, opened 2020-12-18
Fix reset_aggregated_jumps! which is currently a no-op

manuel-rhdt:
Before this commit a call to `reset_aggregated_jumps!(integrator)` has no effect at all. Note that the jump aggregator is stored in the `integrator.cb` field and not in `integrator.opts.callback` (the latter is used to store additional user-provided callbacks). I suppose this commit fixes the code to do what it was intended to do.

### DIFF
--- a/src/aggregators/aggregated_api.jl
+++ b/src/aggregators/aggregated_api.jl
@@ -1,5 +1,5 @@
 function reset_aggregated_jumps!(integrator,uprev = nothing)
-     reset_aggregated_jumps!(integrator,uprev,integrator.cb)
+     reset_aggregated_jumps!(integrator,uprev,integrator.opts.callback)
      nothing
 end
 
--- a/src/aggregators/aggregated_api.jl
+++ b/src/aggregators/aggregated_api.jl
@@ -1,5 +1,5 @@
 function reset_aggregated_jumps!(integrator,uprev = nothing)
-     reset_aggregated_jumps!(integrator,uprev,integrator.opts.callback)
+     reset_aggregated_jumps!(integrator,uprev,integrator.cb)
      nothing
 end
 


manuel-rhdt:
Judging from the test failure, my proposed fix is wrong. I guess the problem is that when using SSAIntegrator, the jump aggregator is not put into `integrator.opts.callbacks` as it is for other DiffEq integrators. Therefore `reset_aggregated_jumps!(integrator)` does not work if `integrator <: SSAIntegrator`. I would gladly like to fix this issue properly but would enjoy some suggestions of how to best proceed.

src/aggregators/aggregated_api.jl
```diff
@@ -1,5 +1,5 @@
 function reset_aggregated_jumps!(integrator,uprev = nothing)
-     reset_aggregated_jumps!(integrator,uprev,integrator.opts.callback)
+     reset_aggregated_jumps!(integrator,uprev,integrator.cb)
```
ChrisRackauckas, line 2: This isn't correct. It's in opts.callback, and it's not a no-op if there's jumps

ChrisRackauckas:
Why would you say it's a no-op? It triggers a reset of a jump.

manuel-rhdt:
Sorry for the confusion. I'll explain what my problem is. I create a pure-jump problem using Catalyst.jl using only mass action jumps to simulate a chemical reaction network. I solve it using the direct SSA implementation in `SSAIntegrator`. There I notice that changing the copy numbers inside a user callback does not recompute the reaction rates correctly, even when I call `reset_aggregated_jumps!` at the end of the callback. Using the debugger I can check that `reset_aggregated_jumps!` is essentially a no-op in that case.

As far as I can tell where the jump-aggregator is stored depends on how you set up the jump problem. For `SSAIntegrator` the jump callback is stored in the field named `cb`. See the init function where the callback from the jump problem is copied to the field `integrator.cb` and the user callback is copied to the field `integrator.opts.callback` https://github.com/SciML/DiffEqJump.jl/blob/fe0d21d8497b50eb3453d7134ce4a14236ed937a/src/SSA_stepper.jl#L65

My guess is that the jump callback should also be stored in `integrator.opts.callback` for `SSAIntegrator`. Do you agree @ChrisRackauckas ?

isaacsas:
@manuel-rhdt what happens if you add a dispatch only for SSAIntegrator?

isaacsas:
@ChrisRackauckas Or maybe in https://github.com/SciML/DiffEqJump.jl/blob/bbc9fc5d06e420f6da1f992b95004e19edf62285/src/aggregators/aggregated_api.jl#L1

there should be a check if we have an SSAIntegrator that then also calls reset on the `cb` field?

isaacsas:
@manuel-rhdt in your user code have you tried calling this version https://github.com/SciML/DiffEqJump.jl/blob/fe0d21d8497b50eb3453d7134ce4a14236ed937a/src/aggregators/aggregated_api.jl#L18
Passing both the fields you mention? Does it work?

manuel-rhdt:
> @manuel-rhdt in your user code have you tried calling this version
> 
> https://github.com/SciML/DiffEqJump.jl/blob/fe0d21d8497b50eb3453d7134ce4a14236ed937a/src/aggregators/aggregated_api.jl#L18
> 
> 
> Passing both the fields you mention? Does it work?

Hi, this is my current workaround which seems to work correctly. I feel that this is not the intended usage, though. Is there a good reason for having a separate `cb` field in `SSAIntegrator`?

isaacsas:
I’m not sure. Maybe @ChrisRackauckas can comment on that. Maybe because it is called every step of the time-stepper for SSAIntegrator, so that (potentially) saves several pointer accesses?

ChrisRackauckas:
Supersceded by https://github.com/SciML/DiffEqJump.jl/pull/161 which explains what's going on. It would be really nice if a test case could be donated for this.

ChrisRackauckas:
>I’m not sure. Maybe @ChrisRackauckas can comment on that. Maybe because it is called every step of the time-stepper for SSAIntegrator, so that (potentially) saves several pointer accesses?

It's moreso that we never cared to setup SSAIntegrator with an `opts` that has the abstol, reltol, etc. since they all don't apply to SSAs, so it ends up kind of being a special case. The easiest solution is to just throw an overload on this one function and it should all be functional.

PR #161, opened 2020-12-21
Fix reset_aggregated_jumps! which is currently a no-op on SSAIntegrator

ChrisRackauckas:
Fixed version of https://github.com/SciML/DiffEqJump.jl/pull/159 . The issue is that SSAIntegrator doesn't conform all the way to the standard integrator interface, which uses integrator.opts.callback. But it doesn't need to conform to everything around continuous tolerances and all of that since it's really a weird object, so this gets a special case for the jump reset that handles the issue.


isaacsas:
Will other discrete callbacks get called now though, or should you use the three-argument form?

isaacsas:
Sorry, mean the four-argument version...

ChrisRackauckas:
This should be sufficient. It just drops the callback set into the arguments to then recurse over it. Hard to tell without a test though, and hard to find a good test for. I'll merge for now since it's better than before, but it really needs some way to test.

PR #162, opened 2020-12-22
SSAIntegrator callback test

manuel-rhdt:
This pull request adds a test to validate the changes made in #161. While writing the test I noticed that there remained some subtle problems in `SSAIntegrator` when the system is "extinct" but there are still tstops to be visited. In that case the integrator previously ignored the tstops. This is wrong since a `DiscreteCallback` can modify `integrator.u` at such a future tstop such that the system is no longer extinct.

~~This PR does slightly change the behaviour of `DiffEqBase.step!` for `SSAIntegrator`. Notably the integrator can no longer make steps after the end of the problem `tspan`. I do not know whether the previous behaviour was more desirable.~~

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -38,8 +38,7 @@ end
 function DiffEqBase.solve!(integrator)
 
     end_time = integrator.sol.prob.tspan[2]
-
-    while integrator.keep_stepping && (integrator.t < end_time) # It stops before adding a tstop over
+    while should_continue_solve(integrator) # It stops before adding a tstop over
         step!(integrator)
     end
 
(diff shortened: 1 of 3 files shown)


src/SSA_stepper.jl
```diff
@@ -145,15 +146,22 @@ DiffEqBase.add_tstop!(integrator::SSAIntegrator,tstop) = integrator.tstop = tsto
...
         integrator.tstops_idx <= length(integrator.tstops) &&
-        integrator.tstops[integrator.tstops_idx] < integrator.tstop
+        integrator.tstops[integrator.tstops_idx] < next_jump_time
 
         integrator.t = integrator.tstops[integrator.tstops_idx]
         integrator.tstops_idx += 1
     else
-        integrator.t = integrator.tstop
+        integrator.t = next_jump_time
+        if integrator.t >= end_time
```
ChrisRackauckas, line 161: this isn't correct. If you call `step!` while at `end_time` it should keep stepping. You should able to use `step!` to step indefinitely. Basically if you keep calling `step!` it should do the same thing as `solve!` until the integration is done. Then you can keep `step!`ing beyond that just making sure to hit tstops.

manuel-rhdt: I see! I added a comment that addresses this concern.

isaacsas:
Thanks @manuel-rhdt for catching that and the fix!

PR #163, opened 2020-12-27
Fix out of bounds error for SSAStepper's DiffEqBase.__init

manuel-rhdt:
In `SSAStepper`'s `DiffEqBase.__init` function, when the kwarg `save_start=false` a local variable `u` gets initialized as an empty array. Later in that function `u[1]` is accessed leading to an error. This is fixed in this pr.

Additionally make `sizehint!` more accurate in the edge case where `save_everystep==false` and `isempty(saveat)`. This is
particularly helpful in situations where one frequently remakes JumpProblems and has to avoid allocations for every `DiffEqBase.solve` call.

Some tests were added to avoid future regressions.

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -136,7 +136,7 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
                                sol,1,prob.tspan[1],
                                cb,_saveat,save_everystep,save_end,cur_saveat,
                                opts,tstops,1,false,true)
-    cb.initialize(cb,copy(prob.u0),prob.tspan[1],integrator)
+    cb.initialize(cb,integrator.u,prob.tspan[1],integrator)
     integrator
 end
 
@@ -224,5 +224,4 @@ function DiffEqBase.terminate!(integrator::SSAIntegrator, retcode = :Terminated)
 end
 
 export SSAStepper
-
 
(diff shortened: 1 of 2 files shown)


src/SSA_stepper.jl
```diff
@@ -128,15 +128,15 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
...
+     sizehint!(u,save_start+save_end)
+     sizehint!(t,save_start+save_end)
    end
 
     integrator = SSAIntegrator(prob.f,copy(prob.u0),prob.tspan[1],prob.tspan[1],prob.p,
                                sol,1,prob.tspan[1],
                                cb,_saveat,save_everystep,save_end,cur_saveat,
                                opts,tstops,1,false,true)
-    cb.initialize(cb,u[1],prob.tspan[1],integrator)
+    cb.initialize(cb,copy(prob.u0),prob.tspan[1],integrator)
```
isaacsas, line 139: Do we even need the copy here or can just pass `prob.u0`? @ChrisRackauckas is there anything in the wrapping `DiscreteCallback` that would require a copy?

isaacsas: Yeah, so this should be calling into
https://github.com/SciML/DiffEqJump.jl/blob/784cddb360167f6b9b808af444fb80422b8023da/src/aggregators/ssajump.jl#L48

There `initialize!` shouldn't modify `u`, so I think it should be safe to pass `prob.u0` directly.

ChrisRackauckas: it can use `integrator.u` which already does the copy.

ChrisRackauckas, line 139:
```suggestion
    cb.initialize(cb,integrator.u,prob.tspan[1],integrator)
```

PR #166, opened 2021-01-07
Call initializer and finalizer for user callback in SSAIntegrator

manuel-rhdt:
While `SSAIntegrator` can handle user callbacks in addition to the jump callbacks we are still missing some features from other solvers of the DiffEq ecosystem. This PR's intention is to add some of those features.

1. With this PR we call the initializer and finalizer functions of the user-provided callback as that is what the user will expect to happen.
2. Users may want to call `DiffEqBase.add_tstop!` inside a callback to dynamically add additional stopping points to `SSAIntegrator` during solve. Since `SSAIntegrator` does support `tstops` it makes sense to allow this. Note `add_tstop!` has to ensure that the `tstops` list always remains sorted. To avoid this extra work in the jump aggregators I added a specialized method `register_next_jump_time!` for `SSAIntegrator` that only updates one field. This should ensure that there are no performance regressions.

With these changes it is also possible to use `SSAIntegrators` with things like `PresetTimeCallback` from the callback library which adds `tstops` on callback initialize.

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -76,7 +76,7 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
                          alias_jump = Threads.threadid() == 1,
                          saveat = nothing,
                          callback = nothing,
-                         tstops = (),
+                         tstops = [],
                          numsteps_hint=100)
     if !(jump_prob.prob isa DiscreteProblem)
         error("SSAStepper only supports DiscreteProblems.")
@@ -147,7 +147,7 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
     integrator
 end
 
-function DiffEqBase.add_tstop!(integrator::SSAIntegrator{<:Any,<:Any,<:Any,<:Any,<:Any,<:Any,<:Any,<:Any,<:Vector},tstop)
+function DiffEqBase.add_tstop!(integrator::SSAIntegrator,tstop)
     insert_index = searchsortedfirst(integrator.tstops, tstop)
     if insert_index >= integrator.tstops_idx
         # we only insert the tstop if insert_index >= integrator.tstops_idx and thus
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -148,10 +148,9 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
 end
 
 function DiffEqBase.add_tstop!(integrator::SSAIntegrator,tstop)
-    insert_index = searchsortedfirst(integrator.tstops, tstop)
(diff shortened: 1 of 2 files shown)


isaacsas:
@manuel-rhdt Thanks! I'll take a look tonight (unless @ChrisRackauckas looks it over earlier).

src/SSA_stepper.jl
```diff
@@ -141,10 +143,25 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
                                cb,_saveat,save_everystep,save_end,cur_saveat,
                                opts,tstops,1,false,true)
     cb.initialize(cb,integrator.u,prob.tspan[1],integrator)
+    DiffEqBase.initialize!(opts.callback,integrator.u,prob.tspan[1],integrator)
     integrator
 end
 
-DiffEqBase.add_tstop!(integrator::SSAIntegrator,tstop) = integrator.tstop = tstop
+function DiffEqBase.add_tstop!(integrator::SSAIntegrator{<:Any,<:Any,<:Any,<:Any,<:Any,<:Any,<:Any,<:Any,<:Vector},tstop)
```
ChrisRackauckas, line 150:
```suggestion
function DiffEqBase.add_tstop!(integrator::SSAIntegrator,tstop)
```

should be fine.

ChrisRackauckas:
it looks good to me, but my look was really quick so @isaacsas should take a look later and then it should be good to go.

src/SSA_stepper.jl
manuel-rhdt: Well, by default SSAIntegrator gets created with tstops=(). Thus in the default case, dynamically inserting tstops is not possible since tuples are immutable. So by default, users would get a cryptic error message like „no method insert! for Tuple{}“.

Maybe that is fine. Or it might be better to change the default tstops to be an empty array (not sure about performance consequences).

ChrisRackauckas: Change it to an empty array. The tuple is likely an over-optimization.

src/SSA_stepper.jl
```diff
@@ -141,10 +143,25 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
...
+    insert_index = searchsortedfirst(integrator.tstops, tstop)
```
isaacsas, line 151: Looks good to me overall. Probably it won't matter for most applications, but couldn't you just search from `integrator.tstops_idx` on?

ChrisRackauckas (approved):

src/SSA_stepper.jl
manuel-rhdt: Good idea!

Issue #177, opened 2021-04-19
Solution becomes flat while using a large timespan


AliCanter:
When solving a `JumpProblem` during large time spans, at one point the solution becomes just flat.
I tried with `SSAStepper()` and `FunctionMap()` and both solvers give similar results.
There are no warnings and the retval is `:Default` (`:Success` when using `FunctionMap()`).

A minimal example of this behavior is the following (adapted from [Catalyst.jl examples](https://catalyst.sciml.ai/dev/tutorials/basic_examples/)):
```julia
using Catalyst
using DiffEqJump
using Plots

rs = @reaction_network begin
  c1, X --> 2X
  c2, X --> 0
  c3, 0 --> X
end c1 c2 c3

p = (1., 2., 50.)
dt = 1.
tspan = (0., 2e6)
u0 = [5]

dprob = DiscreteProblem(u0, tspan, p)
jprob = JumpProblem(dprob, Direct(), rs, save_positions=(false, false))
sol   = solve(jprob, SSAStepper(), saveat=tspan[1]:dt:tspan[2])
plot(sol, lw=2, title="Birth-Death Process")
```

This happens for almost every system I tried, and happens earlier if there are a lot of jumps, so it may be a limit in the number of jumps.

isaacsas:
Can reproduce with multiple SSA methods and when saving all time points.

isaacsas:
Have you ever had this occur when not using `save_positions` or `saveat`? I can't check them for this example as there are too many jumps, but it would be helpful in debugging.

AliCanter:
I don't think so, from what I tried the start of the plateau seems correlated with the amount of jumps that I am able save in RAM.

isaacsas:
I think the issue is floating point error. It appears that right after this line:
https://github.com/SciML/DiffEqJump.jl/blob/b14226d7729fcf06cb1d5ecc4a05a755edfee5ea/src/aggregators/direct.jl#L57

We can have that 
```julia
p.next_jump_time == t
```
is true.

isaacsas:
What seems to be happening is sometimes the time till the next jump is very small, on the order of 1e-13. It is just given by
```julia
ttnj = randexp() / sum_rate
```
In these cases the total propensity for the next reaction, `sum_rate`, is a couple hundred, so perfectly reasonably sized. So this is just coming from a very small value from `randexp` (relative to the current `t` value). Then 
```julia
p.next_jump_time = t + ttnj 
```
[...]

PR #178, opened 2021-04-20
update SSAStepper to allow multiple events per step

isaacsas:
https://github.com/SciML/DiffEqJump.jl/issues/177

@ChrisRackauckas this is what I was thinking of doing. Do you see a problem with this? (One issue is whether we want to allow saving after each jump, what this PR does, or only save after all the jumps.)

If this sounds reasonable let's see if tests pass.

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -44,8 +44,7 @@ function DiffEqBase.solve!(integrator)
     end_time = integrator.sol.prob.tspan[2]
     while should_continue_solve(integrator) # It stops before adding a tstop over
         step!(integrator)
-    end
-
+    end    
     integrator.t = end_time
 
     if integrator.saveat !== nothing && !isempty(integrator.saveat)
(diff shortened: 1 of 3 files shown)


isaacsas:
If this seems reasonable I'll add a test too.

ChrisRackauckas:
This is reasonable for SSAIntegrator. We should also have a good strategy that combines well with the ODE/SDE solvers though, but that could come at a later time. In reality, with hybrid, you'd want to never run into this issue if you're doing it right.

ChrisRackauckas:
>with hybrid, you'd want to never run into this issue if you're doing it right.

That comment is going to come back 2 years from now, I just know it.

isaacsas:
This is a fundamental issue is all the SSAs I think, so it will definitely occur in hybrid too for long enough `tspan`s.

isaacsas:
Shoot, I think we've got a catch-22 with this solution. If we don't include the callbacks in the loop then one potentially misses a callback. If you do include them you potentially execute a callback too many times.

isaacsas:
OK, so this just executes all the jumps, which means that callbacks are no longer exact. It is a hack, but gives a fix for now. I'll keep the issue open and next month can work on a more robust solution.

PR #183, opened 2021-06-25
Next Subvolume Method on Graphs

Vilin97:
An implementation of the Next Subvolume Method (NSM) -- the first spatial method in DiffEqJump. Also added two tests for spatial SSAs and a `utils.jl` file with some utility structs and functions for spatial SSAs.


src/aggregators/ssajump.jl
```diff
@@ -222,6 +224,33 @@ Notes:
...
+Perform linear search for `r` over iterator. Output index j s.t. sum(array[1:j-1])
+< r <= sum(array[1:j]), where array=collect(iterator)
+
+Notes:
+- The iterator must have all positive numbers
+- Returns index zero if the search is unsuccessful. Assumes this corresponds to
+  the case of an infinite next reaction time and so the jump index does not
+  matter.
+"""
+function linear_search(iterator, r)
```
Vilin97, line 239: linear search over an iterator instead of an array.

test/spatial/ABC.jl
```diff
@@ -0,0 +1,177 @@
...
+K = rates[2]/(rates[1]*mesh_size)
+function analyticmean(u, K)
+    α = u[1]; β = u[2]; γ = u[3]
+    @assert β ≥ α "A(0) must not exceed B(0)"
+    K * (α+γ)/(β-α+1) * pFq([-α-γ+1], [β-α+2], -K) / pFq([-α-γ], [β-α+1], -K)
+end
+
+analytic_A = analytic_B = analyticmean(u0, K)
+analytic_C = ((u0[1]+u0[2]+2*u0[3]) - (analytic_A + analytic_B))/2
+equilibrium_state = reshape(vcat([[analytic_A/num_nodes, analytic_B/num_nodes, analytic_C/num_nodes] for node in 1:num_nodes]...),length(u0), num_nodes)
```
Vilin97, line 68: this does not work correctly. Gives negative values

isaacsas (changes requested):
Still some more I need to look at in NSM, so will add on later. (Having PR comment issues in this review...)

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
+"""
+A file with types, structs and functions for spatial simulations
+"""
+
+"""
+stores info for a spatial jump
+"""
+struct SpatialJump{J}
```
isaacsas, line 8: Maybe

```julia
struct SpatialJump{J}
    src::J    # source location
    jidx::Int # index of jump as a diffusion hop or reaction
    dst::J    # destination location
end
```

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+############ abstract spatial system struct ##################
+"""
+Contains all info about the topology of the system
+"""
+abstract type AbstractSpatialSystem end
+
+"""
+returns neighbors of site
+"""
+function neighbors end
```
isaacsas, line 23: What advantage is there to actually declaring these (empty) functions vs. just leaving a comment that this is the API? I think it would perhaps be better to just comment them out here but make clear they are the API / public interface in documentation. (Happy to hear arguments against that though.)

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+return the number of neighbors of a site
+"""
+function num_neighbors end
+
+# TODO make a graph struct with connectivity list, compare with CartesianGrid
+
+################### CartesianGrid <: AbstractSpatialSystem ########################
+#TODO store the number of neighbors for each site or store all neighbors for each site.
+struct CartesianGrid <: AbstractSpatialSystem
+    dimension::Int # dimension of the grid, e.g. 1, 2 or 3
```
isaacsas, line 40: Why a field for the dimension instead of a type parameter? (As in Array{T,N} having N give the dimension?)

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+    linear_size::Int #side length of the grid
+end
```
isaacsas, line 42: Before we go too much further I'd suggest we add support for rectangular grids and make this a vector. Otherwise call this SquareCartesianGrid.

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+
+dimension(grid::CartesianGrid) = grid.dimension
+
+number_of_sites(grid) = grid.linear_size^dimension(grid)
```
isaacsas, line 46: If this is called in hot loops, maybe it makes more sense to store the total number of sites in CartesianGrid and calculate it in a constructor?

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+
+"""
+check if the given site_id is a valid site in the grid
+"""
+is_site(grid,site_id) = site_id >= 1 && site_id <= number_of_sites(grid)
```
isaacsas, line 51: This won't always work right. Think about a 2D, 10x10 grid. You can get location 11 if you calculate the right neighbor of site 10, which is not a valid location. (Or maybe you don't use this in places where that can happen?)

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+
+"""
+return a generator that iterates over the neighbors of the given site in grid
+"""
+function neighbors(grid, site_id)
```
isaacsas, line 56: Drop the generators here and in num_neighbors as we discussed.

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+    for _ in neighbors(grid,site_id)
+        counter += 1
+    end
+    counter
+end
+
+"""
+return the nth neighbor of the site
+"""
+function nth_neighbor(grid,site,n)
```
isaacsas, line 74: This will just switch to using an array?

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+# make function that returns the ith neighbor for a site -- this might be the sink state in case of absorbing
+# or just don't update the target site if it's not a valid site
+
+################### abstract spatial rates struct ###############
+
+abstract type AbstractSpatialRates end
+
+#TODO refactor names of functions?
+
+#return rate of site
```
isaacsas, line 100: #return total rate of site

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+function get_site_rate(spatial_rates_struct, site_id)
+    get_site_reactions_rate(spatial_rates_struct,site_id)+get_site_diffusions_rate(spatial_rates_struct,site_id)
+end
+
+#return rate of reactions at site
+function get_site_reactions_rate end
```
isaacsas, line 106: `# return total reaction rate of the site`

similar for diffusion below

again, do these need to be forward declared or can they just be commented out and made the API for new versions?

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+
+#sets the rate of reaction at site
+function set_site_reaction_rate! end
+
+#sets the rate of diffusion at site
+function set_site_diffusion_rate! end
+
+#################### SpatialRates <: AbstractSpatialRates ######################
+
+# NOTE: for now assume diffusion rate only depends on the site and species (and not the neighbor of the site)
```
isaacsas, line 125: Make clear first index is site and second is species in comments.

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+initializes SpatialRates with zero rates
+"""
+function SpatialRates(ma_jumps, num_species, spatial_system::AbstractSpatialSystem)
+    num_sites = number_of_sites(spatial_system)
+    num_jumps = get_num_majumps(ma_jumps)
+    SpatialRates(num_jumps,num_species,num_sites)
+end
+
+"""
+return rate of reactions at site
```
isaacsas, line 162: return total rate for any reaction at the site?

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+"""
+function get_site_reactions_rate(spatial_rates, site_id)
```
isaacsas, line 164: maybe call `total_site_rx_rate` or `total_site_rxrate`? Similar for the diffusion rate.

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+    spatial_rates.reaction_rates_sum[site_id]
+end
+
+"""
+return rate of diffusions at site
```
isaacsas, line 169: total rate for any diffusive jump from the site

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+    spatial_rates.reaction_rates_sum[site_id] += rate - old_rate
+end
+
+"""
+sets the rate of diffusion at site
+"""
+function set_site_diffusion_rate!(spatial_rates, site_id, species_id, rate)
+    old_rate = spatial_rates.diffusion_rates[site_id][species_id]
+    spatial_rates.diffusion_rates[site_id][species_id] = rate
+    spatial_rates.diffusion_rates_sum[site_id] += rate - old_rate
```
isaacsas, line 204: return nothing or old rate?

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+end
+
+
+# Tests for SpatialRates
```
isaacsas, line 208: Move these comments to a test (even if not included right now), or delete.

src/aggregators/ssajump.jl
isaacsas: Also, why drop the inline?

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+
+    for (i, val) in enumerate(xs)
+        i >= n && return val
+    end
+
+    # catch iterators with no length but actual finite size less then n
+    throw(BoundsError(xs, n))
+end
+
+#TODO use the Sampler + rand interface to draw a random neighbor as described here https://docs.julialang.org/en/v1/stdlib/Random/#Random.Sampler.
```
isaacsas, line 88: Remove these TODOs before we merge. Probably better to just have TODOs in the first PR post as a checklist.

isaacsas:
Last comment for now; maybe instead of diffusion_rate and such say hopping_rate or transport_rate? Just to make clear it isn't assuming diffusion.

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+"""
+reevaluate all rates, recalculate tentative site firing times, and reinit the priority queue
+"""
+function fill_rates_and_get_times!(aggregation::NSMJumpAggregation, u, t)
+    @unpack ma_jumps, diffusion_constants, spatial_system = aggregation
+
+    num_majumps = get_num_majumps(ma_jumps)
+    num_species = length(@view u[:,1]) #NOTE assumes u is a matrix with ith column being the ith site
+    num_sites = number_of_sites(spatial_system)
+    cur_rates = SpatialRates(num_majumps,num_species,num_sites)
```
isaacsas, line 128: Haven’t you already allocated this in the aggregation? Why not fill in the already allocated fields?

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+    @assert cur_rates.diffusion_rates_sum == zeros(typeof(cur_rates.diffusion_rates_sum[1]),num_sites)
+
+    pqdata = Vector{typeof(t)}(undef, num_sites)
+    for site in 1:num_sites
+        update_reaction_rates!(cur_rates, 1:num_majumps, u, ma_jumps, site)
+        update_diffusion_rates!(cur_rates, 1:num_species, diffusion_constants, u, site, spatial_system)
+        pqdata[site] = t + randexp(aggregation.rng) / get_site_rate(cur_rates, site)
+    end
+
+    aggregation.cur_rates = cur_rates
```
isaacsas, line 140: Again, don’t allocate just use the existing allocate field.

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+"""
+function update_state!(p, integrator)
+    jump = p.next_jump
+    if is_diffusion(p, jump)
+        execute_diffusion!(integrator, jump.site, jump.target_site, jump.index)
+    else
+        # u_site = integrator.u[:,jump.site]
+        rx_index = reaction_id_from_jump(p,jump)
+        executerx!((@view integrator.u[:,jump.site]), rx_index, p.ma_jumps)
+        # executerx!(u_site, rx_index, p.ma_jumps)
```
isaacsas, line 219: Cleanup comments.

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+
+recalculate jump rates for jumps that depend on the just executed jump (p.prev_jump)
+"""
+function update_dependent_rates_and_firing_times!(p, u, t)
+    jump = p.prev_jump
+    if is_diffusion(p, jump)
+        source_site = jump.site
+        target_site = jump.target_site
+        update_rates_after_diffusion!(p, u, t, source_site, target_site, jump.index)
+        for site in [source_site, target_site]
```
isaacsas, line 156: Just call the function twice; don’t use a loop…

isaacsas: This allocates a new vector…

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+    integrator.u[species,target_site] += 1
+end
+
+"""
+    reaction_id_from_jump(p,jump)
+
+return reaction id by subtracting the number of diffusive hops
+"""
+function reaction_id_from_jump(p,jump)
+    jump.index - length(@view p.diffusion_constants[:,jump.site])
```
isaacsas, line 254: Use `size`

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+end
+
+"""
+    is_diffusion(p, jump)
+
+true if jump is a diffusion
+"""
+function is_diffusion(p, jump)
+    # size(p.diffusion_constants,1)
+    jump.index <= length(@view p.diffusion_constants[:,jump.site])
```
isaacsas, line 235: Use `size`

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+######################## helper routines for all spatial SSAs ########################
+function update_rates_after_reaction!(p, u, t, site, reaction_id)
+    update_reaction_rates!(p.cur_rates, p.dep_gr[reaction_id], u, p.ma_jumps, site)
+    update_diffusion_rates!(p.cur_rates, p.jumptovars_map[reaction_id], p.diffusion_constants, u, site, p.spatial_system)
+end
+
+function update_rates_after_diffusion!(p, u, t, source_site, target_site, species)
+    for site in [source_site, target_site]
+        update_reaction_rates!(p.cur_rates, p.vartojumps_map[species], u, p.ma_jumps, site)
+        update_diffusion_rates!(p.cur_rates, [species], p.diffusion_constants, u, site, p.spatial_system)
```
isaacsas, line 184: Why not make a version that takes a scalar for `species` and avoid creating a vector? Or else, can it work with a tuple instead?

isaacsas:
@Vilin97 ok, I think that’s it for the main new code. Let me know when you’ve made updates.

src/aggregators/ssajump.jl
Vilin97: Looks like there is no difference noticeable. I tested Direct on gene expression and found no difference in allocations or performance

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+    spatial_rates.diffusion_rates[site_id]
+end
+
+"""
+set the rate of reaction at site
+"""
+function set_site_reaction_rate!(spatial_rates, site_id, reaction_id, rate)
+    old_rate = spatial_rates.reaction_rates[site_id][reaction_id]
+    spatial_rates.reaction_rates[site_id][reaction_id] = rate
+    spatial_rates.reaction_rates_sum[site_id] += rate - old_rate
```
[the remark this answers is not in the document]
Vilin97: Returning old_rate

src/spatial/utils.jl
```diff
@@ -0,0 +1,219 @@
...
+standard constructor, assumes numeric values for rates
+"""
+function SpatialRates(reaction_rates::Vector{Vector{R}}, diffusion_rates::Vector{Vector{R}}) where {R}
+    SpatialRates{R}(reaction_rates, diffusion_rates, [sum(site) for site in reaction_rates], [sum(site) for site in diffusion_rates])
+end
+#QUESTION are the type annotations here correct? Do I need them at all?
+"""
+initializes SpatialRates with zero rates
+"""
+function SpatialRates(num_jumps::Integer,num_species::Integer,num_sites::Integer)
```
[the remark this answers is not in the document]
Vilin97: Changed

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+
+    next_jump = SpatialJump{Int}(typemax(Int),typemax(Int),typemax(Int)) #a placeholder
+    next_jump_time = typemax(typeof(end_time))
+    current_rates = SpatialRates(get_num_majumps(majumps), num_species, number_of_sites(spatial_system))
+
+    NSMJumpAggregation(next_jump, next_jump_time, end_time, current_rates, diffusion_constants, majumps, save_positions, rng, spatial_system; num_specs = num_species, kwargs...)
+end
+
+function aggregate(aggregator::NSM, starting_state, p, t, end_time, constant_jumps, ma_jumps, save_positions, rng; diffusion_constants, spatial_system, kwargs...)
+    num_species = length(@view starting_state[:,1])
```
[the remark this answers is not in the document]
Vilin97: changed

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+    end
+
+    pq = MutableBinaryMinHeap{T}()
+
+    NSMJumpAggregation{J,T,R,C,S,RNG,typeof(dg),typeof(vtoj_map),typeof(jtov_map),typeof(pq),SS}(nj, nj, njt, et, crs, diffusion_constants, maj, sps, rng, dg, vtoj_map, jtov_map, pq, spatial_system)
+end
+
+############################# Required Functions ##############################
+# creating the JumpAggregation structure (function wrapper-based constant jumps)
+function aggregate(aggregator::NSM, num_species, end_time, diffusion_constants, ma_jumps, save_positions, rng, spatial_system; kwargs...)
```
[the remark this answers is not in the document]
Vilin97: The first one takes in only the necessary information. The second one takes in extra inputs such as `p` and `t` in order to comply with the existing interface in `ssajump.jl`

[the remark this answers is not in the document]
isaacsas: I thinks it makes sense to just merge these together into one with the existing interface.

src/spatial/utils.jl
Vilin97: changed

src/spatial/utils.jl
Vilin97: One advantage is that (as far as I understand) the docstring will apply to all methods of this abstract function. Another is that perhaps some of them can be implemented from each other, such that `get_site_rate` for `AbstractSpatialRates`. In OOP languages this is standard practice. I am happy to change it if you say so.

src/spatial/nsm.jl
Vilin97: good idea. Making a version that takes a scalar

src/spatial/nsm.jl
Vilin97: okay

src/spatial/nsm.jl
Vilin97: okay

src/spatial/nsm.jl
Vilin97: got it, thanks

src/spatial/nsm.jl
Vilin97: got it

src/spatial/nsm.jl
Vilin97: okay

src/spatial/nsm.jl
Vilin97: yes, this was because it needed to reset after a simulation. I added a `reset!` function to take care of that without allocating

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+"""
+function fill_rates_and_get_times!(aggregation::NSMJumpAggregation, u, t)
+    @unpack ma_jumps, diffusion_constants, spatial_system = aggregation
+
+    num_majumps = get_num_majumps(ma_jumps)
+    num_species = length(@view u[:,1]) #NOTE assumes u is a matrix with ith column being the ith site
+    num_sites = number_of_sites(spatial_system)
+    cur_rates = SpatialRates(num_majumps,num_species,num_sites)
+
+    @assert cur_rates.reaction_rates_sum == zeros(typeof(cur_rates.reaction_rates_sum[1]),num_sites)
```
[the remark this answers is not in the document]
Vilin97: okay

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+
+######################## SSA specific helper routines ########################
+"""
+reevaluate all rates, recalculate tentative site firing times, and reinit the priority queue
+"""
+function fill_rates_and_get_times!(aggregation::NSMJumpAggregation, u, t)
+    @unpack ma_jumps, diffusion_constants, spatial_system = aggregation
+
+    num_majumps = get_num_majumps(ma_jumps)
+    num_species = length(@view u[:,1]) #NOTE assumes u is a matrix with ith column being the ith site
```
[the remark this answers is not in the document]
Vilin97: okay

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+
+#NOTE integrator and params are not used. They remain to adhere to the interface of `AbstractSSAJumpAggregator` defined in ssajump.jl
+# calculate the next jump / jump time
+function generate_jumps!(p::NSMJumpAggregation, integrator, params, u, t)
+    @unpack cur_rates, rng = p
+
+    p.next_jump_time, site = top_with_handle(p.pq)
+    if rand(rng)*get_site_rate(cur_rates, site) < get_site_reactions_rate(cur_rates, site)
+        rx = linear_search(get_site_reactions_iterator(cur_rates, site), rand(rng) * get_site_reactions_rate(cur_rates, site))
+        p.next_jump = SpatialJump(site, rx+length(@view p.diffusion_constants[:,site]), site)
```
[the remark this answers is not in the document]
Vilin97: okay

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+
+#NOTE integrator and params are not used. They remain to adhere to the interface of `AbstractSSAJumpAggregator` defined in ssajump.jl
+# set up a new simulation and calculate the first jump / jump time
+function initialize!(p::NSMJumpAggregation, integrator, u, params, t)
+    fill_rates_and_get_times!(p, u, t)
+    generate_jumps!(p, integrator, params, u, t)
+    nothing
+end
+
+#NOTE integrator and params are not used. They remain to adhere to the interface of `AbstractSSAJumpAggregator` defined in ssajump.jl
```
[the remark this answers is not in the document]
Vilin97: okay

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+
+    NSMJumpAggregation(next_jump, next_jump_time, end_time, current_rates, diffusion_constants, majumps, save_positions, rng, spatial_system; num_specs = num_species, kwargs...)
+end
+
+function aggregate(aggregator::NSM, starting_state, p, t, end_time, constant_jumps, ma_jumps, save_positions, rng; diffusion_constants, spatial_system, kwargs...)
+    num_species = length(@view starting_state[:,1])
+    aggregate(aggregator, num_species, end_time, diffusion_constants, ma_jumps, save_positions, rng, spatial_system; kwargs...)
+end
+
+#NOTE integrator and params are not used. They remain to adhere to the interface of `AbstractSSAJumpAggregator` defined in ssajump.jl
```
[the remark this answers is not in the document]
Vilin97: okay

src/spatial/nsm.jl
```diff
@@ -0,0 +1,269 @@
...
+    @unpack cur_rates, rng = p
+
+    p.next_jump_time, site = top_with_handle(p.pq)
+    if rand(rng)*get_site_rate(cur_rates, site) < get_site_reactions_rate(cur_rates, site)
+        rx = linear_search(get_site_reactions_iterator(cur_rates, site), rand(rng) * get_site_reactions_rate(cur_rates, site))
+        p.next_jump = SpatialJump(site, rx+length(@view p.diffusion_constants[:,site]), site)
+    else
+        species_to_diffuse = linear_search(get_site_diffusions_iterator(cur_rates, site), rand(rng) * get_site_diffusions_rate(cur_rates, site))
+        #TODO this is not efficient. We iterate over neighbors twice.
+        n = rand(rng,1:num_neighbors(p.spatial_system, site))
```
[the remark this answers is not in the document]
Vilin97: I would imagine that's how it's internally implemented

src/spatial/utils.jl
Vilin97: Yes, you are right. I fixed it

src/spatial/utils.jl
Vilin97: Where should I put them for now? Or can they stay until the merge?

src/spatial/utils.jl
```diff
@@ -0,0 +1,209 @@
...
+function to_coordinates(grid::CartesianGrid{3}, site)
+    temp = mod1(site,grid.linear_sizes[1]*grid.linear_sizes[2])
+    (mod1(temp, grid.linear_sizes[1]),fld1(temp, grid.linear_sizes[1]), fld1(site, grid.linear_sizes[1]*grid.linear_sizes[2]))
+end
+
+from_coordinates(grid::CartesianGrid{1}, x) = x
+from_coordinates(grid::CartesianGrid{2}, (x,y)) = (y-1) * grid.linear_sizes[1] + x
+from_coordinates(grid::CartesianGrid{3}, (x,y,z)) = (y-1) * grid.linear_sizes[1] + x + (z-1)*grid.linear_sizes[1]*grid.linear_sizes[2]
+
+is_site(grid,site_id::Int) = site_id >= 1 && site_id <= num_sites(grid)
```
isaacsas, line 63: `issite`

src/spatial/utils.jl
isaacsas: They can stay till we merge, but we should drop at that point. Better to track them long-term in a spatial TODO issue.

src/spatial/utils.jl
```diff
@@ -0,0 +1,209 @@
...
+function is_site(grid,site_coordinates::Tuple)
```
isaacsas, line 64: Why a Tuple instead of a CartesianIndex?

src/spatial/utils.jl
```diff
@@ -0,0 +1,209 @@
...
+    length(site_coordinates) == dimension(grid) || return false
+    for (i,c) in enumerate(site_coordinates)
+        1 <= c && c <= grid.linear_sizes[i] || return false
+    end
+    return true
+end
+
+# TODO make these iterators so they don't allocate memory
```
isaacsas, line 72: Iterators would be better, but even a Tuple of CartesianIndices would be an improvement.

src/spatial/nsm.jl
isaacsas: Alternatively here you could use a Tuple, but probably better to unroll the loop with only two sites as you did.

src/spatial/nsm.jl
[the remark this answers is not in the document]
isaacsas: No, if you did down it is specialized for linear ranges (so is perhaps better/faster). 

We can look at this later if it is rate limiting in benchmarking. I'd leave as is for now.

src/spatial/utils.jl
Vilin97: I can change it but I don't see why `issite` is better. It's less readable and seems to not follow the current style in the package, e.g. `update_state` instead of `updatestate`

src/spatial/utils.jl
Vilin97: Because I have not heard of `CartesianIndex` :). Why is it better than `Tuple`?

isaacsas: It is the native way to index matrices and such with 2 or more dimensions. Like a tuple but designed for indexing. Maybe checkout the docs on it.

src/spatial/utils.jl
isaacsas: https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/

We aren’t great about this, and some names in DiffEqJump predate it.

src/spatial/utils.jl
isaacsas: There are also probably already routines for converting to linear indices and vice versa in Base.

src/spatial/utils.jl
Vilin97: okay, changed

src/spatial/utils.jl
Vilin97: Will look into this after this PR is merged as agreed today

src/spatial/utils.jl
Vilin97: Will do after the PR is merged as agreed today

Vilin97:
@isaacsas, are we ready to merge?

src/spatial/nsm.jl
isaacsas: @Vilin97 did you want to update this one?

src/spatial/nsm.jl
```diff
@@ -0,0 +1,257 @@
...
+    nothing
+end
+
+"""
+    is_diffusion(p, jump)
+
+true if jump is a diffusion
+"""
+function is_diffusion(p, jump)
+    jump.jidx <= size(p.diffusion_constants,1)
```
isaacsas, line 223: I wonder if we should just store a `numspecies` field in `p` and calculate this in the constructor? This might make a bunch of the code cleaner.

src/spatial/utils.jl
```diff
@@ -0,0 +1,217 @@
+using LightGraphs
```
isaacsas, line 1: Put this in DiffEqJump.jl

src/spatial/utils.jl
```diff
@@ -0,0 +1,217 @@
...
+"""
+abstract type AbstractSpatialSystem end
+
+# Interface:
+# num_sites(spatial_system) = total number of sites
+# neighbors(spatial_system, site) = an iterator over the neighbors of site
+# num_neighbors(spatial_system, site) = number of neighbors of site
+
+################### LightGraph ########################
+num_sites(graph::AbstractGraph) = LightGraphs.nv(graph)
```
isaacsas, line 27: If you are getting name clashes so you need to qualify with `LightGraphs`, you might as well just `import LightGraphs` instead of `using`.

src/spatial/utils.jl
```diff
@@ -0,0 +1,217 @@
...
+neighbors(graph::AbstractGraph, site) = LightGraphs.neighbors(graph, site)
```
isaacsas, line 28: If you are `using LightGraphs` why do you need to define this? Calling `neighbors(graph,site)` should call `LightGraphs.neighbors(graph,site)` already.

test/spatial/ABC.jl
```diff
@@ -0,0 +1,82 @@
...
+end_time = 10.0
+diffusivity = 1.0
+
+# testing on CartesianGrid
+grid = CartesianGrid(dim, linear_size)
+spatial_jump_prob = ABC_setup(grid, starting_site, u0, diffusivity, end_time)
+solution = solve(spatial_jump_prob, SSAStepper())
+mean_end_state = get_mean_end_state(spatial_jump_prob, Nsims)
+diff =  sum(mean_end_state, dims = 2) - non_spatial_mean
+println("max relative error: $(maximum(abs.(diff./non_spatial_mean)))")
```
isaacsas, line 55: comment out

test/spatial/ABC.jl
```diff
@@ -0,0 +1,82 @@
...
+
+# testing on LightGraphs
+lgrid = LightGraphs.grid([linear_size])
+lspatial_jump_prob = ABC_setup(lgrid, starting_site, u0, diffusivity, end_time)
+lsolution = solve(lspatial_jump_prob, SSAStepper())
+
+lmean_end_state = get_mean_end_state(lspatial_jump_prob, Nsims)
+
+diff =  sum(lmean_end_state, dims = 2) - non_spatial_mean
+println("max relative error: $(maximum(abs.(diff./non_spatial_mean)))")
```
isaacsas, line 68: comment out

src/spatial/nsm.jl
isaacsas: We can leave this one for future cleanup.

isaacsas:
@ChrisRackauckas this has a first interface for spatial SSAs if you want to take a look before we merge. Otherwise, once my last few (minor) comments are addressed I'll merge this. This should handle RDMEs with MassActionJumps on a graph and CartesianGrid, with the constraint that the rates to hop from a given location to each neighbor must be the same (i.e. the total out rate can vary by species and site, but the rate to each neighbor for a given species and site is assumed constant). We'll generalize to varying out rates to each neighbor in the future (along with allowing varying reaction rates across sites).

src/aggregators/aggregators.jl
```diff
@@ -79,6 +79,10 @@ systems with many species and many channels, Journal of Physical Chemistry A,
 """
 struct NRM <: AbstractAggregatorAlgorithm end
 
+"""
+citation
```
ChrisRackauckas, line 83: missing

src/spatial/utils.jl
ChrisRackauckas: Why is the location parameterized?

src/aggregators/aggregators.jl
isaacsas: @Vilin97 The original paper is: Elf, Johan and Ehrenberg, M, Spontaneous separation of bi-stable biochemical systems into spatial domains of opposite phases,Systems Biology, 2004 vol. 1(2) pp. 230-236

ChrisRackauckas:
Do the fallbacks work out so that if you apply direct to a spatial problem it works (but is slow)? That would be nice to make sure works for benchmarking.

isaacsas:
Not currently. The flattening PR from last summer would need to be updated. I thought it would actually be easier to get a basic interface for the problem specification done for a spatial algorithm, and now with that interface we can redo the flattening algorithm. 

Right now this reuses `JumpProblem` since there was really nothing new needed at that level (just some kwargs to communicate spatial info). Do you think it would be better to make a new `SpatialJumpProblem`? (That is actually what @Vilin97 did initially but since it was identical I said to stick with the old one for compatibility reasons -- this works with the existing SSAStepper just fine for example, and I think could be coupled to ODEs.) We could check if the problem is spatial via a proxy like one of the kwargs being defined, and flatten within the current `JumpProblem` in that case.

src/spatial/utils.jl
Vilin97: For some reason I get an error if I don't do it. Not sure why.

src/aggregators/aggregators.jl
Vilin97: Thank you. Added

src/spatial/utils.jl
isaacsas: Are you trying to set
```julia
 neighbors(graph::AbstractGraph, site) = neighbors(graph, site) 
```
You don't need to do that...

If not, where / what error are you getting if you just remove this definition?

isaacsas (approved):

ChrisRackauckas:
>Right now this reuses JumpProblem since there was really nothing new needed at that level (just some kwargs to communicate spatial info). Do you think it would be better to make a new SpatialJumpProblem? (That is actually what @Vilin97 did initially but since it was identical I said to stick with the old one for compatibility reasons -- this works with the existing SSAStepper just fine for example, and I think could be coupled to ODEs.) We could check if the problem is spatial via a proxy like one of the kwargs being defined, and flatten within the current JumpProblem in that case.

I think you have the right approach here. Though if a method that is for space is used and a grid isn't given, does it error? I think that would be a good check. 

>Not currently. The flattening PR from last summer would need to be updated. I thought it would actually be easier to get a basic interface for the problem specification done for a spatial algorithm, and now with that interface we can redo the flattening algorithm.

Got it.

src/spatial/utils.jl
Vilin97: In case we ever want to use something like `CartesianIndex` for site locations instead of integers

Vilin97:
If `NSM()` is used and no grid or graph is given it will error in `aggregate`.

src/spatial/nsm.jl
Vilin97: Oops, did not see this before merging. Will update in the next PR

PR #190, opened 2021-07-08
Allow MassActionJumps to use standard parameter vectors

isaacsas:
This should allow a new `MassActionJump` use case
```
p = (1.0, 2.0)

# use 2.0 for the first reaction rate and 1.0 for the second
MassActionJump(reaction_stoich, net_stoich; param_idxs=[2,1], params=p)
```
while preserving compatibility with the old style. Rates are still fixed at construction time, but in this mode they are read from the parameter vector `p` according to the indices in `param_idxs`. This ordering is then saved so that later one can use an updated parameter vector/tuple to update a mass action jump (if it was setup in this way, otherwise the old style will still work just fine).

@ChrisRackauckas if you think this approach makes sense I'll add in a function for remaking in callbacks, which we can subsequently point users too.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DiffEqJump"
 uuid = "c894b116-72e5-5b58-be3c-e6d8d4ac2b12"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "6.15"
+version = "6.15.0"
 
 [deps]
 ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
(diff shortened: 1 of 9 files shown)


isaacsas:
Once this is done I’ll also update Catalyst and MT to use this as the default.

ChrisRackauckas (approved):
This is mostly good. I can still see a potential issue with callbacks, but this is infinitely better than before.

isaacsas:
@ChrisRackauckas I wanted to check the approach with you before handling callbacks... I've just put them in now. Basically `remake` and `reset_aggrevated_jumps!` should now both update `MassActionJump`s if they were initialized with parameter indices.

isaacsas:
The only consideration is whether we want to not always update the `MassActioJump` parameters in `reset_aggregated_jumps!` (i.e. maybe we should allow user control over this with a kwarg).

Issue #189, opened 2021-07-08
Spatial SSAs progress and TODOs


Vilin97:
This summer I am working on implementing spatial SSAs in `DiffEqJump`. There are currently two issues open about it: https://github.com/SciML/DiffEqJump.jl/issues/107 and https://github.com/SciML/DiffEqJump.jl/issues/130. The first PR https://github.com/SciML/DiffEqJump.jl/pull/183 implements the NSM, adds two tests and some utility functions, essentially setting up the interface for spatial SSAs. There are a few TODOs that should not be forgotten:
- [ ] The citation must be added to the just-implemented NSM.
- [ ] More comprehensive tests for spatial SSAs are needed. The two current tests only test spatial systems on 1D grids.
- [ ] The `CartesianGrid` struct needs to be optimized -- probably by storing the number of neighbors for each site so `num_neighbors` is $O(1)$ and making `neighbors` return an iterator instead of allocating an array. 
- [ ] The escaping boundary condition for `CartesianGrid` needs to be supported. This can be done with a special sink site: it neighbors all boundary sites and does not get updated if it is the target site of a hop. This change will require tweaking the code in `nsm.jl`.
- [ ] Various modes of hopping constants need to be supported. In particular, we should support hopping rates of forms $D_s * L_{i,j}$, $L_{s,i,j}$, $D_{s,i}*L_{i,j}$, where $D_s$ is the diffusivity constant of species $s$ and $L$ is the discrete Laplacian matrix. Only the first of these three is implemented in the PR. This change will require tweaking the code in `nsm.jl`.
- [ ] Constant-rate jumps should be supported. Currently, only mass-action jumps are allowed. This change will require tweaking the code in `nsm.jl`.
- [ ] An interface for sampling a neighbor to hop to can be a nice but not necessary perk. For example, using https://docs.julialang.org/en/v1/stdlib/Random/#Random.Sampler.

@isaacsas and @ChrisRackauckas might have more TODOs to add to this list.

isaacsas:
Additional features to plan for:
1. Spatially varying reaction rates (modify MassActionJumps?)
2. Scaling reaction rates by site-dependent factor (e.g. area/volume).

isaacsas:
Some more stuff:

- [ ] Spatially distributed reactions and product placement (nearest neighbors, but also neighbors within some stencil)
- [ ] Spatial SSA that follows individual particles or just non-empty voxels (often faster for systems with less particle than spatial sites).

[...]

PR #192, opened 2021-07-10
Spatial TODOs

Vilin97:
Completing TODOs from https://github.com/SciML/DiffEqJump.jl/issues/189. So far, wrote a pure diffusion test on a 2D grid and added `numspecies` to the NSM aggregator.
EDIT: wrote, optimized and benchmarked several versions of `CartesianGrid`, and implemented support of hopping rates that vary from neighbor to neighbor. Other forms of hopping rates, such as D_s * L_{i,j}, can now easily be added without modifying the code in `nsm.jl`. This completes all but three last TODOs.


Vilin97:
Still need to figure out how to write the `neighbors` function without allocations

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+    function to_spatial(stoch, num_nodes)
+        out = []
+        for site in 1:num_nodes
+            push!(out, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), stoch))
+        end
+        vcat(out...)
+    end
+    rx_netstoch =  to_spatial(netstoch, num_nodes)
+    rx_reactstoch =  to_spatial(reactstoch, num_nodes)
+    rx_rates = vcat([rates for i in 1:num_nodes]...)
```
isaacsas, line 43: I think splatting and `vcat` can potentially have performance issues on large systems (i.e. many spatial sites). Just preallocate the array and fill it in block by block, or else perhaps use `append!`?

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+
+    #hops
+    hop_netstoch = []
+    hop_reacstoch = []
+    hop_rates = []
+
+    # assuming hopping_constants isa Vector{Matrix} where hopping_constants[src][species, index] is the hop const of species from src to neighbor at index.
+    for src in vertices(graph)
```
isaacsas, line 51: What if the user didn't pass a graph (i.e. a Cartesian grid)?

isaacsas: So this doesn't work with other `hopping_constant` forms? (Like if the out rate is constant?)

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+            end
+        end
+    end
+
+    total_netstoch = convert(Vector{Vector{Pair{Int, Int}}}, vcat(hop_netstoch, rx_netstoch))
+    total_reacstoch = convert(Vector{Vector{Pair{Int, Int}}}, vcat(hop_reacstoch, rx_reactstoch))
+    total_rates = convert(Vector{Float64}, vcat(hop_rates, rx_rates))
+    ma_jump = MassActionJump(total_rates, total_reacstoch, total_netstoch)
+
+    flattened_u0 = reshape(u0, length(u0))
```
isaacsas, line 69: or just use `vec(u0)`.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+    prob = DiscreteProblem(flattened_u0, tspan, total_rates)
+
+    JumpProblem(prob, alg, ma_jump; kwargs...)
```
isaacsas, line 72: I think it would be better for this function to be an intermediary within `JumpProblem`, and return the data that the current `JumpProblem` constructor needs at https://github.com/SciML/DiffEqJump.jl/blob/512a711fb53c1793fc46a9c55b2eaa4030c14c29/src/problem.jl#L96

isaacsas (changes requested):
I still need to review NSM and utils, but here are my comments on everything else.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
+using DiffEqJump, DiffEqBase, LightGraphs
+
+rates = [0.1, 1.0]
```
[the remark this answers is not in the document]
Vilin97: Forgot to delete. Deleted now.

src/spatial/nsm.jl
```diff
@@ -239,16 +212,7 @@ end
 return reaction id by subtracting the number of hops
 """
 function reaction_id_from_jump(p,jump)
```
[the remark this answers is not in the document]
Vilin97: It is not. Moving it and other helper functions common to all spatial SSAs to `utils.jl`.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+u0 = ones(Int, num_species, num_nodes)
+alg = Direct()
+
+hopping_rate = 1.0
+hopping_constants = Vector{Matrix{Float64}}(undef, num_nodes) # TODO QUESTION is this how we want to accept hopping constants? 
+for site in 1:num_nodes
+    hopping_constants[site] = hopping_rate*ones(num_species, DiffEqJump.num_neighbors(graph, site))
+end
+
+#TODO QUESTION should accept variable rx rates?
```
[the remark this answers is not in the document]
Vilin97: Leave variable rx rates for a future PR? Okay.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+
+# reaction
+function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rates::AbstractArray, graph::AbstractGraph, u0::Matrix{Int}, tspan, alg, hopping_constants::Vector{Matrix{F}}; kwargs...) where F <: Number
```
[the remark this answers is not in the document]
Vilin97: The reason I want to take `rates` as input is that in `MassActionJump` only scaled rates are stored. I could write a function called `unscalerate` -- inverse to `scalerate`. Then I would first unscale rates from massaction jumps and then flatten. If you think it would be better, I'll go ahead and do that.

test/spatial/ABC.jl
```diff
@@ -2,75 +2,69 @@ using DiffEqJump, DiffEqBase
...
+# testing on CartesianGrid
+grids = [DiffEqJump.CartesianGrid1(dims), DiffEqJump.CartesianGrid2(dims), DiffEqJump.CartesianGrid3(dims), LightGraphs.grid(dims)]
+jump_problems = JumpProblem[JumpProblem(prob, alg, majumps, hopping_constants=hopping_constants, spatial_system = grid, save_positions=(false,false)) for grid in grids]
+# setup flattenned jump prob
+graph = LightGraphs.grid(dims)
+hopping_constants = Vector{Matrix{Float64}}(undef, num_nodes)
+for site in 1:num_nodes
+    hopping_constants[site] = hopping_rate*ones(num_species, DiffEqJump.num_neighbors(graph, site))
+end
+push!(jump_problems, DiffEqJump.flatten(netstoch, reactstoch, rates, graph, starting_state, tspan, NRM(), hopping_constants))
```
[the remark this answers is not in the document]
Vilin97: Got it. Done

test/spatial/benchmarks.jl
```diff
@@ -0,0 +1,85 @@
+# benchmarks for various grids
```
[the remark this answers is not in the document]
Vilin97: I'll remove it from the PR for now then

test/spatial/diffusion.jl
```diff
@@ -10,51 +11,74 @@ function get_mean_sol(jump_prob, Nsims, saveat)
...
 diffusivity = 0.1
-dim = 1
-grid = DiffEqJump.CartesianGrid(dim, linear_size)
-num_nodes = DiffEqJump.num_sites(grid)
+dim = 2
+dims = Tuple([linear_size for i in 1:dim])
+num_nodes = prod(dims)
 
 # Starting state setup
 starting_state = zeros(Int, length(u0), num_nodes)
```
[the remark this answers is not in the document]
Vilin97: At the moment this would break things. It would be easy to support receiving input in this form. In order to support output in this form, there can be a `cartesianize`, which takes a solution with "unrolled" shape and reshapes it into Cartesian shape.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+            push!(out, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), stoch))
+        end
+        vcat(out...)
+    end
+    rx_netstoch =  to_spatial(netstoch, num_nodes)
+    rx_reactstoch =  to_spatial(reactstoch, num_nodes)
+    rx_rates = vcat([rates for i in 1:num_nodes]...)
+
+    #hops
+    hop_netstoch = []
```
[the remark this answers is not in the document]
Vilin97: fixed

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+hopping_constants = Vector{Matrix{Float64}}(undef, num_nodes) # TODO QUESTION is this how we want to accept hopping constants? 
+for site in 1:num_nodes
+    hopping_constants[site] = hopping_rate*ones(num_species, DiffEqJump.num_neighbors(graph, site))
+end
+
+#TODO QUESTION should accept variable rx rates?
+
+# reaction
+function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rates::AbstractArray, graph::AbstractGraph, u0::Matrix{Int}, tspan, alg, hopping_constants::Vector{Matrix{F}}; kwargs...) where F <: Number
+    graph = SimpleDiGraph(graph)
```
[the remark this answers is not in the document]
Vilin97: fixed

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+    num_rxs = length(rates)
+    num_species = size(u0, 1)
+    num_nodes = nv(graph)
+    @assert size(u0, 2) == num_nodes
+    spec_CI = CartesianIndices((num_species, num_nodes))
+    spec_LI = LinearIndices((num_species, num_nodes))
+
+    #reactions
+    function to_spatial(stoch, num_nodes)
+        out = []
```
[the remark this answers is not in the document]
Vilin97: fixed

src/spatial/flatten.jl
```diff
@@ -0,0 +1,73 @@
...
+        for site in 1:num_nodes
+            push!(out, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), stoch))
+        end
+        vcat(out...)
```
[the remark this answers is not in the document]
Vilin97: good point. Fixed

src/spatial/flatten.jl
Vilin97: I eliminated all but three `vcat`s.

src/spatial/flatten.jl
Vilin97: I added support for other grids like `CartesianGrid`. The reason I did it this way at first is that the grid itself is not used in `solve`. It is only used in flattening. Therefore, I thought there would be no reason for the user to pass in something specialized like `CartesianGrid` if `LightGraphs.grid` is available.

If `hopping_constants` are of another form, when we flatten, each hop must still get its own rate. Thus, there is no performance to be gained from accepting other forms of `hopping_constants`. I can add it for user convenience though.

src/spatial/flatten.jl
Vilin97: Agree. Fixed it.

src/spatial/flatten.jl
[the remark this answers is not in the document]
Vilin97: Update: I did add support for variable reaction rates in flattening. It makes more sense, because each flattened reaction gets its own rates after flattening anyway. So may as well accept variable rates as input.

src/spatial/flatten.jl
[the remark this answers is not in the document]
isaacsas: Why would you need to unscale the rate? The scalings there should still be applied in the spatial case. They are scalings due to the reaction order, not volumes or such. It isn't generally true that `p` maps directly to the rates in a `MassActionJump`, so you need to work with `scaled_rates` as the rate constants.

isaacsas (changes requested):

src/spatial/nsm.jl
```diff
@@ -23,15 +20,15 @@ mutable struct NSMJumpAggregation{J,T,R<:AbstractSpatialRates,C,S,RNG,DEPGR,VJMA
...
-function NSMJumpAggregation(nj::SpatialJump{J}, njt::T, et::T, crs::R, hopping_constants::C,
-                                      maj::S, sps::Tuple{Bool,Bool},
-                                      rng::RNG, spatial_system::SS; num_specs, vartojumps_map=nothing, jumptovars_map=nothing, dep_graph=nothing, kwargs...) where {J,T,S,R,C,RNG,SS}
+function NSMJumpAggregation(nj::SpatialJump{J}, njt::T, et::T, rx_rates::RX, hop_rates::HOP, sps::Tuple{Bool,Bool},
+                                      rng::RNG, spatial_system::SS; num_specs, vartojumps_map=nothing, jumptovars_map=nothing, dep_graph=nothing, kwargs...) where {J,T,RX,HOP,RNG,SS}
 
     # a dependency graph is needed
     if dep_graph === nothing
-        dg = DiffEqJump.make_dependency_graph(num_specs, maj)
+        dg = DiffEqJump.make_dependency_graph(num_specs, rx_rates.ma_jumps)
```
isaacsas, line 31: What is `rx_rates`? How is it getting passed in?

src/spatial/flatten.jl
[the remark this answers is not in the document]
Vilin97: I added a keyword arg `scale_rates`, which is false in case the passed in rates come from `MassActionJump`

src/spatial/flatten.jl
```diff
@@ -0,0 +1,76 @@
...
+                dst_idx = spec_LI[spec_CI[species,dst]]
+                src_idx = spec_LI[spec_CI[species,src]]
+                push!(hop_netstoch, [dst_idx => 1, src_idx => -1])
+                push!(hop_reacstoch, [src_idx => 1])
+                push!(hop_rates, hopping_constants_at_src[species, i])
+            end
+        end
+    end
+
+    total_netstoch = convert(Vector{Vector{Pair{Int, Int}}}, vcat(hop_netstoch, rx_netstoch))
```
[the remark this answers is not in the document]
Vilin97: You are right. I had these because I was getting an error when stoichiometry vectors of type `Any[]` were passed in. But now that I am taking them from `MassActionJump`, there is no error. I deleted these `convert`

src/spatial/flatten.jl
```diff
@@ -0,0 +1,76 @@
...
+    total_reacstoch = convert(Vector{Vector{Pair{Int, Int}}}, vcat(hop_reacstoch, rx_reactstoch))
+    total_rates = convert(Vector{F}, vcat(hop_rates, vec(rx_rates)))
+    ma_jump = MassActionJump(total_rates, total_reacstoch, total_netstoch)
```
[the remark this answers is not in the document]
Vilin97: Done

src/problem.jl
```diff
@@ -70,15 +70,20 @@ JumpProblem(prob,jumps::JumpSet;kwargs...) = JumpProblem(prob,NullAggregator(),j
 
 function JumpProblem(prob, aggregator::AbstractAggregatorAlgorithm, jumps::JumpSet;
                      save_positions = typeof(prob) <: DiffEqBase.AbstractDiscreteProblem ? (false,true) : (true,true),
-                     rng = Xorshifts.Xoroshiro128Star(rand(UInt64)), kwargs...)
+                     rng = Xorshifts.Xoroshiro128Star(rand(UInt64)), spatial_system=nothing, hopping_constants=nothing, kwargs...)
 
+  ## Spatial jumps handling
+  massaction_jump = jumps.massaction_jump
+  if spatial_system !== nothing && hopping_constants !== nothing && !is_spatial(aggregator)
+    prob, massaction_jump = flatten(massaction_jump, prob, spatial_system, hopping_constants; kwargs...)
```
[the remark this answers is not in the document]
Vilin97: It's unclear to me what it means to flatten something that is not a massaction jump. We can just use `flatten` as an intermediate function in the final interface

src/spatial/flatten.jl
```diff
@@ -0,0 +1,76 @@
...
+        for site in 1:num_nodes
+            append!(out, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), stoch))
+        end
+        out
+    end
+    rx_netstoch =  to_spatial(netstoch, num_nodes, spec_LI, spec_CI)
+    rx_reactstoch =  to_spatial(reactstoch, num_nodes, spec_LI, spec_CI)
+
+    #hops
+    hop_netstoch = eltype(netstoch)[]; sizehint!(hop_netstoch, num_nodes*num_species*6)
```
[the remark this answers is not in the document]
Vilin97: I'm not doing that because this need not be a grid. The size of `hop_netstoch` is num_nodes*(sum of outdegrees of all nodes in graph).

src/spatial/utils.jl
```diff
@@ -26,191 +22,383 @@ abstract type AbstractSpatialSystem end
...
 """
-return total hopping rate out of site
+update rates of single species at site
 """
-function total_site_hop_rate(spatial_rates, site_id)
-    spatial_rates.hop_rates_sum[site_id]
+function update_hop_rate!(hop_rates::HopRates2, species::Int, u, site, spatial_system)
+    rates_at_site = hop_rates.rates[site]
+    @inbounds old_rate = sum(@view rates_at_site[species,:])
+    @inbounds rates_at_site[species,:] = (@view hop_rates.hopping_constants[site][species,:]) * u[species,site]
```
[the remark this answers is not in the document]
isaacsas: Ahh, this is not so great a memory layout since you are slicing across a row and this is a frequently called function. Could it be better to transpose so the layout in this case is site by species?

isaacsas (changes requested):

src/spatial/flatten.jl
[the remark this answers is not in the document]
Vilin97: Changed to `sizehint!`ing the correct number

src/spatial/utils.jl
```diff
@@ -26,191 +22,383 @@ abstract type AbstractSpatialSystem end
...
+    dims::NTuple{N, Int} #side lengths of the grid
+    nums_neighbors::Vector{Int}
+    CI::CartesianIndices{N, T}
+    LI::LinearIndices{N, T}
+    offsets::Vector{CartesianIndex{N}}
 end
-
-function CartesianGrid(dimension, linear_size)
-    CartesianGrid{dimension}([linear_size for i in 1:dimension])
+function CartesianGrid1(dims::Tuple)
```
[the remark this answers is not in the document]
Vilin97: Maybe it would make sense to only keep one of the three?

src/spatial/utils.jl
```diff
@@ -26,191 +22,383 @@ abstract type AbstractSpatialSystem end
...
 """
-function hop_rates_at_site end
+total_site_hop_rate(hop_rates::AbstractHopRates, site) = @inbounds hop_rates.sum_rates[site]
+
+############## hopping rates of form L_{s,i} ################
+
+HopRates(hopping_constants::Matrix{F}) where F <: Number = HopRates1(hopping_constants)
+HopRates(hopping_constants::Vector{Matrix{F}}) where F <: Number = HopRates2(hopping_constants)
+
+struct HopRates1{F} <: AbstractHopRates
```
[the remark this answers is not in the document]
Vilin97: Changed to `HopRatesUnifNbr`

src/spatial/utils.jl
```diff
@@ -26,191 +22,383 @@ abstract type AbstractSpatialSystem end
...
-# NOTE: for now assume hopping rate only depends on the site and species (and not the neighbor of the site)
-struct SpatialRates{R} <: AbstractSpatialRates
-    rx_rates::Matrix{R} # rx_rates[i,j] is rate of reaction i at site j
-    hop_rates::Matrix{R} # hop_rates[i,j] is rate of species i at site j
-    rx_rates_sum::Vector{R} # [sum of reaction rates for site 1,...,sum of reaction rates for last site]
-    hop_rates_sum::Vector{R} # [sum of hopping rates for site 1,...,sum of hopping rates for last site]
+"""
+update rates of single species at site
+"""
+function update_hop_rate!(hop_rates::HopRates1, species::Int, u, site, spatial_system)
```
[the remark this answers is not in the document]
Vilin97: Not needed. Fixed

src/spatial/utils.jl
```diff
@@ -26,191 +22,383 @@ abstract type AbstractSpatialSystem end
...
+total_site_hop_rate(hop_rates::AbstractHopRates, site) = @inbounds hop_rates.sum_rates[site]
+
+############## hopping rates of form L_{s,i} ################
+
+HopRates(hopping_constants::Matrix{F}) where F <: Number = HopRates1(hopping_constants)
+HopRates(hopping_constants::Vector{Matrix{F}}) where F <: Number = HopRates2(hopping_constants)
+
+struct HopRates1{F} <: AbstractHopRates
+    hopping_constants::Matrix{F} # hopping_constants[i,j] is the hop constant of species i at site j
+    rates::Matrix{F} # rates[i,j] is rate of reaction i at site j
```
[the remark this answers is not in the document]
Vilin97: fixed

src/spatial/utils.jl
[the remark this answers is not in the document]
Vilin97: Or we could just drop `old_rate` (also, I completely forgot to actually return it). If some other algorithm needs it, it can be computed before calling `update_hop_rate!`

src/spatial/utils.jl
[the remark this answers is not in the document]
Vilin97: I deleted `CartesianGrid3` and renamed all other utility structs to descriptive names

src/spatial/nsm.jl
```diff
@@ -110,18 +108,19 @@ end
...
 
     pqdata = Vector{typeof(t)}(undef, num_sites)
-    for site in 1:num_sites
-        update_reaction_rates!(cur_rates, 1:num_majumps, u, ma_jumps, site)
-        update_hop_rates!(cur_rates, 1:num_species, hopping_constants, u, site, spatial_system)
-        pqdata[site] = t + randexp(aggregation.rng) / total_site_rate(cur_rates, site)
+    @inbounds for site in 1:num_sites
+        update_rx_rates!(rx_rates, 1:num_rxs, u, site)
+        update_hop_rates!(hop_rates, 1:numspecies, u, site, spatial_system)
+        pqdata[site] = t + randexp(rng) / (total_site_rx_rate(rx_rates, site)+total_site_hop_rate(hop_rates, site))
```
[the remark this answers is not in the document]
Vilin97: brought it back

src/spatial/utils.jl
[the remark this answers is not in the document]
Vilin97: I transposed both `hopping_constants` and `rates` internally (the input is still the same) for better memory access.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,96 @@
...
+function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rx_rates::Vector, spatial_system, u0::Matrix{Int}, tspan, hopping_constants::Vector{Matrix{F}}; kwargs...) where F <: Number
+    num_nodes = num_sites(spatial_system)
+    rates = reshape(repeat(rx_rates, num_nodes), length(rx_rates), num_nodes)
+    flatten(netstoch, reactstoch, rates, spatial_system, u0, tspan, hopping_constants; kwargs...)
+end
+
+"""
+convert to Float64 if type Any
+"""
+function flatten(netstoch::AbstractArray{Any}, reactstoch::AbstractArray{Any}, rx_rates::Vector{Any}, spatial_system, u0::Matrix{Int}, tspan, hopping_constants::Vector{Matrix{F}}; kwargs...) where F <: Number
```
isaacsas, line 44: Why type `Any`? We should never by working with stoichiometries that are type any.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,96 @@
...
+    net_stoch = convert(Vector{Vector{Pair{Int64, Int64}}}, netstoch)
+    react_stoch = convert(Vector{Vector{Pair{Int64, Int64}}}, reactstoch)
+    rates = convert(Vector{Float64}, rx_rates)
```
isaacsas, line 47: What are you trying to do here? We don't want to change the type; everything should work fine with Float32s for example.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,96 @@
...
+    num_nodes = num_sites(spatial_system)
+    rates = reshape(repeat(rx_rates, num_nodes), length(rx_rates), num_nodes)
+    flatten(netstoch, reactstoch, rates, spatial_system, u0, tspan, hopping_constants; kwargs...)
+end
+
+"""
+convert to Float64 if type Any
+"""
+function flatten(netstoch::AbstractArray{Any}, reactstoch::AbstractArray{Any}, rx_rates::Vector{Any}, spatial_system, u0::Matrix{Int}, tspan, hopping_constants::Vector{Matrix{F}}; kwargs...) where F <: Number
+    net_stoch = convert(Vector{Vector{Pair{Int64, Int64}}}, netstoch)
```
isaacsas, line 45: Why are all these converts needed?

src/spatial/flatten.jl
```diff
@@ -0,0 +1,96 @@
...
+
+    #reactions
+    for site in 1:num_nodes
+        append!(total_netstoch, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), netstoch))
+        append!(total_reactstoch, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), reactstoch))
+    end
+    append!(total_rates, vec(rx_rates)) # assuming rx_rates isa Matrix where rx_rates[rx, site] is the rate of rx at site
+
+    # put everything together
+    ma_jump = MassActionJump(total_rates, total_reactstoch, total_netstoch; nocopy = true, scale_rates = scale_rates)
```
isaacsas, line 92: If `total_rates` are coming from a user-supplied `MassActionJump`, I think we can assume that they were already scaled in creating it. So here `scale_rates=false` is probably what we want. Sorry for any confusion about this!

src/spatial/flatten.jl
Vilin97: This happens if we do `MassActionJump([], [], [])`, which is done in, say, `diffusion.jl`.

src/spatial/flatten.jl
Vilin97: If the user passes in a `MassActionJump` with arrays of type `Any`, as in `MassActionJump([], [], [])`, this catches that. What would be a better way to do this?

src/spatial/flatten.jl
Vilin97: The method on line 7 (https://github.com/SciML/DiffEqJump.jl/pull/192/files/a330c42c4a8909b909753c8bc32005369201de8f#diff-208a047f849b0efd9cf21d343ed068871f368976846397bc992f78ce89512f9bR7) is the one that takes in a `MassActionJump`. It then calls the main `flatten` method (https://github.com/SciML/DiffEqJump.jl/pull/192/files/a330c42c4a8909b909753c8bc32005369201de8f#diff-208a047f849b0efd9cf21d343ed068871f368976846397bc992f78ce89512f9bR54) with `scale_rates = false`.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,90 @@
...
+if hopping_constants is a matrix, assume hopping_constants[i,j] is species i, site j
+"""
+function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rx_rates::Vector, spatial_system, u0::Matrix{Int}, tspan, hopping_constants::Matrix{F}; kwargs...) where F <: Number
+    num_nodes = num_sites(spatial_system)
+    num_specs = size(u0, 1)
+    @assert size(hopping_constants) == size(u0) # hopping_constants[i,j] is species i, site j
+    hop_constants = Vector{Matrix{F}}(undef, size(hopping_constants, 2))
+    for site in 1:num_nodes
+        num_nbs = num_neighbors(spatial_system, site)
+        hop_constants[site] = reshape(repeat(hopping_constants[:,site], num_nbs), num_specs, num_nbs)
```
isaacsas, line 30: `@view hopping_constants[:,site]`

src/spatial/flatten.jl
```diff
@@ -0,0 +1,90 @@
...
+    
+    flatten(netstoch, reactstoch, rx_rates, spatial_system, u0, tspan, hop_constants; kwargs...)
+end
+
+"""
+if reaction rates is a vector, assume reaction rates are equal across sites
+"""
+function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rx_rates::Vector, spatial_system, u0::Matrix{Int}, tspan, hopping_constants::Vector{Matrix{F}}; kwargs...) where F <: Number
+    num_nodes = num_sites(spatial_system)
+    rates = reshape(repeat(rx_rates, num_nodes), length(rx_rates), num_nodes)
```
isaacsas, line 41: Why does the previous `flatten` also not do this?

src/spatial/flatten.jl
```diff
@@ -0,0 +1,90 @@
...
+    num_species = size(u0, 1)
+    num_nodes = num_sites(spatial_system)
+    num_rxs = length(reactstoch)
+    @assert length(hopping_constants) == size(u0, 2) == size(rx_rates, 2) == num_nodes
+    @assert length(netstoch) == length(reactstoch) == size(rx_rates, 1) == num_rxs
+    spec_CI = CartesianIndices((num_species, num_nodes))
+    spec_LI = LinearIndices((num_species, num_nodes))
+
+    sum_outdegrees = sum(num_neighbors(spatial_system, site) for site in 1:num_nodes)
+    num_jumps = num_species*(sum_outdegrees + num_rxs)
```
isaacsas, line 58: Why is there a `num_species*num_rxs` term and not just `num_rxs`?

src/spatial/flatten.jl
```diff
@@ -0,0 +1,90 @@
...
+                push!(total_netstoch, [dst_idx => 1, src_idx => -1])
+                push!(total_reactstoch, [src_idx => 1])
+                push!(total_rates, hopping_constants_at_src[species, i])
+            end
+        end
+    end
+
+    #reactions
+    for site in 1:num_nodes
+        append!(total_netstoch, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), netstoch))
```
isaacsas, line 80: Can this be written a bit more efficiently as a `foreach` that `push!` a single `map` call instead of appending two map calls?

isaacsas: Again, I'm just trying to keep memory use down since it can grow quickly in flattening.

src/spatial/flatten.jl
```diff
@@ -0,0 +1,90 @@
...
+        append!(total_reactstoch, map(rx_stoch -> map(p -> Pair(spec_LI[spec_CI[p[1],site]], p[2]), rx_stoch), reactstoch))
+    end
+    append!(total_rates, vec(rx_rates)) # assuming rx_rates isa Matrix where rx_rates[rx, site] is the rate of rx at site
```
isaacsas, line 83: But doesn't the second `flatten` version above leave `rx_rates` a vector?

src/spatial/nsm.jl
```diff
@@ -110,18 +108,19 @@ end
...
 
-    pqdata = Vector{typeof(t)}(undef, num_sites)
-    for site in 1:num_sites
-        update_reaction_rates!(cur_rates, 1:num_majumps, u, ma_jumps, site)
-        update_hop_rates!(cur_rates, 1:num_species, hopping_constants, u, site, spatial_system)
-        pqdata[site] = t + randexp(aggregation.rng) / total_site_rate(cur_rates, site)
+    num_reactions = num_rxs(rx_rates)
+    num_nodes = num_sites(spatial_system)
+
+    pqdata = Vector{typeof(t)}(undef, num_nodes)
```
isaacsas, line 119: Should this be reset too?

src/spatial/nsm.jl
```diff
@@ -139,118 +138,25 @@ function update_dependent_rates_and_firing_times!(p, u, t)
...
-            update_site_time!(p.pq, p.rng, p.cur_rates, site, t)
-        end
+        update_site_time!(p, source_site, t)
+        update_site_time!(p, target_site, t)
     else
         site = jump.src
         update_rates_after_reaction!(p, u, site, reaction_id_from_jump(p,jump))
-        update_site_time!(p.pq, p.rng, p.cur_rates, site, t)
+        update_site_time!(p, site, t)
     end
```
isaacsas, line 147: return `nothing`

src/spatial/utils.jl
```diff
@@ -2,6 +2,7 @@
 A file with types, structs and functions for spatial simulations
 """
 
+using LightGraphs
```
isaacsas, line 5: This should be in DiffEqJump.jl not here.

src/spatial/utils.jl
```diff
@@ -26,191 +22,358 @@ abstract type AbstractSpatialSystem end
...
+
+############## hopping rates of form L_{s,i} ################
+
+HopRates(hopping_constants::Matrix{F}) where F <: Number = HopRatesUnifNbr(hopping_constants)
+HopRates(hopping_constants::Vector{Matrix{F}}) where F <: Number = HopRatesGeneral(hopping_constants)
+
+struct HopRatesUnifNbr{F} <: AbstractHopRates
+    hopping_constants::Matrix{F} # hopping_constants[i,j] is the hop constant of species i at site j
+    rates::Matrix{F} # rates[i,j] is total hopping rate of species i at site j
+    sum_rates::Vector{F} # sum_rates[j] is the sum of reaction rates at site j
```
isaacsas, line 210: sum of the hopping rates?

src/spatial/utils.jl
```diff
@@ -26,191 +22,358 @@ abstract type AbstractSpatialSystem end
...
 """
-function SpatialRates(ma_jumps, num_species, spatial_system::AbstractSpatialSystem)
-    num_sites = num_sites(spatial_system)
-    num_jumps = get_num_majumps(ma_jumps)
-    SpatialRates(num_jumps,num_species,num_sites)
+function HopRatesGeneral(hopping_constants::Vector{Matrix{F}}, transpose = true) where F <: Number
+    if transpose
+        hopping_constants = map(r -> r = collect(r'), hopping_constants) #transpose for better memory layout
+    end
+    rates = deepcopy(hopping_constants)
```
isaacsas, line 282: For future PR we should optimize this to just assume they are all in the right layout from the user (to minimize copying internally).

src/spatial/utils.jl
```diff
@@ -26,191 +22,358 @@ abstract type AbstractSpatialSystem end
...
-return total hopping rate out of site
+update rates of single species at site
 """
-function total_site_hop_rate(spatial_rates, site_id)
-    spatial_rates.hop_rates_sum[site_id]
+function update_hop_rate!(hop_rates::HopRatesGeneral, species, u, site, spatial_system)
+    rates_at_site = hop_rates.rates[site]
+    @inbounds old_rate = sum(@view rates_at_site[:,species])
+    @inbounds @. rates_at_site[:,species] = (@view hop_rates.hopping_constants[site][:,species]) * u[species,site]
+    @inbounds hop_rates.sum_rates[site] += sum(@view rates_at_site[:,species]) - old_rate
```
isaacsas, line 314: Maybe a better way to optimize this is to store the partial sums of the hopping constants instead of the constants themselves, and then use those for sampling the next direction? Then we just multiply their already stored total sum, which is fixed, by the new amount of the species to update the `sum_rate`? Wouldn't that avoid all these vector operations?

isaacsas (changes requested):

src/spatial/flatten.jl
Vilin97: It was a typo. I now changed from `rx_rates::Vector` to `rx_rates::AbstractArray` in the previous `flatten`.

src/spatial/flatten.jl
Vilin97: Right. It should be `num_nodes * num_rxs` instead. Fixed.

src/spatial/flatten.jl
Vilin97: The second `flatten` is supposed to call the third `flatten` if `rx_rates` is a vector.

src/spatial/nsm.jl
Vilin97: What do you mean? `pqdata` is not stored in the aggregation, so it cannot be reset.

isaacsas: I think it is stored in the priority queue though, so you might be able to reuse it. We can save that for the next PR.

PR #193, opened 2021-07-16
MassActionJump param update functor

isaacsas:
This allows users to give a functor struct that corresponds to the parameter update mapper. If the user just gives the indices of each parameter in `p` via `param_idxs`, then a default one is automatically constructed. With this I think we can make an analogous struct for JumpSystems in ModelingToolkit that will allow updates when some of the MassActionRates are actually symbolic functions.


test/ssa_callback_test.jl
```diff
@@ -66,4 +66,21 @@ function paffect!(integrator)
   reset_aggregated_jumps!(integrator)
 end
 sol = solve(jprob, SSAStepper(), tstops=[1000.0], callback=DiscreteCallback(pcondit,paffect!))
-@test sol[1,end] == 100
\ No newline at end of file
+@test sol[1,end] == 100
+
+maj1 = MassActionJump([1 => 1],[1 => -1, 2 => 1]; param_idxs=1, params=p)
```
ChrisRackauckas, line 71: I don't quite see why each MAJ needs a pointer to `p`.

isaacsas: So it can determine the rate constant to store when being constructed (`p` isn’t actually saved internally). I guess we could store a dummy value and initialize it later, though we’d still need to know the rate’s ultimate type at the time the jump is constructed.

isaacsas: So an alternative would be to optionally allow the rate field type to be passed, with a default of Float64, and then initialize the jump rate constant in the JumpProblem constructor.

ChrisRackauckas: If you use `@set!` from SetField, the initialization can be to just `nothing` and it should resolve in the solve. Since it's not mutable, that should compile to be free.

isaacsas: OK, I updated this so that if `p` is not passed then in the `JumpProblem` constructor `update_params!` is called to fill in the MassActionJump's rate constants from `p`.

isaacsas: Would it be better to resolve in `solve` or in building the `JumpProblem` (which is what I had updated too)?

ChrisRackauckas: Yeah, prob makes more sense.

src/problem.jl
```diff
@@ -72,13 +72,22 @@ function JumpProblem(prob, aggregator::AbstractAggregatorAlgorithm, jumps::JumpS
                      save_positions = typeof(prob) <: DiffEqBase.AbstractDiscreteProblem ? (false,true) : (true,true),
                      rng = Xorshifts.Xoroshiro128Star(rand(UInt64)), kwargs...)
 
+  # initialize the MassActionJump rate constants with the user parameters
```
isaacsas, line 75: @ChrisRackauckas this aliases all the non-rate constant fields while deferring constructing the rates until `JumpProblem`. Prior to that I just set the rate field to be `nothing` as you suggested.

ChrisRackauckas:
Allowing general kwargs is something I want to do less of 😅

isaacsas:
I think it is needed here for some callers. Otherwise the JumpProblem in ModelingTollkit JumpSystems would need to know all the various MassActionJump parameters.

isaacsas:
Is there a performance reason to leave it out?

ChrisRackauckas:
No it's a correctness thing. Once you allow it, then even incorrect parameters don't throw errors: https://github.com/SciML/DifferentialEquations.jl/issues/651

ChrisRackauckas:
How many parameters? Could they instead be listed out?

isaacsas:
Three kwargs I think that would need to be put into any JumpProblem constructors now, both here and in ModelingToolkot.

ChrisRackauckas:
Yeah I would prefer to do that. And in general, my goal is to eliminate the solver-level kwarg splatting ASAP.

isaacsas:
@ChrisRackauckas I'll plan to merge once tests pass unless you have any other objections.

PR #195, opened 2021-07-27
added `reset!` for `PriorityTable`

Vilin97:
Wrote a `reset!` function for the `PriorityTable` used in `DirectCR` and `RSSACR`. This slightly reduces allocations per run, since a new table does not get allocated every time. Here is a benchmark for before and after. There is a small improvement
image

### DIFF
--- a/src/aggregators/directcr.jl
+++ b/src/aggregators/directcr.jl
@@ -57,8 +57,7 @@ function DirectCRJumpAggregation(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T,
     ratetogroup = rate -> priortogid(rate, minexponent)
 
     # construct an empty initial priority table -- we'll reset this in init
-    num_jumps = length(crs)
-    rt = PriorityTable(ratetogroup, zeros(T, num_jumps), minrate, 2*minrate)
+    rt = PriorityTable(ratetogroup, zeros(T, 1), minrate, 2*minrate)
 
     DirectCRJumpAggregation{T,S,F1,F2,RNG,typeof(dg),typeof(rt),typeof(ratetogroup)}(
                                             nj, nj, njt, et, crs, sr, maj, rs, affs!, sps, rng,
--- a/src/aggregators/directcr.jl
+++ b/src/aggregators/directcr.jl
@@ -56,11 +56,9 @@ function DirectCRJumpAggregation(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T,
     minrate = 2.0^minexponent
     ratetogroup = rate -> priortogid(rate, minexponent)
 
-    # construct an empty initial priority table -- we'll overwrite this in init anyways...
-    rt = PriorityTable{T,Int,Int,typeof(ratetogroup)}(minrate, 2*minrate,
-                                                        Vector{PriorityGroup{T,Vector{Int}}}(),
-                                                        Vector{T}(), zero(T),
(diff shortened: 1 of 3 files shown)


src/aggregators/directcr.jl
```diff
@@ -56,11 +56,9 @@ function DirectCRJumpAggregation(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T,
...
     ratetogroup = rate -> priortogid(rate, minexponent)
 
-    # construct an empty initial priority table -- we'll overwrite this in init anyways...
-    rt = PriorityTable{T,Int,Int,typeof(ratetogroup)}(minrate, 2*minrate,
-                                                        Vector{PriorityGroup{T,Vector{Int}}}(),
-                                                        Vector{T}(), zero(T),
-                                                        Vector{Tuple{Int,Int}}(), ratetogroup)
+    # construct an empty initial priority table -- we'll reset this in init
+    num_jumps = length(crs)
+    rt = PriorityTable(ratetogroup, zeros(T, num_jumps), minrate, 2*minrate)
```
isaacsas, line 61: Will this make the first bucket really big in memory (length num_jumps), and hold that memory forever?

isaacsas: (since you don't reset the memory used by the table in `reset!`?)

src/aggregators/directcr.jl
```diff
@@ -87,11 +85,11 @@ function initialize!(p::DirectCRJumpAggregation, integrator, u, params, t)
     # initialize rates
     fill_rates_and_sum!(p, u, params, t)
 
-    # if no maxrate was set, use largest initial rate (pad by 2 for an extra group)
-    isinf(p.maxrate) && (p.maxrate = 2*maximum(p.cur_rates))
-
     # setup PriorityTable
-    p.rt   = PriorityTable(p.ratetogroup, p.cur_rates, p.minrate, p.maxrate)
```
isaacsas, line 94: How does the initial table change here? Is it the same (i.e. the same set of buckets)?

Vilin97: Yes, the same set of buckets. If these buckets were created in the first run, chances are the same buckets will be needed for other runs. So the idea is to just empty all the buckets and fill them again instead of recreating the whole table

isaacsas:
Could you run an example, like the BCR, both with and without the PR just to confirm the running time is not changed at all?

src/aggregators/directcr.jl
Vilin97: Yes, it will. And that's bad. I'll change that.

src/aggregators/directcr.jl
```diff
@@ -87,11 +84,11 @@ function initialize!(p::DirectCRJumpAggregation, integrator, u, params, t)
     # initialize rates
     fill_rates_and_sum!(p, u, params, t)
 
-    # if no maxrate was set, use largest initial rate (pad by 2 for an extra group)
-    isinf(p.maxrate) && (p.maxrate = 2*maximum(p.cur_rates))
```
isaacsas, line 91: I went to merge, and noticed that it now appears we don't actually use the `p.maxrate` with the updates. Should we just remove that field from the aggregation? Otherwise this is good to go.

isaacsas: Actually, we can remove this later on. Maybe we want to tweak this down the line to support using that field. So I'll merge as is now.

PR #198, opened 2021-08-02
Improved `HopRatesGeneral`

Vilin97:
Now `HopRatesGeneral` uses cumulative sums of hopping constants avoiding unnecessary summations and vector operations. Also improved the interface so that hopping constants are now passed in as Matrix{Vector}, with [s,i] being vector of hopping constants of species s at site i to its neighbors

### DIFF
--- a/src/spatial/flatten.jl
+++ b/src/spatial/flatten.jl
@@ -25,7 +25,7 @@ function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rx_rates::A
     hop_constants = Matrix{Vector{F}}(undef, size(hopping_constants))
     for ci in CartesianIndices(hop_constants)
         (species, site) = Tuple(ci)
-        hop_constants[ci] = repeat([hopping_constants[species, site]], num_neighbors(spatial_system, site))
+        hop_constants[ci] = hopping_constants[species, site] * ones(num_neighbors(spatial_system, site))
     end
     flatten(netstoch, reactstoch, rx_rates, spatial_system, u0, tspan, hop_constants; kwargs...)
 end
--- a/src/spatial/flatten.jl
+++ b/src/spatial/flatten.jl
@@ -33,7 +33,7 @@ end
 """
 if reaction rates is a vector, assume reaction rates are equal across sites
 """
-function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rx_rates::Vector, spatial_system, u0::Matrix{Int}, tspan, hopping_constants::Vector{Matrix{F}}; kwargs...) where F <: Number
+function flatten(netstoch::AbstractArray, reactstoch::AbstractArray, rx_rates::Vector, spatial_system, u0::Matrix{Int}, tspan, hopping_constants::Matrix{Vector{F}}; kwargs...) where F <: Number
     num_nodes = num_sites(spatial_system)
     rates = reshape(repeat(rx_rates, num_nodes), length(rx_rates), num_nodes)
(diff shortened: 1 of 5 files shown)


src/spatial/flatten.jl
```diff
@@ -18,25 +18,22 @@ function flatten(ma_jump, prob::DiscreteProblem, spatial_system, hopping_constan
...
-    @assert size(hopping_constants) == size(u0) # hopping_constants[i,j] is species i, site j
-    hop_constants = Vector{Matrix{F}}(undef, size(hopping_constants, 2))
-    for site in 1:num_nodes
-        num_nbs = num_neighbors(spatial_system, site)
-        hop_constants[site] = reshape(repeat((@view hopping_constants[:,site]), num_nbs), num_specs, num_nbs)
+    @assert size(hopping_constants) == size(u0)
+    hop_constants = Matrix{Vector{F}}(undef, size(hopping_constants))
+    for ci in CartesianIndices(hop_constants)
+        (species, site) = Tuple(ci)
+        hop_constants[ci] = repeat([hopping_constants[species, site]], num_neighbors(spatial_system, site))
```
isaacsas, line 28: Why not
```
hop_constants[ci] = hopping_constants[species, site] * ones(num_neighbors(spatial_system, site))
```

src/spatial/utils.jl
```diff
@@ -266,29 +266,26 @@ end
 
 ############## hopping rates of form L_{s,i,j} ################
 struct HopRatesGeneral{F} <: AbstractHopRates
-    hopping_constants::Vector{Matrix{F}} # hopping_constants[i][j,s] is the hopping constant at site i of species s to jth neighbor
-    rates::Vector{Matrix{F}} # rates[i][j,s] is the hopping rate at site i of species s to jth neighbor
+    hop_const_cumulative_sums::Matrix{Vector{F}} # hop_const_cumulative_sums[s,i] is the vector of cumulative sums of hopping constants of species s at site i
```
isaacsas, line 269: Why not use an `Array{F,3}` where the first index ranges over the cumulative sums of the neighbors, then species and then location? (Or perhaps location than species if that makes more sense.)

src/spatial/utils.jl
```diff
@@ -297,20 +294,20 @@ end
...
 update rates of single species at site
 """
 function update_hop_rate!(hop_rates::HopRatesGeneral, species, u, site, spatial_system)
-    rates_at_site = hop_rates.rates[site]
-    @inbounds old_rate = sum(@view rates_at_site[:,species])
-    @inbounds @. rates_at_site[:,species] = (@view hop_rates.hopping_constants[site][:,species]) * u[species,site]
-    @inbounds hop_rates.sum_rates[site] += sum(@view rates_at_site[:,species]) - old_rate
+    rates = hop_rates.rates
+    @inbounds old_rate = rates[species, site]
+    rates[species, site] = u[species, site] * hop_rates.hop_const_cumulative_sums[species, site][end]
```
isaacsas, line 309: `@inbounds`

isaacsas (changes requested):
I left you a few comments. Main one is what Matrix{Vector} instead of Array{F,3}?

src/spatial/utils.jl
Vilin97: Because the number of neighbors is different for different sites

isaacsas: Yeah, I was thinking structured grids again…

isaacsas (approved):

PR #199, opened 2021-08-04
made `tstops` type-stable in `SSAStepper`

Vilin97:


src/SSA_stepper.jl
```diff
@@ -75,13 +75,14 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
                          alias_jump = Threads.threadid() == 1,
                          saveat = nothing,
                          callback = nothing,
-                         tstops = [],
```
isaacsas, line 78: This needs to stay a keyword argument, but you could set it to
```julia
tstops = eltype(prob.prob.tspan)[]
```

isaacsas (changes requested):

isaacsas (approved):

src/SSA_stepper.jl
```diff
@@ -75,13 +75,12 @@ function DiffEqBase.__init(jump_prob::JumpProblem,
...
+                         tstops = eltype(jump_prob.prob.tspan)[]
```
isaacsas, line 78: Needs a comma

PR #200, opened 2021-08-05
Import initialize!

ChrisRackauckas:
```julia
julia> pop = read_pumas(pkdata,
                       observations = [:dv], 
                       covariates = [:age, :wt, :doselevel, :isPM, :isfed, :sex])
Population
  Subjects: 18
  Covariates: age, wt, doselevel, isPM, isfed, sex
  Observations: WARNING: both OrdinaryDiffEq and DiffEqJump export "initialize!"; uses of it in module Pumas must be qualified
dv
```

Some dispatch must be shadowing instead of extending DiffEqBase, so this is a quick fix.

### DIFF
--- a/src/DiffEqJump.jl
+++ b/src/DiffEqJump.jl
@@ -8,7 +8,7 @@ using DataStructures, PoissonRandom, Random, ArrayInterface
 using FunctionWrappers, UnPack
 using LightGraphs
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
CI didn't trigger?

PR #202, opened 2021-08-06
Now the general form of hopping rates with a cartesian grid is faster

Vilin97:
wrote HopRatesGeneralGrid -- an optimized version of `HopRatesGeneral` for cartesian grids. It takes about twice less time to sample a neighbor now.

### DIFF
--- a/src/spatial/hop_rates.jl
+++ b/src/spatial/hop_rates.jl
@@ -4,6 +4,12 @@ A file with structs and functions for sampling hops and updating hopping rates
 
 ### spatial hop rates ###
 abstract type AbstractHopRates end
+
+HopRates(hopping_constants::Matrix{F}, spatial_system) where F <: Number = HopRatesUnifNbr(hopping_constants)
+HopRates(hopping_constants::Matrix{F}, grid::Union{CartesianGridRej, CartesianGridIter}) where F <: Number = HopRatesUnifNbr(hopping_constants)
+HopRates(hopping_constants::AbstractArray, spatial_system) = HopRatesGeneral(hopping_constants)
+HopRates(hopping_constants::AbstractArray, grid::Union{CartesianGridRej, CartesianGridIter}) = HopRatesGeneralGrid(hopping_constants, grid)
+
 """
     update_hop_rates!(hop_rates::AbstractHopRates, species::AbstractArray, u, site, spatial_system)
 
@@ -17,19 +23,17 @@ end
 
 """
     total_site_hop_rate(hop_rates::AbstractHopRates, site)
-    
+
 return total hopping rate out of site
 """
 total_site_hop_rate(hop_rates::AbstractHopRates, site) = @inbounds hop_rates.sum_rates[site]
 
 ############## hopping rates of form L_{s,i} ################
 
-HopRates(hopping_constants::Matrix{F}) where F <: Number = HopRatesUnifNbr(hopping_constants)
-HopRates(hopping_constants::AbstractArray) where F <: Number = HopRatesGeneral(hopping_constants)
 
 struct HopRatesUnifNbr{F} <: AbstractHopRates
     "hopping_constants[i,j] is the hop constant of species i at site j"
-    hopping_constants::Matrix{F} 
+    hopping_constants::Matrix{F}
 
     "rates[i,j] is total hopping rate of species i at site j"
     rates::Matrix{F}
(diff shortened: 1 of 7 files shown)


Vilin97:
I'm done working on this. Ready to merge once reviewed.

src/spatial/topology.jl
```diff
@@ -61,6 +69,24 @@ function neighbors(grid, site)
     Iterators.map(off -> LI[off+I], Iterators.filter(off -> off+I in CI, grid.offsets))
 end
 
+"""
+given a vector of hopping constants of length num_nbs(grid, site), return a vector of size 2*(dimension of grid) with zeros at indices where the neighbor is out of bounds
```
isaacsas, line 73: of length `num_neighbors...`?

src/spatial/topology.jl
```diff
@@ -61,6 +69,24 @@ function neighbors(grid, site)
...
+"""
+function pad_hop_vec(grid, site, hop_vec::Vector{F}) where F <: Number
```
isaacsas, line 75: `pad_hop_vec!` and have it take the vector to fill in as the first input?

src/spatial/hop_rates.jl
```diff
@@ -168,7 +172,84 @@ update rates of single species at site
...
+    rates = zeros(F, size(hopping_constants)[2:3])
+    sum_rates = vec(sum(rates, dims=1))
+    HopRatesGeneralGrid{F}(hop_const_cumulative_sums, rates, sum_rates)
+end
+
+function HopRatesGeneralGrid(hopping_constants::Matrix{Vector{F}}, grid) where F <: Number
+    new_hopping_constants = Array{F, 3}(undef, 2*dimension(grid), size(hopping_constants)...)
+    for ci in CartesianIndices(hopping_constants)
+        species, site = Tuple(ci)
+        new_hopping_constants[:, species, site] = pad_hop_vec(grid, site, hopping_constants[ci])
```
isaacsas, line 216: `pad_hop_vec!(@view new_hopping_constant[:,species,site], grid, site,...)` ?

src/spatial/topology.jl
```diff
@@ -29,9 +29,17 @@ function potential_offsets(dimension::Int)
...
 end
 
+dimension(grid) = length(grid.dims)
 num_sites(grid) = prod(grid.dims)
 num_neighbors(grid, site) = grid.nums_neighbors[site]
 
+"""
+    nth_potential_nbr(grid, site, n)
+
+return nth potential neighbor (it might be out of bounds)
```
isaacsas, line 39: Maybe just say returns the `nth` Cartesian grid neighbor ignoring boundaries?

src/spatial/topology.jl
```diff
@@ -61,6 +69,24 @@ function neighbors(grid, site)
...
+"""
+function pad_hop_vec(grid, site, hop_vec::Vector{F}) where F <: Number
+    length(hop_vec) == 2*dimension(grid) && return hop_vec
+    new_hop_vec = zeros(2*dimension(grid))
+    CI = grid.CI; offsets = grid.offsets
+    @inbounds I = CI[site]
+    nbr_counter = 1
+    for (i,off) in enumerate(grid.offsets)
+        if I + off in CI
+            new_hop_vec[i] = hop_vec[nbr_counter]
```
isaacsas, line 83: @inbounds

src/spatial/hop_rates.jl
```diff
@@ -168,7 +172,84 @@ update rates of single species at site
...
+    println(io, "HopRates with $num_specs species and $num_sites sites, optimized for CartesianGrid. \nHopping constants of form L_{s,i,j} where s is species, i is source and j is destination.")
+end
+
+"""
+    HopRatesGeneralGrid(hopping_constants::Matrix{Vector{F}}) where F <: Number
+
+initializes HopRates with zero rates
+"""
+function HopRatesGeneralGrid(hopping_constants::Array{F, 3}) where F <: Number
+    hop_const_cumulative_sums = mapslices(cumsum, hopping_constants, dims = 1)
```
isaacsas, line 206: For allocation reasons perhaps we should require the `cumsum`'ed hopping constants to be passed in, or at least add a keyword argument that lets users pre-cumsum them to save memory. (This should hold for each type where possible.)

src/spatial/hop_rates.jl
```diff
@@ -168,7 +172,84 @@ update rates of single species at site
...
+    rates = zeros(F, size(hopping_constants)[2:3])
+    sum_rates = vec(sum(rates, dims=1))
```
isaacsas, line 208: Why do you need a sum here if you just set `rates` to be all zeros?

src/spatial/hop_rates.jl
isaacsas: Can we do the `cumsum` in place here before calling the one-argument version to save memory?

isaacsas: In fact, since this is all internal why make a one-argument version? Just finish the code off inside this version...

isaacsas (changes requested):

src/spatial/hop_rates.jl
Vilin97: I put `cumsum!` in here to avoid the extra allocation. I think having two versions is convenient, e.g. for testing in `test_utils.jl`. We can also expose the main constructor to the user because it might be more convenient to pass in `Array{F, 3}` instead of `Matrix{Vector{F}}` for `hopping_constants`.

src/spatial/topology.jl
Vilin97: This is in a constructor, which will only be called once at aggregation step. Do we really need `@inbounds` for things that are called once?

src/spatial/hop_rates.jl
Vilin97: Done. Using kwarg `do_cumsum = true` now.

src/spatial/hop_rates.jl
Vilin97: I don't. Changed.

isaacsas (approved):

PR #208, opened 2021-08-21
New priority table

Vilin97:
Rewrote the `PriorityTable` to be faster. The image shows the comparison of the old `PriorityTable` (PT) with the new one (PT2), as well as mutable binary heap. 

image


isaacsas:
Could you put your table benchmark script in a Gist?

just a heads up; I want to play around with what is going on here and PT design a bit, so I’m not going to merge this immediately. In particular, I want to see if this speed up can be recapitulated just by making the groups immutable. I also want to isolate what in the table is giving the (logarithmic?) scaling we see in the full spatial benchmarks. If it is from sampling the next event that suggests we may need a different strategy for bin management.

isaacsas:
For example, maybe we want to keep track of the index of the last non-empty group and start searching from it (if searching is part of what is slow).

Vilin97:
I added the Gist with the benchmarking script in the description of the PR.

Vilin97:
Some observations after experimenting with the model from Sanft: 

1. The number of bins is independent of the mesh size
2. The time to sample and the time to update rates both grow with the number of sites as evident from the plot below.
3. There is about one empty group on average, and three groups are searched on average. The total number of groups is around 54, so both of those are negligible.

image

PR #219, opened 2022-02-08
working on spatially varying reaction rates, WIP

Vilin97:
Want to be able to handle spatially varying reaction rates.


src/spatial/reaction_rates.jl
```diff
@@ -52,13 +52,20 @@ end
...
     ma_jumps = rx_rates.ma_jumps
     @inbounds for rx in rxs
         set_rx_rate_at_site!(rx_rates, site, rx, evalrxrate((@view u[:,site]), rx, ma_jumps))
     end
 end
 
+function update_rx_rates!(rx_rates::RxRates{F,M}, rxs, u, site) where {F, M <: SpatialMassActionJump}
+    ma_jumps = rx_rates.ma_jumps
+    @inbounds for rx in rxs
+        set_rx_rate_at_site!(rx_rates, site, rx, evalrxrate((@view u[:,site]), rx, ma_jumps, site))
```
isaacsas, line 65: We no longer need the view with the custom version of `evalrxrate` right? Couldn't we pass it the whole matrix `u` and just have it use `u[spec,site]` in the `SpatialMassActionJump` version?

src/spatial/spatial_massaction_jump.jl
```diff
@@ -0,0 +1,39 @@
+struct SpatialMassActionJump{T,S,U,V} <: AbstractMassActionJump
+    scaled_rates::T
```
isaacsas, line 2: What about the hybrid case we discussed with both a vector and matrix for splitting up the reactions?

src/spatial/spatial_massaction_jump.jl
```diff
@@ -0,0 +1,39 @@
...
+function SpatialMassActionJump(ma_jumps::MassActionJump{T,S,U,V}, rx_coefficients::Nothing) where {T,S,U,V}
+    SpatialMassActionJump{T, S, U, V}(ma_jumps.scaled_rates, ma_jumps.reactant_stoch, ma_jumps.net_stoch, ma_jumps.param_mapper)
+end
+    
+get_num_majumps(spatial_majump::SpatialMassActionJump) = size(spatial_majump.scaled_rates, 1)
+rate_at_site(rx, site, spatial_majump::SpatialMassActionJump{T,S,U,V} where {T <: Matrix, S, U, V}) = spatial_majump.scaled_rates[rx, site]
+rate_at_site(rx, site, spatial_majump::SpatialMassActionJump{T,S,U,V} where {T <: Vector, S, U, V}) = spatial_majump.scaled_rates[rx]
+using_params(spatial_majump::SpatialMassActionJump) = false
+
+function evalrxrate(speciesvec::AbstractVector{T}, rxidx::S, majump::SpatialMassActionJump, site::Int) where {T,S}
```
isaacsas, line 27: Now pass a matrix for `speciesvec`?

test/spatial/ABC.jl
```diff
@@ -24,6 +24,9 @@ netstoch = [[1 => -1, 2 => -1, 3 => 1],[1 => 1, 2 => 1, 3 => -1]]
 rates = [0.1/mesh_size, 1.]
 majumps = MassActionJump(rates, reactstoch, netstoch)
 
+rx_coefficients = ones(get_num_majumps(majumps), num_nodes)
```
isaacsas, line 27: Can we get test coverage for the different mass action jump variants, constant in space via a vector, non-constant in space via a matrix, and a hyrbid of the two?

It would be good to have one accuracy test for each case to ensure no bugs.

src/spatial/spatial_massaction_jump.jl
```diff
@@ -0,0 +1,39 @@
+struct SpatialMassActionJump{T,S,U,V} <: AbstractMassActionJump
+    scaled_rates::T
+    reactant_stoch::S
+    net_stoch::U
+    param_mapper::V
+end
+
+# rx_coefficients[i,j] is the coefficient of reaction i at site j by which the reaction rate will be multiplied
+function SpatialMassActionJump(ma_jumps::MassActionJump{T,S,U,V}, rx_coefficients::Matrix) where {T,S,U,V}
```
isaacsas, line 9: I'm confused about the use case for this constructor. If the user gives an existing `MassActionJump`, the `SpatialMassActionJump` shouldn't have any spatial matrix but just copy or alias the rate vector from the `MassActionJump`. (In fact, we should have a `nocopy=false` kwarg to allow control on whether these are copied over.)

Then, we should also have a separate `SpatialMassActionJump` constructor, analogous to the `MassActionJump` constructor, which has similar keywords and takes the various fields directly (along with scaling the rates). (Though we can ignore the parameter mapping stuff for now.)

isaacsas (changes requested):

src/spatial/reaction_rates.jl
Vilin97: We can. What is the benefit of doing it that way?

isaacsas: Avoids creating a `view`, which in the past had some overhead / performance cost (though it has been improved a lot over time). It just doesn't seem needed, so why keep it. This is the spatial version of `evalrxrate`, so it makes sense for that version to know that `u` is a matrix anyways.

src/spatial/spatial_massaction_jump.jl
isaacsas: Three cases we'd like `SpatialMassActionJump` to support:
1. `scaled_rates == Vector` and `scaled_spatial_rates == Nothing`.
2. `scaled_rates == Nothing` and `scaled_spatial_rates == Matrix`.
3. `scaled_rates == Vector` and `scaled_spatial_rates == Matrix`.
Note, we don't need to call the last field `scaled_spatial_rates`, that is just a placeholder. A more compact but representative name would be good.

src/spatial/spatial_massaction_jump.jl
isaacsas: Isn't this just changing `speciesvec[specstoch[1]]` to `speciesmat[specstoch[1],site]` on one line?

src/spatial/reaction_rates.jl
```diff
@@ -52,13 +52,20 @@ end
 
 update rates of all reactions in rxs at site
 """
-function update_rx_rates!(rx_rates, rxs, u, site)
+function update_rx_rates!(rx_rates::RxRates{F,M}, rxs, u, site) where {F, M <: MassActionJump}
```
isaacsas, line 55: Why do we still need this version? Shouldn't we always just `SpatialMassActionJump`s internally for  the spatial solvers now?

src/spatial/spatial_massaction_jump.jl
```diff
@@ -0,0 +1,65 @@
...
+SpatialMassActionJump(ma_jumps::MassActionJump{T,S,U,V}; scale_rates = true, useiszero = true, nocopy=false) where {T,S,U,V} = SpatialMassActionJump(ma_jumps.scaled_rates, ma_jumps.reactant_stoch, ma_jumps.net_stoch, ma_jumps.param_mapper; scale_rates = scale_rates, useiszero = useiszero, nocopy=nocopy)
+
+##############################################
+
+get_num_majumps(spatial_majump::SpatialMassActionJump) = length(spatial_majump.uniform_rates) + size(spatial_majump.spatial_rates, 1)
+using_params(spatial_majump::SpatialMassActionJump) = false
+
+function rate_at_site(rx, site, spatial_majump::SpatialMassActionJump)
+    num_unif_rxs = length(spatial_majump.uniform_rates)
+    rx <= num_unif_rxs ? spatial_majump.uniform_rates[rx] : spatial_majump.spatial_rates[rx-num_unif_rxs, site]
```
isaacsas, line 50: Weren't we going to allow one of `uniform_rates` or `spatial_rates` to be `nothing`, so dispatch could avoid having to branch here if just one of them is being used?

src/spatial/spatial_massaction_jump.jl
```diff
@@ -0,0 +1,65 @@
...
+"""
+uniform rates go first in ordering
+"""
+    function SpatialMassActionJump{F,S,U,V}(uniform_rates::Vector{F}, spatial_rates::Matrix{F}, reactant_stoch::S, net_stoch::U, param_mapper::V, scale_rates::Bool, useiszero::Bool, nocopy::Bool) where {F, S, U, V}
+        uniform_rates = nocopy ? uniform_rates : copy(uniform_rates)
+        spatial_rates = nocopy ? spatial_rates : copy(spatial_rates)
+        reactant_stoch = nocopy ? reactant_stoch : copy(reactant_stoch)
+        for i in eachindex(reactant_stoch)
+            if useiszero && (length(reactant_stoch[i]) == 1) && iszero(reactant_stoch[i][1][1])
+            reactant_stoch[i] = typeof(reactant_stoch[i])()
```
isaacsas, line 19:
```suggestion
                reactant_stoch[i] = typeof(reactant_stoch[i])()
```

isaacsas (changes requested):

src/spatial/reaction_rates.jl
Vilin97: `MassActionJump` has more utilities like parameter mapping and merging so I thought it's good to leave it as a sort of backward compatibility

PR #220, opened 2022-02-20
Change default rng on 1.7 up

isaacsas:


ChrisRackauckas (approved):

ChrisRackauckas:
Did you get a rough estimate of how much faster this is now?

Should DiffEqNoiseProcess.jl also change its defaults (for SDEs)?

isaacsas:
I haven't benchmarked yet, so I can't say. But I think we should switch anyways since the default rng should be better from a statistical point of view too (as I understand it).

isaacsas:
I always had concerns about the *128 family as some members failed some of the standard test suites (I never found anything though about the one we were using, hence why I left it).

isaacsas:
A very simple test:
```julia
using Random, RandomNumbers, BenchmarkTools

N = 1000000

function bdpath(N,rng)
    x = 0.0
    for i = 1:N
        x += rand(rng)
    end
end

rng1 = Random.default_rng()
rng2 = Xorshifts.Xoroshiro128Star(rand(UInt64))

b1 = @benchmark bdpath($N,$rng1)
b2 = @benchmark bdpath($N,$rng2)

display("Julia:")
@show median(b1)

display("Xorshift:")
@show median(b2)
```
gives
```
"Julia:"
median(b1) = TrialEstimate(997.375 μs)

"Xorshift:"
median(b2) = TrialEstimate(1.571 ms)
```
on my M1 MBA with 1.7.1.

Using `randn` instead of `rand` gives
```julia
"Julia:"
median(b1) = TrialEstimate(3.684 ms)
"Xorshift:"
median(b2) = TrialEstimate(3.841 ms)
```
so less of an improvement there.

ChrisRackauckas:
Still pretty decent. Add in the thread safety and 👍 this is great.

isaacsas:
Yeah, that was great improvement in base.

Issue #221, opened 2022-02-21
Strange Crash for SortingDirect under obscure circumstances


TorkelE:
There is a weird crash I have encountered when using `SortingDirect`. The circumstances are rather obscure, but unique for this solver, and doesn't really make sense.

This code causes a crash when trying to solve the problem:
```julia
# Activate local environment.
import Pkg
Pkg.activate(".")

using DiffEqJump
using ModelingToolkit
using ReactionNetworkImporters

model = loadrxnetwork(BNGNetwork(), "multisite2_no_obs.txt");
dprob = DiscreteProblem(model.rn,model.u₀,(0.0,0.0),model.p); dprob = remake(dprob,u0=Int64.(dprob.u0));
jprob = JumpProblem(model.rn,dprob,SortingDirect(),save_positions=(false,false));
solve(remake(jprob,tspan=(0.0,10.0)),SSAStepper());
```
basically, I load a model, create a jump problem (with `tsap=(0.0,0.0)`). Then I use `remake` to change the `tspan` to `(0.0,10.0)` and solve it using `SortingDirect`. This crashes julia on jupyter (also tried when running a script, see the end). The weird thing is that if I change the solver to any other (say `RSSA`), or sets a different initial timespan (`(0.0,10.0)`), or both, things are fine:
```julia
# Activate local environment.
import Pkg
Pkg.activate(".")

using DiffEqJump
using ModelingToolkit
using ReactionNetworkImporters

model = loadrxnetwork(BNGNetwork(), "multisite2_no_obs.txt"");
dprob = DiscreteProblem(model.rn,model.u₀,(0.0,10.0),model.p); dprob = remake(dprob,u0=Int64.(dprob.u0));
jprob = JumpProblem(model.rn,dprob,SortingDirect(),save_positions=(false,false));
solve(remake(jprob,tspan=(0.0,10.0)),SSAStepper());
```
```julia
# Activate local environment.
import Pkg
Pkg.activate(".")

using DiffEqJump
using ModelingToolkit
using ReactionNetworkImporters

model = loadrxnetwork(BNGNetwork(), "multisite2_no_obs.txt"");
dprob = DiscreteProblem(model.rn,model.u₀,(0.0,0.0),model.p); dprob = remake(dprob,u0=Int64.(dprob.u0));
jprob = JumpProblem(model.rn,dprob,RSSA(),save_positions=(false,false));
solve(remake(jprob,tspan=(0.0,10.0)),SSAStepper());
```
[...]

PR #222, opened 2022-02-21
fix bug in not registering new tspan[2] values

isaacsas:
Closes https://github.com/SciML/DiffEqJump.jl/issues/221


isaacsas:
Luckily none of the SSAs actually really use the `end_time` in a core way except `SortingDirect` and `DirectCR` I think, (i.e. where we stop calculating the next jump if the next event time is after it). They all instead rely on the time stepper to stop at the right time, hence why the crash in https://github.com/SciML/DiffEqJump.jl/issues/221 wasn't occurring for RSSA.

isaacsas:
@ChrisRackauckas the test crash seems to be in StochasticDiffEq?

PR #225, opened 2022-05-02
Attempt to fix complex states with VariableRateJumps

isaacsas:
@ChrisRackauckas this was an attempt to fix

https://discourse.julialang.org/t/differentialequations-jl-package-variable-rate-jumps-with-complex-variables/80366/2

but tests are now failing (locally) in DiffEqBase with respect to callback conditions now being complex. 

Also, this has pointed out that there is no way in some places to get at the user-specified `rng` in `JumpProblem`, which means that we may have a situation where the rng for constant rate jumps is different than for variable rate jumps...

### DIFF
--- a/src/problem.jl
+++ b/src/problem.jl
@@ -191,10 +191,9 @@ function extend_problem(prob::DiffEqBase.AbstractODEProblem,jumps)
   function jump_f(du::ExtendedJumpArray,u::ExtendedJumpArray,p,t)
     prob.f(du.u,u.u,p,t)
     update_jumps!(du,u,p,t,length(u.u),jumps.variable_jumps...)
-  end
-  T = realtype(eltype(prob.u0))
+  end  
   ttype = eltype(prob.tspan)
-  u0 = ExtendedJumpArray(prob.u0,[T(-randexp(ttype)) for i in 1:length(jumps.variable_jumps)])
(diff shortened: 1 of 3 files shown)


isaacsas:
OK, with this change tests seem to pass locally, but I'm not sure if this is a bad approach with regards to expectations in DiffEqBase or such.

isaacsas:
I updated to save the `rng` within the `JumpProblem` and use it consistently for `VariableRateJump`s in DiffEqJump.

isaacsas:
Closes https://github.com/SciML/DiffEqJump.jl/issues/226

ChrisRackauckas:
This looks fine. It looks like JumpProblem's remake needs to be updated for the extended jump problem definition as well.

rveltz:
what do you mean by extended problem?

ChrisRackauckas:
the pseudo values used for the rootfinding to get events

isaacsas:
> what do you mean by extended problem?

@rveltz for systems using `VariableRateJump`s the solution vector `u` is augmented into a struct that also stores the sampled exponential random variates for finding event times (as Chris mentioned). See https://github.com/SciML/DiffEqJump.jl/blob/master/src/extended_jump_array.jl

Issue #239, opened 2022-06-18
SSAIntegrator not playing nice with PresetTimeCallback -- ERROR: type SSAIntegrator has no field tdir


jae0:
Hi there. An expected field seems to be missing (tdir):  

`ERROR: type SSAIntegrator has no field tdir
Stacktrace:
  [1] getproperty
    @ ./Base.jl:42 [inlined]
  [2] (::DiffEqCallbacks.var"#62#65"{typeof(SciMLBase.INITIALIZE_DEFAULT), Bool, Vector{Float64}, typeof(affectpresets!)})(c::DiscreteCallback{DiffEqCallbacks.var"#60#63"{Vector{Float64}}, DiffEqCallbacks.var"#61#64"{typeof(affectpresets!)}, DiffEqCallbacks.var"#62#65"{typeof(SciMLBase.INITIALIZE_DEFAULT), Bool, Vector{Float64}, typeof(affectpresets!)},  ...
`

Minimal example to create the error follows:
`
using Catalyst, DifferentialEquations 

 rs = @reaction_network begin
   g, X --> 2X
   l, X --> 0
 end g l  

 dprob = DiscreteProblem(rs, [1000.0], (0.0, 100.0), ( 1.0, 1.0 ) )
 jprob = JumpProblem(rs, dprob, Direct(), save_positions=(false, false))
 cbtimes = [20.0, 30.0]
 affectpresets!(integrator) = integrator.u[1] += 1000
  
 jsol = solve(jprob, SSAStepper(), saveat=0.1, tstops=cbtimes, callback=PresetTimeCallback( cbtimes, affectpresets! )  )

`

isaacsas:
Added a fix in https://github.com/SciML/DiffEqJump.jl/pull/240

A few notes on your code:
1. Don't use ` PresetTimeCallback` and manually set `tstops` in the `solve` call, or you'll end up with the `tstops` included twice. (i.e. the `PresetTimeCallback` already sets `tstops` for you.)
2. I'd recommend making your initial condition an integer vector instead of `Float64` unless you are trying to simulate a hybrid system that includes ODEs or SDEs for some components.

jae0:
Great to hear. Thanks! And hats off to you. Jae

PR #240, opened 2022-06-19
Add tdir

isaacsas:
Closes https://github.com/SciML/DiffEqJump.jl/issues/239

@ChrisRackauckas how should the DiffEqJump version be bumped for a new release since the parametric type signature for `SSAStepper` has changed? (Though I'd guess no one is actually dispatching on its parametric form...)


ChrisRackauckas:
Adding parametric types isn't breaking because you can still dispatch on the less in the case of the more. But yeah, that's not public API so this is fine.

ChrisRackauckas (approved):

isaacsas:
OK, then I'll just make a patch release.

ChrisRackauckas:
Okay cool, https://github.com/SciML/DiffEqBase.jl/pull/778 should handle that.

PR #244, opened 2022-06-23
format the repo

ChrisRackauckas:


src/aggregators/direct.jl
```diff
@@ -1,135 +1,199 @@
...
 # creating the JumpAggregation structure (tuple-based constant jumps)
-function aggregate(aggregator::Direct, u, p, t, end_time, constant_jumps,
-                    ma_jumps, save_positions, rng; kwargs...)
-
-  # handle constant jumps using tuples
-  rates, affects! = get_jump_info_tuples(constant_jumps)
-
-  build_jump_aggregation(DirectJumpAggregation, u, p, t, end_time, ma_jumps,
-                          rates, affects!, save_positions, rng; kwargs...)
+function aggregate(
```
isaacsas, line 33: I am very much not in favor of this type of reformatting. This is using an absolutely enormous amount of vertical space, and making it much harder to actually read the (very simple two-line) function (which now barely fits on my laptop screen). Is there not a rule that can wrap multiple arguments? The closing parenthesis also has a weird placement that gives the impression the function has ended...

isaacsas: Are you using different settings than ModelingToolkit uses? It doesn't seem to reformat in this way... See https://github.com/SciML/ModelingToolkit.jl/blob/9cb6a4d139162ac59d7bda77d1e836b7500f4d61/src/systems/jumps/jumpsystem.jl#L111

PR #252, opened 2022-07-03
Ficticious jump algorithm for time-dependent variable rate jumps.

pihop:
Extends the VariableRateJump interface to allow for user-defined bounds on time-dependent jump rates and look-ahead horizons for which the bound is valid. Implements a fictitious jump rejection algorithm based on [https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1004923](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1004923).

Relevant issue: [https://github.com/SciML/JumpProcesses.jl/issues/28](https://github.com/SciML/JumpProcesses.jl/issues/28)

### DIFF
--- a/src/aggregators/aggregators.jl
+++ b/src/aggregators/aggregators.jl
@@ -144,8 +144,18 @@ doi: 10.1063/1.4928635
 """
 struct DirectCRDirect <: AbstractAggregatorAlgorithm end
 
+"""
+The Extrande method for simulating variable rate jumps with user-defined bounds
+on jumps rates and validity intervals via rejection.
+
+Stochastic Simulation of Biomolecular Networks in Dynamic Environments, Voliotis
+M, Thomas P, Grima R, Bowsher CG, PLOS Computational Biology 12(6): e1004923.
+(2016); doi.org/10.1371/journal.pcbi.1004923
+"""
+struct Extrande <: AbstractAggregatorAlgorithm end
+
 const JUMP_AGGREGATORS = (Direct(), DirectFW(), DirectCR(), SortingDirect(), RSSA(), FRM(),
-                          FRMFW(), NRM(), RSSACR(), RDirect())
+                          FRMFW(), NRM(), RSSACR(), RDirect(), Extrande())
 
 # For JumpProblem construction without an aggregator
 struct NullAggregator <: AbstractAggregatorAlgorithm end
@@ -167,3 +177,6 @@ needs_vartojumps_map(aggregator::RSSACR) = true
 is_spatial(aggregator::AbstractAggregatorAlgorithm) = false
 is_spatial(aggregator::NSM) = true
 is_spatial(aggregator::DirectCRDirect) = true
+
+is_ficticious(aggregator::AbstractAggregatorAlgorithm) = false
+is_ficticious(aggregator::Extrande) = true
--- a/src/aggregators/extrande.jl
+++ b/src/aggregators/extrande.jl
@@ -1,0 +1,133 @@
+# Define the aggregator. 
+struct Extrande <: AbstractAggregatorAlgorithm end
+
+"""
+Extrande sampling method for jumps with defined rate bounds.
+"""
+
+nullaffect!(integrator) = nothing
+const NullAffectJump = ConstantRateJump((u,p,t) -> 0.0, nullaffect!)
+
+mutable struct ExtrandeJumpAggregation{T,S,F1,F2,F3,F4,RNG} <: AbstractSSAJumpAggregator
+    next_jump::Int
+    prev_jump::Int
+    next_jump_time::T
+    end_time::T
+    cur_rates::Vector{T}
+    sum_rate::T
+    ma_jumps::S
+    rate_bnds::F3
+    wds::F4
+    rates::F1
+    affects!::F2
+    save_positions::Tuple{Bool,Bool}
+    rng::RNG
+end
+
+function ExtrandeJumpAggregation(nj::Int, njt::T, et::T, crs::Vector{T}, sr::T, maj::S, rs::F1, affs!::F2, sps::Tuple{Bool,Bool}, rng::RNG; rate_bounds::F3, windows::F4, kwargs...) where {T,S,F1,F2,F3,F4,RNG}
+
+    ExtrandeJumpAggregation{T,S,F1,F2,F3,F4,RNG}(nj, nj, njt, et, crs, sr, maj, rate_bounds, windows, rs, affs!, sps, rng)
+end
+
+
+############################# Required Functions ##############################
+function aggregate(aggregator::Extrande, u, p, t, end_time, constant_jumps,
+                   ma_jumps, save_positions, rng; bounded_va_jumps, kwargs...)
+
+    rates, affects! = get_jump_info_fwrappers(u, p, t, (constant_jumps..., bounded_va_jumps..., NullAffectJump))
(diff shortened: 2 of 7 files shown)


isaacsas:
@pihop still taking a look over this -- I've been delayed by the US holiday and travel for it. One quick comment is it would be great if you could add a test that actually tests accuracy by comparing some statistic against simulations using the current approach.

src/aggregators/aggregators.jl
```diff
@@ -181,7 +191,11 @@ needs_vartojumps_map(aggregator::RSSACR) = true
 # true if aggregator supports variable rates
 supports_variablerates(aggregator::AbstractAggregatorAlgorithm) = false
 supports_variablerates(aggregator::Coevolve) = true
+supports_variablerates(aggregator::Extrande) = true
 
 is_spatial(aggregator::AbstractAggregatorAlgorithm) = false
 is_spatial(aggregator::NSM) = true
 is_spatial(aggregator::DirectCRDirect) = true
+
+is_ficticious(aggregator::AbstractAggregatorAlgorithm) = false
```
isaacsas, line 200: `is_ficticious` doesn't seem to ever be used so can be dropped?

src/aggregators/extrande.jl
```diff
@@ -0,0 +1,140 @@
...
+    majumps = p.ma_jumps
+    @inbounds for i in 1:get_num_majumps(majumps)
+        p.cur_rates[i] = evalrxrate(u, i, majumps)
+        dt = randexp(p.rng) / p.cur_rates[i]
+        if dt < ttnj
+            ttnj = dt
+            nextrx = i
+        end
+    end
+    nextrx, ttnj
```
isaacsas, line 76: I'm a little confused, as the extrande method does not generate `ttnj` for individual reactions, but instead calculates the total propensity like the `Direct` method. What you are doing here seems more like Gillespie's first reaction method.

src/aggregators/extrande.jl
```diff
@@ -0,0 +1,140 @@
...
+end
+
+@fastmath function next_extrande_jump(p::ExtrandeJumpAggregation, u, params, t)
+    ttnj = typemax(typeof(t))
+    nextrx = zero(Int)
+    Wmin = typemax(typeof(t))
+    Bmax = typemax(typeof(t))
+
+    # Calculate the total rate bound and the largest common validity window.
+    Ws = zeros(typeof(t), length(p.wds))
```
isaacsas, line 86: You don't want to allocate a vector every step of the algorithm. Why not just calculate the minimum `Ws` and `Bs` inside the for loop instead of saving these values?

src/aggregators/extrande.jl
```diff
@@ -0,0 +1,140 @@
...
+            cur_rates = p.cur_rates
+            @inbounds for i in idx:length(cur_rates)
+                cur_rates[i] = cur_rates[i] + prev_rate
+                prev_rate = cur_rates[i]
+            end
+
+            UBmax = rand(p.rng) * Bmax
+            ttnj = prop_ttnj
+            if p.cur_rates[end] ≥ UBmax
+                @inbounds nextrx = findfirst(x -> x ≥ UBmax, p.cur_rates)
```
isaacsas, line 116: Probably shouldn't create an anonymous function at each step too. Don't we have a helper function to do this kind of search already written?

src/aggregators/extrande.jl
```diff
@@ -0,0 +1,140 @@
...
+        else
+            ttnj = Wmin
+        end
+    end
+
+    return nextrx, ttnj
+end
+
+function generate_jumps!(p::ExtrandeJumpAggregation, integrator, u, params, t)
+    nextmaj, ttnmaj = next_ma_jump(p, u, params, t)
```
isaacsas, line 127: Again, this doesn't seem like the extrande method, which I would think would calculate the total propensity and its bound, and then use those to sample which reaction occurs and use the bound to sample the reaction time.

isaacsas (changes requested):
It would be good to add this to the Hawkes test too and make sure that passes, but the bigger thing to address is that this doesn't seem to actually implement Extrande to me (based on the pseudo-code in Box 1 of the PLOS Comp Bio paper). 

If you are trying to implement a different method that is fine, but we shouldn't call this Extrande then, and should reference the relevant paper.

src/aggregators/aggregators.jl
pihop: Yes, `is_ficticious` can be dropped.

src/aggregators/extrande.jl
pihop: Yes, the Gillespie's stuff shouldn't have stayed there.

src/aggregators/extrande.jl
pihop: Something like this ok?
```  
if !isempty(p.rate_bnds)
    Bmax = typeof(t)(0.)
    @inbounds for i in 1:length(p.wds)
        Wmin = min(Wmin, p.wds[i](u, params, t))
        Bmax += p.rate_bnds[i](u, params, t)
    end
end
```

src/aggregators/extrande.jl
pihop: Couldn't find a relevant helper function but I can just write a while loop iterating over the reaction indices instead (like RDirect does).
```
@inbounds while p.cur_rates[nextrx] < UBmax
    nextrx += 1
end
```

src/aggregators/extrande.jl
pihop: Yes, that shouldn't have stayed in. I'll submit the fixes with Extrande added to the Hawkes test.

src/aggregators/extrande.jl
```diff
@@ -0,0 +1,117 @@
...
+
+@fastmath function next_extrande_jump(p::ExtrandeJumpAggregation, u, params, t)
+    ttnj = typemax(typeof(t))
+    nextrx = zero(Int)
+    Wmin = typemax(typeof(t))
+    Bmax = typemax(typeof(t))
+
+    # Calculate the total rate bound and the largest common validity window.
+    if !isempty(p.rate_bnds)
+        Bmax = typeof(t)(0.)
```
isaacsas:
```suggestion
        Bmax = zero(t)
```

src/aggregators/extrande.jl
```diff
@@ -0,0 +1,117 @@
...
+                                                       rate_bounds, windows, rs, affs!, sps,
+                                                       rng)
+end
+
+############################# Required Functions ##############################
+function aggregate(aggregator::Extrande, u, p, t, end_time, constant_jumps,
+                   ma_jumps, save_positions, rng; variable_jumps = (), kwargs...)
+    ma_jumps_ = !isnothing(ma_jumps) ? ma_jumps : ()
+    rates, affects! = get_jump_info_fwrappers(u, p, t,
+                                              (constant_jumps..., variable_jumps..., ma_jumps_...,
```
isaacsas: This doesn't make sense as `MassActionJump`s don't use the same interface as `ConstantRateJump`s or `VariableRateJump`s. So while you can merge the rate functions from the latter two you need to keep track of `MassActionJump`s separately and handle their rate calculation separately. See what we do in `Direct` for handling them.

isaacsas (changes requested):
You need to update `next_extrande_jump` to handle `MassActionJump`s too, see how we do it in `Direct.jl`. They work differently than `ConstantRateJump`s or `VariableRateJump`s.

Please also add `extrande` to some of the tests for constant rate problems so we can make sure it is working ok there too.

isaacsas:
Looks like a real CI failure.

PR #259, opened 2022-08-20
Change Int64 to Int8 in nums_neighbors

Vilin97:


isaacsas:
Does this have any appreciable performance impact?

Vilin97:
Not directly relevant, but using `Graphs`  requires about 10x more memory over the current `CartesianGrid`, and 100x more memory over the `CartesianGrid` that uses Int8.
image

isaacsas (approved):
Ok, just wanted to check performance didn’t decrease — I had a recollection that in Julia `Float16` can give lower performance than `Float64`  — but I didn’t think that would be an issue for `Int8`. Thanks for checking it!

PR #263, opened 2022-08-27
redundant coveralls

ArnoStrouwen:
Don't know if this is needed for some special reason, but most of the packages did away with coveralls?


isaacsas:
I find it nicer for reading the diffs so want to keep it.

PR #264, opened 2022-08-30
Remove CartesianGridIter.

Vilin97:
`CartesianGridRej` performs significantly better than `CartesianGridIter`.


Vilin97:
`CartesianGridRej`  is almost 2x faster than `CartesianGridIter`.
image

isaacsas (approved):

isaacsas:
LGTM

PR #266, opened 2022-09-06
drop fastmaths

isaacsas:
I keep thinking we should drop these and then forgetting, so I'm putting it through now before I forget again...

### DIFF
--- a/src/aggregators/direct.jl
+++ b/src/aggregators/direct.jl
@@ -58,7 +58,7 @@ end
 # calculate the next jump / jump time
 function generate_jumps!(p::DirectJumpAggregation, integrator, u, params, t)
     p.sum_rate, ttnj = time_to_next_jump(p, u, params, t)
-    @fastmath p.next_jump_time = t + ttnj
+    p.next_jump_time = t + ttnj
     @inbounds p.next_jump = searchsortedfirst(p.cur_rates, rand(p.rng) * p.sum_rate)
     nothing
 end
@@ -66,8 +66,8 @@ end
 ######################## SSA specific helper routines ########################
 
 # tuple-based constant jumps
-@fastmath function time_to_next_jump(p::DirectJumpAggregation{T, S, F1, F2, RNG}, u, params,
-                                     t) where {T, S, F1 <: Tuple, F2 <: Tuple, RNG}
+function time_to_next_jump(p::DirectJumpAggregation{T, S, F1, F2, RNG}, u, params,
+                           t) where {T, S, F1 <: Tuple, F2 <: Tuple, RNG}
     prev_rate = zero(t)
     new_rate = zero(t)
     cur_rates = p.cur_rates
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
What was the effect of these ones? Some should just be replaced with `@muladd`?

src/aggregators/direct.jl
```diff
@@ -58,16 +58,16 @@ end
 # calculate the next jump / jump time
 function generate_jumps!(p::DirectJumpAggregation, integrator, u, params, t)
     p.sum_rate, ttnj = time_to_next_jump(p, u, params, t)
-    @fastmath p.next_jump_time = t + ttnj
```
isaacsas, line 61: @ChrisRackauckas I'll go through these for you. This is just an addition, so I don't see why we'd benefit from @fastmath (and `ttnj` could be `Inf` so I'm not sure it is reasonable to keep `fastmath`).

src/aggregators/direct.jl
```diff
@@ -58,16 +58,16 @@ end
...
+    p.next_jump_time = t + ttnj
     @inbounds p.next_jump = searchsortedfirst(p.cur_rates, rand(p.rng) * p.sum_rate)
     nothing
 end
 
 ######################## SSA specific helper routines ########################
 
 # tuple-based constant jumps
-@fastmath function time_to_next_jump(p::DirectJumpAggregation{T, S, F1, F2, RNG}, u, params,
```
isaacsas, line 69: In this function we do a cumulative sum and one division at the end (unless the `@fastmath` propagates into the rate and `randexp` functions, in which case it would apply within them too).

src/massaction_rates.jl
```diff
@@ -3,8 +3,8 @@
 # stochiometric coefficient.
 ###############################################################################
 
-@inline @fastmath function evalrxrate(speciesvec::AbstractVector{T}, rxidx::S,
```
isaacsas, line 6: Here all the calculation should be with integers except the final multiplication.

src/massaction_rates.jl
```diff
@@ -19,17 +19,17 @@
     @inbounds return val * majump.scaled_rates[rxidx]
 end
 
-@inline @fastmath function executerx!(speciesvec::AbstractVector{T}, rxidx::S,
```
isaacsas, line 22: This should again be integer calculations.

src/aggregators/direct.jl
ChrisRackauckas: Using `add_fast` here I think is fine though, since it wouldn't decrease the accuracy and NaN or Inf would both trigger the same exit behavior in the integrator

src/aggregators/direct.jl
ChrisRackauckas: `@fastmath` doesn't propagate: it's a high level find-replace https://github.com/JuliaLang/julia/blob/master/base/fastmath.jl#L34-L85 so it's not like the C fastmath in that sense. It would need to use something like Cassette to do that anyways.

ChrisRackauckas: I guess it can set compile flags too? I haven't seen that (https://github.com/JuliaLang/julia/blob/master/base/fastmath.jl#L141-L143), but we can also just directly use add_fast and such.

isaacsas: Sure, I can change the relevant floating point additions in these functions to use `add_fast`. We can get `Inf`s for the rates (and hence the cumulative sum), so as long as that propagates correctly we should be fine (and it seemed to propagate correctly before even with `@fastmath`).

isaacsas:
@ChrisRackauckas has something changed with how the `kwarg`s in a problem get passed through to `solve` recently? The test error seems to arise from the callback not getting propagated anymore to the solve call.

isaacsas:
@ChrisRackauckas does this look ok to you now with `add_fast` being used in those previously `@fastmath` functions?

ChrisRackauckas:
yes

PR #276, opened 2022-12-07
`Coevolve` aggregator for `VariableRateJump` 

gzagatti:
This PR introduces the `Coevolve` aggregator for `VariableRateJump`. This aggregator supports the `SSAStepper()` integrator for the efficient simulation of jump processes with time-varying rates.

Some background, in `JumpProcesses.jl` most of the algorithms are borrowed from the biochemistry literature. However, jump processes have also been extensively studied in the statistics literature where they go by point processes (see Daley and Vere-Jones [1]). In the point process literature, we find many of the simulation algorithms implemented in this library that go by other names. Among the most popular one is Ogata's algorithm [2] which is very similar to Gillespie's direct method. Ogata's algorithm as adapted in Daley [1] is a general-purpose algorithm for any point process that evolves through time. Farajtabar et al. [3] modifies Ogata's algorithm to use a priority queue in order to simulate a compound Hawkes process using a priority queue. Here I adapt Farajtabar's COEVOLVE algorithm to simulate any evolutionary point process, thus the name of the algorithm.

This algorithm requires user-defined bounds and a dependency graph. Therefore, I extend the `VariableRateJump` to accept three aditional keyword arguments `lrate`, `urate` and `L`. `lrate(u, p, t)` is a function that computes the lower-bound of the rate from `t` to `t + L`, `urate` computes the upper-bound, and `L` computes the interval for which the bounds are valid.

A toy example is given below
```julia
using JumpProcesses, Plots
L(u,p,t) = (1/p[1])*2
rate(u,p,t) = t*p[1]*u[1]
lrate(u, p, t) = rate(u, p, t)
urate(u,p,t) = rate(u, p, t + L(u,p,t))
affect!(integrator) = integrator.u[1] -= 1
vrj = VariableRateJump(rate, affect!; lrate=lrate, urate=urate, L=L)
u0 = [1_000.]
p = (1.0,)
tspan = (0., 10.)
prob = DiscreteProblem(u0, tspan, p)
jprob = JumpProblem(prob, Coevolve(), vrj; dep_graph=[[1]])
sol = solve(jprob, SSAStepper())
plot(sol, title="Trajectory", legend=false)
```

image

I have performed some benchmarks to evaluate the performance of the model. I ran the jump benchmarks of [`SciMLBenchmarks`](https://github.com/gzagatti/SciMLBenchmarks.jl/tree/jumps-queue-method/benchmarks/Jumps). `Coevolve` performs on par with the alternatives, with a fixed penalty over `NRM`. The `QueueMethod` basically reduces to the `NRM` when all the jumps are `ConstantRateJump` or `MassActionJump`, but I was not able to equate the `NRM` performance.

| Benchmark                  | Direct | FRM    | SortingDirect | NRM    | DirectCR | RSSA   | RSSACR | Coevolve |
|----------------------------|--------|--------|---------------|--------|----------|--------|--------|-------------|
| Diffusion CTRW             | 5.1 s  |        | 1.2 s         | 0.7 s  | 0.4 s    | 1.6 s  | 0.4 s  | 1 s         |
| Multistate Model           | 0.1 s  | 0.2 s  | 0.1 s         | 0.2 s  | 0.2 s    | 0.1 s  | 0.1 s  | 0.4 s       |
| Neg. Feed. Gene Expression | 0.2 ms | 0.3 ms | 0.2 ms        | 0.4 ms | 0.4 ms   | 0.4 ms | 0.8 ms | 0.4 ms      |
| Marchetti Gene Expression  | 0.4 s  | 0.6 s  | 0.4s          | 0.9 s  | 0.0.9 s  | 0.6 s  | 0.9 s  | 1.3 s       |

More complex examples include the compound Hawkes process which was the main motivation for contributing this PR.

image

I have also developed a new benchmark using the compound Hawkes process and found significant improvements compared to using the ODE solver. Please check my branch of [`SciMLBenchmarks`](https://github.com/gzagatti/SciMLBenchmarks.jl/blob/jumps-queue-method/benchmarks/Jumps/Hawkes_process.jmd) for more details. Note that the `Direct` method does not run within the allowed time when the number of nodes reaches above 40.

| V  | Direct (n) | Direct (median time) | Coevolve (n) | QueueMethod (median time) |
|----|------------|----------------------|-----------------|---------------------------|
| 1  | 50         | 124.5 μs             | 50              | 3.4 μs                    |
| 10 | 50         | 28.6  ms             | 50              | 455.5 μs                  |
| 20 | 50         | 223.8 ms             | 50              | 5.2 ms                    |
| 30 | 50         | 715.1 ms             | 50              | 7.9 ms                    |
| 40 | 31         | 715.1 ms             | 50              | 30.2 ms                   |
| 50 | 15         | 4.1 s                | 50              | 61.7 ms                   |
| 60 | 10         | 6.7 s                | 50              | 137.1 ms                  |
| 70 | 6          | 10.9 s               | 50              | 202.9 ms                  |
| 80 | 4          | 17.8 s               | 50              | 346.4 ms                  |
| 90 | 3          | 27.24 s              | 50              | 642.0 ms                  |

I have added unit tests and updated the documentation to reflect the changes introduced here.

As I completed this PR, I learnt about PR #252 which also extends `VariableRateJump`. The author implements a different algorithm which also uses rate bounds. While the algorithm implemented in that PR is analogous to rejection-based algorithms, the algorithm in this PR is analogous to next-reaction methods. So they both face different trade-offs.

I am looking forward to your feedback.

[1] D. J. Daley and D. Vere-Jones, An Introduction to the Theory of Point Processes: Volume I: Elementary Theory and Methods, 2nd ed. New York: Springer-Verlag, 2003. doi: 10.1007/b97277.

[2] Y. Ogata, “On Lewis’ simulation method for point processes,” IEEE Transactions on Information Theory, vol. 27, no. 1, pp. 23–31, Jan. 1981, doi: 10.1109/TIT.1981.1056305.

[3] M. Farajtabar, Y. Wang, M. Gomez-Rodriguez, S. Li, H. Zha, and L. Song, “COEVOLVE: a joint point process model for information diffusion and network evolution,” J. Mach. Learn. Res., vol. 18, no. 1, pp. 1305–1353, Jan. 2017, doi: 10.5555/3122009.3122050.


isaacsas:
This looks great! Thanks also for so comprehensively commenting and adding docs/tests.

I’ll try to give this a comprehensive review later today.

gzagatti:
I renamed `QueueMethod` to `Coevolve` to better reflect the source of the algorithm. Also, I felt `QueueMethod` was too generic and the `Method` suffix a bit redundant.

isaacsas:
Sorry, grading took longer than expected today, so probably won’t get to looking this over till tomorrow afternoon (last day of classes here…).

gzagatti:
Don't worry. Grading is always time consuming. Thanks for the update.

src/jumps.jl
```diff
@@ -393,19 +437,27 @@ function JumpSet(vj, cj, rj, maj::MassActionJump{S, T, U, V}) where {S <: Number
     JumpSet(vj, cj, rj, check_majump_type(maj))
 end
 
-JumpSet(jump::ConstantRateJump) = JumpSet((), (jump,), nothing, nothing)
-JumpSet(jump::VariableRateJump) = JumpSet((jump,), (), nothing, nothing)
-JumpSet(jump::RegularJump) = JumpSet((), (), jump, nothing)
-JumpSet(jump::AbstractMassActionJump) = JumpSet((), (), nothing, jump)
-function JumpSet(; variable_jumps = (), constant_jumps = (),
+function JumpSet(jump::VariableRateJump)
```
isaacsas, line 440: Why all the changes to `JumpSet` construction?

src/problem.jl
```diff
@@ -185,39 +185,72 @@ function JumpProblem(prob, aggregator::AbstractAggregatorAlgorithm, jumps::JumpS
...
-    if typeof(jumps.variable_jumps) <: Tuple{}
+    if length(jumps.variable_jumps) == 0 && (length(jumps.constant_jumps) == 0) &&
+       (maj === nothing) && !is_spatial(aggregator)
+        # check if there are no jumps
+        new_prob = prob
+        variable_jump_callback = CallbackSet()
+        cont_agg = JumpSet().variable_jumps
+        disc_agg = nothing
+        constant_jump_callback = CallbackSet()
+    elseif typeof(aggregator) <: Coevolve
```
isaacsas, line 200: Can we add a trait `supports_variablerates` which defaults to `false` but `Coevolve` overloads to `true`? Since we'd ultimately like to support other vrj supporting algorithms.

src/jumps.jl
```diff
@@ -50,28 +50,57 @@ jump process is
...
+    """When planning to use the `Coevolve` aggregator, function `urate(u, p,
+    t)` that computes the upper bound of the rate in interval `t` to `t + L` at time
+    `t` given state `u`, parameters `p`. This is not required if using another
+    aggregator."""
+    urate::R3
+    """When planning to use the `Coevolve` aggregator, function `L(u, p,
+    t)` that computes the interval  length `L` starting at time `t` given state
+    `u`, parameters `p` for which the rate is bounded between `lrate` and
+    `urate`. This is not required if using another aggregator."""
+    L::R4
```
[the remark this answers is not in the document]
isaacsas: Also, we should think a bit about interface. If `L` is going to require some subset of `urate`, `lrate` and/or the actual rate, we might want to have a way to (optionally) pass them to `L` if the algorithm has valid values for them (avoiding the need to call them extra times). For example, we could require `L` to support passing them in as optional keywords (with default values of `nothing`). In this way the algorithm can just pass in those that it knows, and leave any others the user may need for their custom `L` function to calculate.

Alternatively, would it make sense to merge all three functions into one that returns the upper/lower bounds and the time window?

src/aggregators/coevolve.jl
```diff
@@ -0,0 +1,206 @@
...
+            t = t + _L
+            continue
+        end
+        if _t >= tstop
+            break
+        end
+        _lrate = lrate(u, params, t)
+        if _lrate > _urate
+            error("The lower bound should be lower than the upper bound rate for t = $(t) and i = $(i), but lower bound = $(_lrate) > upper bound = $(_urate)")
+        elseif _lrate < _urate
```
isaacsas, line 179: So `lrate` is not strictly required for the algorithm; it is just helpful if for some problems it is easier to calculate than the full `rate` function?

In that case, couldn't we make it optional, and here if it's field is `nothing` just sample with the actual rate, but if it is present sample with both as you currently do? I guess we might want to have a bitvector that stores that information in the aggregator to avoid having to examine the type within the core loop here.

isaacsas (commented):
I took a high level pass. Once we settle the higher-level issues in my comments and below, I'll take a more careful look at the documentation and some of the other changes.

Some general questions:
1. Should `L` take the upper/lower/current rates as optional keywords to avoid having to recalculate them (when they are available already from the algorithm)?
2. Should we consider merging `urate`, `lrate` and `L` into one function that returns all three? I can argue for/against this, so I'd be interested in hearing your thoughts.
3. Why the changes to `JumpSet`s? Are you sure this doesn't break their working with and preserving`Tuple`s for `Direct` simulations?
4. I'd prefer to keep `ConstantRateJump`s distinguished in the algorithms, and more generally just handle them and `MassActionJump`s specially. We can assume an ordering of `MassActionJump`s, `ConstantRateJump`s and `VariableRateJump`s for agrregators that support all three. In that case, couldn't you then just reuse the `NRM` time updates for them? I'd be fine with some refactoring of `NRM` to make that easier to reuse (or even considering merging `NRM` and `Coevolve` if we can make it functionally equivalent for systems without `VariableRateJump`s).
6. Can you update some of the other tests that loop through most of the aggregators and add `Coevolve` to them? It'd be good to check that it works for mixes of `MassActionJump`s and `ConstantRateJump`s too.

src/aggregators/coevolve.jl
```diff
@@ -0,0 +1,206 @@
...
+                                 lrates, urates, Ls) where {T, S, F1, F2, RNG, U}
+    if dep_graph === nothing
+        if (get_num_majumps(maj) == 0) || !isempty(rs)
+            error("To use VariableRateJumps with the Queue Method algorithm a dependency graph between jumps must be supplied.")
+        else
+            dg = make_dependency_graph(length(u), maj)
+        end
+    else
+        # using a Set to ensure that edges are not duplicate
+        dg = [Set{Int}(append!([], jumps, [var]))
```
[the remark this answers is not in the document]
gzagatti: That makes sense.

src/aggregators/coevolve.jl
```diff
@@ -0,0 +1,206 @@
...
+        return _t
+    end
+    return typemax(t)
+end
+
+# reevaulate all rates, recalculate all jump times, and reinit the priority queue
+function fill_rates_and_get_times!(p::CoevolveJumpAggregation, u, params, t)
+    @unpack end_time = p
+    num_jumps = get_num_majumps(p.ma_jumps) + length(p.rates)
+    jump_times = Vector{eltype(t)}(undef, num_jumps)
```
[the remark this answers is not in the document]
gzagatti: You're right it should be `typeof`.

src/jumps.jl
```diff
@@ -81,14 +110,29 @@ struct VariableRateJump{R, F, I, T, T2} <: AbstractJump
 end
 
 function VariableRateJump(rate, affect!;
+                          lrate = nothing, urate = nothing,
+                          L = nothing, rootfind = true,
                           idxs = nothing,
-                          rootfind = true,
-                          save_positions = (true, true),
+                          save_positions = (lrate !== nothing && urate !== nothing &&
```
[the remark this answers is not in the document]
gzagatti: Good catch. This was required on a different iteration of this branch where I tried to carry the history of the process in the algorithm, but then I realized this was not necessary.

src/jumps.jl
gzagatti: The current problem with `JumpSet` is that it cannot handle large systems. 

Currently, the initialization of `JumpSet` uses tuples to specialize the type of `JumpSet`. While this could potentially be more efficient for small systems, it raises a stack overflow error in type inference for larg-ish systems. 

You can try for your self:
```julia
jumps = [ConstantRateJump((u, p, t) -> p[1], (integrator) -> integrator.u[i] += 1) for i in 1:1_000]
jumpset = JumpSet(jumps...)
```
Which will raise
```
Internal error: stack overflow in type inference of split_jumps( ....
```

Modifying `JumpSet` to use an Array instead of a Tuple in the case of `ConstantRateJump` and `VariableRateJump` improves the speed of the initialization and solves the stack overflow error.

My understanding is that `JumpSet` is an intermediary container to facilitate the initialization of `JumpProblem`. After initialization `JumpSet` is basically discarded from `JumpProblem` which then holds all the tuples created by `JumpSet`. However, the aggregators which do the heavy lifting do not even use these tuples as they create their own data structures to handle the jumps. 

So, is it really necessary to create tuples if they are just being used as intermediary containers to facilitate initialization? Would it not be better just to pass them around as arrays? 

The changes in `JumpSet` construction are meant to address this issue.

src/aggregators/coevolve.jl
```diff
@@ -0,0 +1,206 @@
...
+        idx = i - num_majumps
+        rate, lrate, urate, L = p.rates[idx], p.lrates[idx], p.urates[idx], p.Ls[idx]
+    end
+    return rate, lrate, urate, L
+end
+
+function next_time(p::CoevolveJumpAggregation{T}, u, params, t, i, tstop::T) where {T}
+    @unpack rng = p
+    rate, lrate, urate, L = get_rates(p, i, u)
+    while t < tstop
```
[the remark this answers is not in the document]
gzagatti: I would rather handle all jumps with the same algorithm. 

I think that this module is not yet designed to make these type of of optimizations for the user. In ideal world, the user would not need to choose an aggregator and the library would 
decide for itself the best one. 

Since we are not there yet, I would argue that the current API promises the user that the same aggregator will be used for all discrete jumps in the problem. We should then honor this promise. Therefore, we should not surprise the user by silently calling aggregator B when aggregator A was invoked.

I think the penalty for using `Coevolve` instead of `NRM` for a problem involving only `MassActionJump`s and `ConstantRateJump`s is fixed and small enough that it should not make a huge difference for a naive user. As for an advanced user, she/he would know to call `NRM` when dealing with such problems.

With that being said, I think that `Coevolve` should have a performance very close to `NRM` in the cases involving only `MassActionJump`s and/or `ConstantRateJump`s  as the logic in the `next_time` function of `Coevolve` should reduce to the one in `NRM`. 

It seems like the penalty might be imposed from the `get_rates` function, but I'm not sure yet.

src/aggregators/coevolve.jl
```diff
@@ -0,0 +1,206 @@
...
+    @inbounds deps = p.dep_gr[p.next_jump]
+    @unpack end_time, pq = p
+    for (ix, i) in enumerate(deps)
+        ti = next_time(p, u, params, t, i, end_time)
+        update!(pq, i, ti)
+    end
+    nothing
+end
+
+function get_rates(p::CoevolveJumpAggregation, i, u)
```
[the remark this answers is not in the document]
gzagatti: Are you saying that we should move the logic in this function to inside of `next_time` instead of calling it from there?

In `NRM` the `cur_rates` vector is updated inside of the function `update_dependent_rates!` by calling the function `calculate_jump_rate`, reproduced below: 
```julia
@inline function calculate_jump_rate(ma_jumps, num_majumps, rates, u, params, t, rx)
    if rx <= num_majumps
        return evalrxrate(u, rx, ma_jumps)
    else
        @inbounds return rates[rx - num_majumps](u, params, t)
    end
end
```

This function also branches to compute the rates for either `MassActionJump` or `ConstantRateJump`. So, in a sense it does very similar to `get_rates` in this PR.

One of the main differences is that `get_rates` is not inlined. I'm not sure if this could cause a performance penalty.

src/jumps.jl
```diff
@@ -81,14 +110,29 @@ struct VariableRateJump{R, F, I, T, T2} <: AbstractJump
...
+    if !((lrate === nothing && urate === nothing && L === nothing) ||
+         (lrate !== nothing && urate !== nothing && L !== nothing))
+        error("Either `lrate`, `urate` and `L` must be nothing, or all of them must be defined.")
+    end
+
+    VariableRateJump(rate, affect!, lrate, urate, L, idxs, rootfind,
+                     interp_points, save_positions, abstol, reltol)
+end
+
+function VariableRateJump(jump::ConstantRateJump)
```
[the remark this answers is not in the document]
gzagatti: The problem is that the aggregators are not jump "aware", which also came as a surprise to me. When `JumpProblem` is initialized it calls `aggregate` to initialize the aggregator. This function expects an iterable of jumps which does not distinguish between `ConstantRateJump` and `VariableRateJump`. 

I defined this constructor to be used internally, it makes it easier to construct an iterable of `VariableRateJump` to pass to the `aggregate` function of `Coevolve`. As such, I do not mention this in the documentation as an option for creating `VariableRateJump`s.

As you suggested below, we could use the index of the jump to distinguish between `MassActionJump`, `ConstantRateJump` and `VariableRateJump`. Problems arise when the aggregator expects a dependency graph. In this case, the user needs to be aware of this rule so that when he passes `VariableRateJump`s before `ConstantRateJump`s. It must be clear whether graph indices should be declared in the order passed by the user or in the order that the aggregator expects. An option would be to take the dependency graph passed by the user and re-map it internally to the way we sort the indices for the aggregator. However, I feel this might be prone to errors and more difficult to debug for the final user since graph indices in the aggregator will not correspond to the one used to initialize the problem.

Alternatively, the aggregator could keep a vector that tracks the type of jump for each index. We could pass this vector when calling `aggregate`. This might exert a penalty cost, but might be cleaner and leave the door open for other algorithms to distinguish between jump types in any order.

src/jumps.jl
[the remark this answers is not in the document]
gzagatti: I did some experimentation with different APIs including the one you suggested of having a single function that returns the upper/lower bounds and the time window. I eventually settled for this one, as I thought it could be simpler and more modular. Since defining `L` is more art, than science. The user could play with different definitions of `L` without touching `urate` or `lrate`. Also, there might be aggregators that do need  either one of `L`, `urate` or `lrate`.

Previously, I also had problems with defining `FunctionWrapper`s for complex outputs. But then, I also tried incorporating closures and event history which was the main reason why I struggled with `FunctionWrapper`s. Maybe this time I will get luckier.

`Coevolve` will evaluate `urate` and `L` right at the beginning of the main loop. While `lrate`, is optional and only evaluated later. Ideally, to avoid re-evaluating functions, the user could defined memoized functions. The [`Memoization.jl`](https://github.com/marius311/Memoization.jl) library could be an option for this. Internally, we might want to consider wrapping these functions with the `@memoize` macro, but I'm not sure whether this would bring the same efficiency gains as using the keyword argument.

Alternatively, I would propose that we require two functions `urate` and `lrate`. `urate` returns the upper-bound and the time-interval; while `lrate` is optional and takes `urate, and `L` as optional keywords to avoid re-evaluation. The main drawback is the lack of modularity with the potential gain in efficiency if not using memoization.

Let me know what you think.

src/aggregators/coevolve.jl
gzagatti: That's right `lrate` is not strictly required. It just makes the algorithm faster if supplied. The safe default would be to have `lrate = 0` instead.

gzagatti:
Thanks for your suggestions! See answers to your comments below. I will be pushing some changes later today. I am looking forward to hearing your reactions after I make the pushes.

> 1. Should L take the upper/lower/current rates as optional keywords to avoid having to recalculate them (when they are available already from the algorithm)?
> 2. Should we consider merging urate, lrate and L into one function that returns all three? I can argue for/against this, so I'd be interested in hearing your thoughts.

See my answer above on this. 

> 3. Why the changes to JumpSets? Are you sure this doesn't break their working with and preservingTuples for Direct simulations?

See my answer above on this. But, so far I have not faced any issues with using the other aggregators and all tests pass.

What do you mean by `preservingTuples for Direct simulations`?

> 4. I'd prefer to keep ConstantRateJumps distinguished in the algorithms, and more generally just handle them and MassActionJumps specially. We can assume an ordering of MassActionJumps, ConstantRateJumps and VariableRateJumps for agrregators that support all three. In that case, couldn't you then just reuse the NRM time updates for them? I'd be fine with some refactoring of NRM to make that easier to reuse (or even considering merging NRM and Coevolve if we can make it functionally equivalent for systems without VariableRateJumps).

See my answer above on this. In summary, I think we should honour the current API that tells aggregator A will be used to solve the problem, so aggregator B should not be used. 

With that being said, the main loop in `Coevolve` should reduce to the one in `NRM` when using `MassActionJump`s and/or `ConstantRateJump`s. 

The main differences is that `NRM` re-uses the random number to re-compute the new time in [`update_dependent_rates` at line 103](https://github.com/SciML/JumpProcesses.jl/blob/f923fadd97d564fbc046d4be529e592f06f282c3/src/aggregators/nrm.jl#L103):
```julia
update!(p.pq, rx, t + oldrate / cur_rates[rx] * (p.pq[rx] - t))
```

This come at the cost of indexing the mutable binary heap. 

On the other hand, `Coevolve` always generate a new random number when updating the jump. Everything else in `Coevolve` should be the same since the loop short-circuits when evaluating the next time of any jump with an infinity time window `L`. 

The solution then is to use the random number generated previously to re-compute `s` (the time to the next jump), which should get us closer to the performance of `NRM`. This should not be hard to implement.

> 5. Can you update some of the other tests that loop through most of the aggregators and add Coevolve to them? It'd be good to check that it works for mixes of MassActionJumps and ConstantRateJumps too.

I think I have now updated all tests that loop through aggregators to include `Coevolve`. Let me know otherwise.

src/problem.jl
gzagatti: Yes, that's a good idea.

src/aggregators/coevolve.jl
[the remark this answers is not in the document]
isaacsas: I think you are going to have performance issues due to generating anonymous functions inside the hot loop (i.e. in `get_rates`). Generally creating functions is going to have an overhead, and moreover capturing a variable (`_rate`) can also have performance issues in Julia. Since there is no real reason to do this please take it out.

If you don't want to reuse `NRM`s time updates then just modify `next_time` to check if `i < (Number of mass action jumps + constant rate jumps)` and simply explicitly get the new rate via `evalrxrate` and/or the appropriate `rates[idx]`, and sample a new time for them. 

I suspect just making these changes will give performance quite close to `NRM` for systems with only `MassActionJump`s and `ConstantRateJump`s.

src/jumps.jl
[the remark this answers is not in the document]
isaacsas: I think keeping them separate is fine for now, we can add the optional kwargs with cached values later own that might help reduce performance overhead.

src/jumps.jl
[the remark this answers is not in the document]
isaacsas: Can't you just add a kwarg to hold the `VariableRateJump`s in `aggregate`? 

Let's just assume we order `MassActionJump`s, `ConstantRateJump`s and then `VariableRateJump`s in dependency graphs. We already assume the first two have that order now (and say so in the docs -- but perhaps this needs to be made more clear).

Then you can assume `rates` are ordered first by `ConstantRateJump`s and then by `VariableRateJump`s, or just make a separate `vrjrates` field. Likewise you can shorten `urates`, `lrates` and `Ls` to just have length equal to the number of `VariableRateJump`s.

src/jumps.jl
isaacsas: Can't you just use the `JumpSet` kwarg-based constructor? 

https://github.com/SciML/JumpProcesses.jl/blob/f923fadd97d564fbc046d4be529e592f06f282c3/src/jumps.jl#L400

I thought that works with vectors just fine? It may be we need some doc improvements to point users towards it when one has lots of jumps, but I'd prefer not to change the behavior as other packages may rely on it currently.

gzagatti:
I have made all the changes following your feedback. You can now look at the modifications for a second round of comments.

In particular, I have used the same `NRM` trick of re-using the random number when updating the dependent rates which brought some efficiency gains. I am still not beating the `NRM` performance but it's now much closer to it.

| Benchmark                  | Direct | FRM    | SortingDirect | NRM    | DirectCR | RSSA   | RSSACR | Coevolve    |
|----------------------------|--------|--------|---------------|--------|----------|--------|--------|-------------|
| Diffusion CTRW             | 4.9 s  |        | 1.2 s         | 0.7 s  | 0.4 s    | 1.7 s  | 0.4 s  | 0.7 s       |
| Multistate Model           | 0.1 s  | 0.2 s  | 0.1 s         | 0.2 s  | 0.2 s    | 0.1 s  | 0.1 s  | 0.3 s       |
| Neg. Feed. Gene Expression | 0.2 ms | 0.3 ms | 0.2 ms        | 0.3 ms | 0.4 ms   | 0.4 ms | 0.8 ms | 0.4 ms      |
| Marchetti Gene Expression  | 0.5 s  | 0.7 s  | 0.4 s         | 0.8 s  | 0.9 s    | 0.6 s  | 0.8 s  | 1.0 s       |

src/jumps.jl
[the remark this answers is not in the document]
gzagatti: Ok, then I will mark this conversation as resolved.

isaacsas:
The new numbers look much better! However, I would still strongly prefer to drop the function creation and just explicitly handle `MassActionJump`s and `ConstantRateJump`s rate/time updates, bypassing the rejection while loop / logic. As I said above, there is no reason the performance shouldn't be identical I would think.

src/aggregators/coevolve.jl
isaacsas: We can make this a future performance tweak / update.

src/jumps.jl
gzagatti: I reverted these changes to avoid potentially breaking any other packages that rely on it. 

With that being said, I did not experience problems when running benchmarks that use all sort of dependencies from the SciML family of packages.

Thus, I think it might be worth opening an issue to discuss this topic since initializing, `JumpSet` and by extension `JumpProblem` with splatting will again raise stack overflow errors for large networks. Even when not raising errors, it can be unnecessarily slower.

src/aggregators/coevolve.jl
gzagatti: I have already made `lrate` optional. So when `lrate` is not passed to `VariableRateJump` but `urate` and `L` are, `lrate` becomes: 
```julia
lrate = (u, p, t) -> zero(typeof(t))
```

src/aggregators/coevolve.jl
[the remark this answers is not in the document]
gzagatti: Resolving this conversation as it is reflected in other sections of this PR.

src/jumps.jl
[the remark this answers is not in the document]
gzagatti: I have removed the `VariableRateJump` initializer from a `ConstantRateJump`. 

Now, I understand your suggestion. `JumpSet` will split the jumps into the three categories, so their indices will invariably need to be ordered in blocks.

I have bundled `ConstantRateJump`s with `VariableRateJump`s in a vector. Then passing this vector to `aggregate`. I have not created a separate field `vrjrates` as it is not needed.

gzagatti:
Yet, one more iteration. I have completely separated the treatment of each jump type in `Coevolve`. I believe you might prefer this last iteration because of that.

 In terms of performance, we still obtain very similar values from the previous two iterations:

| Benchmark                  | Direct | FRM    | SortingDirect | NRM    | DirectCR | RSSA   | RSSACR | Coevolve    |
|----------------------------|--------|--------|---------------|--------|----------|--------|--------|-------------|
| Diffusion CTRW             | 4.9 s  |        | 1.2 s         | 0.7 s  | 0.4 s    | 1.7 s  | 0.3 s  | 0.7 s       |
| Multistate Model           | 0.1 s  | 0.2 s  | 0.1 s         | 0.2 s  | 0.2 s    | 0.1 s  | 0.1 s  | 0.3 s       |
| Neg. Feed. Gene Expression | 0.2 ms | 0.2 ms | 0.2 ms        | 0.3 ms | 0.4 ms   | 0.4 ms | 0.7 ms | 0.4 ms      |
| Marchetti Gene Expression  | 0.4 s  | 0.6 s  | 0.4 s         | 0.8 s  | 0.8 s    | 0.6 s  | 0.8 s  | 1.0 s       |

I'll post the performance for the Hawkes benchmark as soon as it is completed.

gzagatti:
Here is the performance for the Hawkes benchmark, much improved from the first iteration:

| V  | Direct (n) | Direct (median time) | Coevolve (n)    | QueueMethod (median time) |
|----|------------|----------------------|-----------------|---------------------------|
| 1  | 50         | 203.7 μs             | 50              | 2.6 μs                    |
| 10 | 50         | 30.9  ms             | 50              | 451.1 μs                  |
| 20 | 50         | 221.5 ms             | 50              | 2.2 ms                    |
| 30 | 50         | 776.5 ms             | 50              | 10.7 ms                   |
| 40 | 33         | 1.9 s                | 50              | 23.0 ms                   |
| 50 | 16         | 3.8 s                | 50              | 46.0 ms                   |
| 60 | 10         | 6.6 s                | 50              | 73.1 ms                   |
| 70 | 6          | 11.7 s               | 50              | 120.4 ms                  |
| 80 | 4          | 17.9 s               | 50              | 188.7 ms                  |
| 90 | 3          | 26.8 s               | 50              | 258.4 ms                  |

src/jumps.jl
isaacsas: The issue is you want to pass `JumpSet`s a single heterogeneous vector of different jump types?

If that is the problem it seems like we should just add a new constructor that dispatches on getting a single `AbstractVector` as the input, manually splits the jumps, and calls either the four-argument version or the keyword version.

If you want to PR that (in a new PR please) I'd be fine with adding such a constructor.

isaacsas:
@gzagatti I made some tutorial wording changes and fixes to get it to build docs locally for preview purposes. Can you follow the directions at 

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

to enable maintainers to push to your fork's PR branch so I can push to the PR? Thanks!

gzagatti:
The "Allow edits by maintainers" is activated. Let me know if you are still having trouble pushing to the fork branch.

isaacsas:
Weird, pushing to the pr branch seems to update your repo but not update the PR. I've never had that happen before.

I'm just going to push to a new PR that can be for updates to the docs from this PR. (So we can finish this one as is and I'll just tweak the docs there and merge it after this merges.)

isaacsas:
Nevermind, looks like it did work. Strange.

isaacsas:
@gzagatti just to update you. I've gone through and updated a subset of the tutorials, along with updating some of the new doc strings. I also changed `L` to `rateinterval` as I think we need something more descriptive for this field. If you have a better name though I'm happy to change it again.

Stil TODO on my end:
1. Finish tweaking the docs I haven't yet gotten to.
2. Figure out how we want to handle `JumpProblem`. It seems quite a bit more complicated now as rewritten, so it would be nice to try to simplify it. Also, we'd like to make sure we can handle mixes of the two `VariableRateJump` types now (what I call general vs. bounded `VariableRateJump`s in the tutorial/docs I've updated). Adding a test for this would be good too (i.e. a mix of bounded and unbounded `VariableRateJump`s, and hence requiring being over an `ODEProblem`, but still using `Coevolve` for the bounded jumps).
3. (Future work): After this PR we should figure out why performance is below NRM for non-`VariableRateJump` systems. Given we are using `FunctionWrapper`s there may also be a small gain to be had by keeping `lrate=nothing` when no `lrate` is specified, and skipping the call to the `lrate` function in the rejection sampling code.

A few minor comments for you:
- I updated the docs so they can build locally. As they are dynamic it is good to try that out yourself to make sure everything looks ok and the links/examples all work. You just go in the doc folder, activate the Project.toml there, and call `include("make.jl")`. The resulting build directory should then have the local doc build.
- Since we assume Julia 1.6, you don't need `f(u; p = p)` you can just say `f(u; p)` and if there is a variable with the name `p` it will get picked up for you.

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -371,60 +472,69 @@ both must be constant other than changes due to some other `ConstantRateJump` or
...
   push!(H, integrator.t)
   nothing
 end
-jump3 = VariableRateJump(rate3, affect3!; lrate=lrate, urate=urate, L=L)
+jump3 = VariableRateJump(rate3, affect3!; lrate, urate, rateinterval)
 ```
+Note that here we set the lower bound rate to be the normal SIR infection rate,
+and set the upper bound rate equal to the new rate of infection (`rate3`). As
+long as `u[1]` and `u[2]` are unchanged by another jump, for any `s` in `[t,t +
+rateinterval(u,p,t)]` we have that `lrate(u,p,t) <= rate3(u,p,s) <= urate(u,p,t)`.
```
gzagatti, line 535: I would add an observation that the rates will be re-computed when `u[1]` and `u[2]` change to avoid the reader assuming that `u[1]` or `u[2]` cannot change.
```suggestion
Given `u[1]` and `u[2]`, for any `s` in `[t, t + rateinterval(u,p,t)]` we have 
that `lrate(u,p,t) <= rate3(u,p,s) <= urate(u,p,t)`. Note that the bounds are
automatically recomputed when `u[1]` or `u[2]` changes using the dependency 
graph supplied to `Coevolve`.
```

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -371,60 +472,69 @@ both must be constant other than changes due to some other `ConstantRateJump` or
...
   integrator.u[1] -= 1     # S -> S - 1
   integrator.u[2] += 1     # I -> I + 1
   push!(H, integrator.t)
   nothing
 end
-jump3 = VariableRateJump(rate3, affect3!; lrate=lrate, urate=urate, L=L)
+jump3 = VariableRateJump(rate3, affect3!; lrate, urate, rateinterval)
 ```
+Note that here we set the lower bound rate to be the normal SIR infection rate,
+and set the upper bound rate equal to the new rate of infection (`rate3`). As
```
gzagatti, line 533:
```suggestion
and set the upper bound rate equal to the new rate of infection (`rate3`).
```

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -371,60 +472,69 @@ both must be constant other than changes due to some other `ConstantRateJump` or
...
+long as `u[1]` and `u[2]` are unchanged by another jump, for any `s` in `[t,t +
```
gzagatti, line 534:
```suggestion
```

gzagatti (commented):
Thanks for the doc updates. They look much better. I have two minor comments for you to consider. 

Also, great to learn about a simpler way of passing keyword arguments.

I will build the docs locally to see if everything is working.

By the way, I re-ran the benchmarks. I get closer results. `Coevolve` sometimes beats `NRM`.

| Benchmark                  | Direct  | FRM     | SortingDirect | NRM     | DirectCR | RSSA    | RSSACR  | Coevolve    |
|----------------------------|---------|---------|---------------|---------|----------|-------- |-------- |-------------|
| Diffusion CTRW                  | 4.85 s  | 15.88 s | 1.17 s        | 0.87 s  | 0.43 s   | 1.91 s  | 0.37 s  | 0.77  s      |
| Multistate Model                 | 0.12 s  | 0.22 s  | 0.12 s        | 0.27 s  | 0.20 s   | 0.11 s  | 0.16 s  | 0.28 s      |
| Neg. Feed. Gene Expression          | 0.17 ms | 0.24 ms | 0.21 ms       | 0.44 ms | 0.38 ms  | 0.35 ms | 0.73 ms | 0.40 ms     |
| Marchetti Gene Expression        | 0.44 s  | 0.62 s  | 0.39 s        | 0.75 s  | 0.79 s   | 0.59 s  | 0.83 s  | 0.96 s      |

Some of the differences probably come from my machine, so running them on a dedicated server might provide more consistent results. However, it seems that it is only the Marchetti model that displays substantial and persistent difference.

isaacsas:
@gzagatti do the changes I made look ok to you? Otherwise I think this is good to go (though some more tests mixing bounded and general `VariableRateJump`s, along with the other jump types would be good at some point).

docs/src/tutorials/simple_poisson_process.md
```diff
@@ -175,67 +175,101 @@ by adding or subtracting a constant vector from `u`.
...
+`ConstantRateJump`s to appear before any `VariableRateJump`s. As such, the
+`ConstantRateJump` representing the death process will have internal index 1,
+and our new bounded `VariableRateJump` for birth will have internal index 2.
+Since birth modifies the population size `u[1]`, and death occurs at a rate
+proportional to `u[1]`, when birth occurs we need to recalculate both of rates.
+In contrast, death does not change `u[1]`, and so when death occurs we only need
+to recalculate the death rate. The graph below encodes the dependents of the
+death (`dep_graph[1]`) and birth (`dep_graph[2]`) jumps respectively
 ```@example tut1
-dep_graph = [[2], []]
```
gzagatti: Here I had omitted the self-loops since they are automatically included by the algorithm. My thought is that we might give the user the false impression that if he did not include the self-loops they would not get updated which is wrong.

Second, my understanding is that when jump `1` takes place then jumps `1` and `2` needs to be updated given that the dependencies come from `deps = p.dep_gr[p.next_jump]`. We have that `deathrate(u,p,t) = p.μ * u[1]`, so it should be updated when `p.next_jump == 1`. 

On the other hand, when jump `2` takes place the rate of `1` does not need an update.

Therefore, I derived `dep_grap = [[2], []]`. I think your description is also in agreement with this.

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -499,19 +446,191 @@ function that gets evaluated is
...
+
+## [Defining the Jumps Directly using a bounded `VariableRateJump`](@id VariableRateJumpWithBnds)
+
+Assume that the infection rate is now decreasing over time. That is, when
+individuals get infected they immediately reach peak infectivity. The force of
+infection then decreases exponentially to a basal level. In this case, we must
+keep track of the time of infection events. Let the history ``H(t)`` contain the
+timestamps of all ``I(t)`` active infections. The rate of infection is then
+```math
+\beta S(t) I(t) + \alpha S(t) \sum_{t_i \in H(t)} \exp(-\gamma (t - t_i))
```
gzagatti, line 492: To agree with code I would use $\beta_1$

```suggestion
\beta_1 S(t) I(t) + \alpha S(t) \sum_{t_i \in H(t)} \exp(-\gamma (t - t_i))
```

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -499,19 +446,191 @@ function that gets evaluated is
...
+```
+where ``\beta`` is the basal rate of infection, ``\alpha`` is the spike in the
```
gzagatti, line 494:
```suggestion
where ``\beta_1`` is the basal rate of infection, ``\alpha`` is the spike in the
```

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -499,19 +446,191 @@ function that gets evaluated is
...
+rate of infection, and ``\gamma`` is the rate at which the spike decreases. Here
+we choose parameters such that infectivity rate due to a single infected
+individual returns to the basal rate after spiking to ``\beta + \alpha``. In
```
gzagatti, line 497:
```suggestion
individual returns to the basal rate after spiking to ``\beta_1 + \alpha``. In
```

gzagatti (commented):
Thanks for your review. 

I like the simplifications to the `JumpProblem` and the `Coevolve` aggregator.

I have made some comments on the documentation, otherwise it's all good on my side.

docs/src/jump_types.md
```diff
@@ -26,116 +27,111 @@ a stochastic differential equation to have jumps is commonly known as a Jump
...
-optimized for exact methods (also known in the biochemistry literature as SSA
-algorithms). `ConstantRateJump`s, `VariableRateJump`s and `MassActionJump`s can
-be added to standard DiffEq algorithms since they are simply callbacks, while
-`RegularJump`s require special algorithms.
+Since exact methods simulate each individual jump, they may become
+computationally expensive to simulate processes over timescales that involve
+*many* jump occurrences. As an alternative, inexact τ-leaping methods take
+discrete steps through time, over which they simultaneously execute many jumps.
+These methods can be much faster as they do not need to simulate the realization
+of every individual jump event. τ-leaping methods trade accuracy for speed, and
```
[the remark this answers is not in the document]
isaacsas: I think it would make sense to add some tau-leap examples to the first and second tutorials, and we can mention this there. It isn’t well documented,meted currently anyways…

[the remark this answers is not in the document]
isaacsas: I’m going to leave that for a follow up PR.

docs/src/tutorials/simple_poisson_process.md
isaacsas: Jump 1 is death as it is constant rate here, jump 2 is birth as it is variable rate.

gzagatti: Ok, got it. Where I got confused was in `jprob = JumpProblem(dprob, Coevolve(), vrj1, deathcrj; dep_graph)`, since `vrj1` comes before `deathcrj`

But as we had discussed before and is somewhere else in the documentation, constant rate comes before variable rates in the dependency graph. Maybe worth pointing this out for the inattentive eye.

isaacsas: Yup, the paragraph I added before the example explains this.

isaacsas:
All merged. Thanks @gzagatti thus is a great PR.

gzagatti:
You're welcome. Thanks for your support @isaacsas!

palmtree2013:
Hi, thank you for the great work to support `VariableRateJump` type! Maybe I missed something, in https://github.com/SciML/JumpProcesses.jl/blob/2afd12369ef5e848543a9e315d2ab02c8f44a667/src/aggregators/coevolve.jl#L30
it seems to me that `!isempty(rs)` should check whether the JumpProblem has either `ConstantRateJump` or `VariableRateJump`. But in the construction of the `Coevolve` aggregator, `rates` only refers to `VariableRateJump`
https://github.com/SciML/JumpProcesses.jl/blob/2afd12369ef5e848543a9e315d2ab02c8f44a667/src/aggregators/coevolve.jl#L66
should the condition `!isempty(rs)` be changed to `!isempty(urates)` instead? 
Thank you!

isaacsas:
Hi-, good catch. Could you submit a PR with that change?

PR #277, opened 2022-12-10
Move spatial tutorial from SciMLTutorials

ChrisRackauckas:
Now that we have package docs for all of the packages and doctests, it makes sense for the tutorials to be built as part of the testing process. It also keeps everything organized by package. Thus SciMLTutorials is getting cannibalized into separate package docs, and this is the one for JumpProcesses. I did a few updates (DiffEqJump, LightGraphs, links) so hopefully this is enough. I might need a little bit of help getting this one completed though, since there are a few I need to tend to.


ChrisRackauckas:
@Vilin97 @isaacsas

isaacsas:
I can finish this one off.

ChrisRackauckas:
Thanks, I'll just close the tab then and move onto the next!

ChrisRackauckas:
Looks like this just worked. Okay!

isaacsas:
The tutorial needs some updates, but I can followup with those.

PR #282, opened 2023-01-05
Tiny fix of `Coevolve()` with the dependency graph check

palmtree2013:
this pull request is related to https://github.com/SciML/JumpProcesses.jl/pull/276#issuecomment-1372219257


ChrisRackauckas:
Can you add a test which catches this? I guess it should be a case with mixed jumps?

ChrisRackauckas:
Use Catalyst to generate the problem so it has the dependency graph? Or use one of the hand-built ones and extend it?

isaacsas:
I just added a test for this based on the MWE.

isaacsas (approved):
GTG if tests pass I think.

isaacsas:
Thanks @palmtree2013

palmtree2013:
Thank you for adding the test

PR #283, opened 2023-01-05
Out of place variable rate jump extensions

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/923

But for some reason it fails because `jumps` is not a JumpSet?

### DIFF
--- a/src/problem.jl
+++ b/src/problem.jl
@@ -273,7 +273,7 @@ function extend_problem(prob::DiffEqBase.AbstractODEProblem, jumps; rng = DEFAUL
 
     ttype = eltype(prob.tspan)
     u0 = ExtendedJumpArray(prob.u0,
-                            [-randexp(rng, ttype) for i in 1:length(jumps)])
+                           [-randexp(rng, ttype) for i in 1:length(jumps)])
     remake(prob, f = ODEFunction{isinplace(prob)}(jump_f), u0 = u0)
 end
 
--- a/src/problem.jl
+++ b/src/problem.jl
@@ -254,22 +254,46 @@ end
 function extend_problem(prob::DiffEqBase.AbstractODEProblem, jumps; rng = DEFAULT_RNG)
     _f = SciMLBase.unwrapped_f(prob.f)
 
-    jump_f = let _f = _f
-        function jump_f(du::ExtendedJumpArray, u::ExtendedJumpArray, p, t)
-            _f(du.u, u.u, p, t)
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
@isaacsas variable rate jump tests are failing on the latest release?

src/problem.jl
```diff
@@ -254,22 +254,46 @@ end
...
 
-    jump_f = let _f = _f
-        function jump_f(du::ExtendedJumpArray, u::ExtendedJumpArray, p, t)
-            _f(du.u, u.u, p, t)
-            update_jumps!(du, u, p, t, length(u.u), jumps...)
+    if isinplace(prob)
+        jump_f = let _f = _f
+            function (du::ExtendedJumpArray, u::ExtendedJumpArray, p, t)
+                _f(du.u, u.u, p, t)
+                update_jumps!(du, u, p, t, length(u.u), jumps.variable_jumps...)
```
isaacsas, line 261: `extend_problem` is now passed just the variable jumps from `JumpProblem`, and not the full jump set since that is all that it was using. I did this because we now filter the types of variable rate jumps from `jumps.variable_jumps`, so only a subset of them may be wanted / handled here.

isaacsas: Otherwise `JumpProblem` would need to create a new `JumpSet` that just holds the new (subset) of `VariableRateJump`s needed here.

isaacsas, line 261:
```suggestion
                update_jumps!(du, u, p, t, length(u.u), jumps...)
```

ChrisRackauckas: I see, it's because I was working off an old master branch, rebasing makes all of the failures go away except in my code.

src/problem.jl
```diff
@@ -254,22 +254,45 @@ end
...
+                return du
+            end
         end
     end
+
     ttype = eltype(prob.tspan)
     u0 = ExtendedJumpArray(prob.u0,
-                           [-randexp(rng, ttype) for i in 1:length(jumps)])
-    remake(prob, f = ODEFunction{true}(jump_f), u0 = u0)
+                            [-randexp(rng, ttype) for i in 1:length(jumps.variable_jumps)])
```
isaacsas, line 276:
```suggestion
                            [-randexp(rng, ttype) for i in 1:length(jumps)])
```

ChrisRackauckas:
For some reason my local formatter is acting up so I'll merge, someone just needs to format some time.

PR #288, opened 2023-01-30
Modifies rateinterval API to accept computed `urate`.

gzagatti:
A problem with the current implementation of `Coevolve` is that it might compute `urate` twice. Once when computing the actual `urate` and another time if `rateinterval` calls `urate`.

This problem is clear in the Hakwes benchmark. Below you can see a profile of the simulation:

image

This PR modifies the API of `rateinterval` to `rateinterval(u, p, t, urate)`. With this modification, you can see that much less time is spent in `rateinterval`.

image

The downside is a more verbose API. The API will also break for when using `VariableRate` with `Coevolve`. However, since the aggregator is relatively new, it might be worth considering this change. I am happy to discuss more elegant alternatives.

### DIFF
--- a/src/aggregators/coevolve.jl
+++ b/src/aggregators/coevolve.jl
@@ -1,7 +1,7 @@
 """
 Queue method. This method handles variable intensity rates.
 """
-mutable struct CoevolveJumpAggregation{T, S, F1, F2, RNG, GR, PQ} <:
+mutable struct CoevolveJumpAggregation{T, S, F1, F2, F3, RNG, GR, PQ} <:
                AbstractSSAJumpAggregator
     next_jump::Int                    # the next jump to execute
     prev_jump::Int                    # the previous jump that was executed
@@ -18,7 +18,7 @@ mutable struct CoevolveJumpAggregation{T, S, F1, F2, RNG, GR, PQ} <:
     pq::PQ                            # priority queue of next time
     lrates::F1                        # vector of rate lower bound functions
     urates::F1                        # vector of rate upper bound functions
-    rateintervals::F1                 # vector of interval length functions
+    rateintervals::F3                 # vector of interval length functions
     haslratevec::Vector{Bool}         # vector of whether an lrate was provided for this vrj
 end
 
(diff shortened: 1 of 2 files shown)


isaacsas:
Let me think about the interface here a bit. This may be the easiest way to go, but I wonder if it would be better to have some kind of dispatch-based resolution where one can either have `urate` return both, or you provide two independent functions.

gzagatti:
Take your time. It might also be worth taking into account how to handle the `lrate` API as well, since it is called after `urate` and `rateinterval`.

I tried playing with the multiple dispatch mechanism but could not find an elegant solution.

gaurav-arya:
Seconding the idea of allowing a single function to return multiple quantities. Also, to throw another spanner into the works: in my use case, it would be most efficient to compute all three of `rate`, `urate`, and `rateinterval` jointly in a single function, as they all share some intermediary work:)

gaurav-arya:
Perhaps it would be best to just have a separate kwarg? E.g. the user can specifiy `rate_and_bounds(u,p,t)` returning the triplet of `(rate(u,p,t), urate(u,p,t), rateinterval(u,p,t))`. If only `rate_and_bounds` were specified and not e.g. `rate`, then we could populate `rate`, `urate` and `rateinterval` as `rate_and_bounds(u,p,t)[1]` etc., so that algorithms which don't know about this special form (e.g. those that don't care about the bounds) still work as normal. But we could specialize on it in Coevolve.

isaacsas:
This requires some thought since I think frequently one won't want to evaluate the rate simultaneous to the rate bounds for performance reasons, so we'd also want to support versions that don't evaluate `rate` until it is needed in the sampling. We also need to consider that we don't want to get in a situation where we have type instabilities by getting a mix of functions with different return signatures / have to hard code a huge number of possible cases.

I wonder if we might instead have a way of communicating to users that we haven't updated time and/or the system state since the last time one of the rate functions was called (so a user could just cache the values and know they are still valid if it makes sense for them to generate them all at once).

PR #290, opened 2023-02-05
Fix ldiv! ambiguity on v1.9

ChrisRackauckas:
Fixes https://github.com/SciML/JumpProcesses.jl/issues/289


isaacsas:
Does this mean we'll have to add this specialization for other factorization methods / types too?

ChrisRackauckas:
It's just for the stiff ODE solves, which will always use LU here.

isaacsas:
What about a sparse jacobian?

PR #307, opened 2023-03-28
Direct method allocations fix

isaacsas:
Closes https://github.com/SciML/JumpProcesses.jl/issues/305

@ChrisRackauckas could you look over the generated function I'm using. I'm always iffy on what is ok to do with them...


isaacsas:
Also, some minimal benchmarking using the example in the `makeprob` function within the test gives that this is 3x faster than using `FunctionWrappers` on that system.

ChrisRackauckas (approved):

PR #309, opened 2023-04-03
function wrapper type instability fix

isaacsas:
@ChrisRackauckas what do you think about this approach? Basically, when not using tuples for affects we store them as `affects!::Any` (but they are concretely a `Vector{Any}`). When `init` is called the `Vector{Any}` that is stored is concretized into a vector of type-stable function wrappers using the integrator type as input. Then I add a manual dispatch when `affects!` isn't a tuple to handle pulling out this vector from the `affects!::Any` field.

Preliminary testing is that overall I'm seeing a half-percent or less speed decrease when using `Direct` with tuples (so still much faster than before the recent generated function approach I put in), but a ~17% or more speed up when using function wrappers. Moreover, the memory allocation issues go away, so I think we now have type-stability outside the highest level call into the jump aggregator callback.

If this approach looks reasonable I'll cleanup and refactor the code from `Direct` to work with all the SSAs. 

One nice thing about the setup here is that it wouldn't be too bad to support affect! tuples for any method now, which can give a nice speed up (Direct vs. DirectFW on a small system is still like a 40% speed difference). We might want to add a similar manual dispatch for `rate` functions so that we can also support them as tuples with other SSAs. I’m thinking we could then switch based on the number of jumps to automatically use tuples vs. function wrappers for any method.

### DIFF
--- a/src/aggregators/coevolve.jl
+++ b/src/aggregators/coevolve.jl
@@ -2,7 +2,7 @@
 Queue method. This method handles variable intensity rates.
 """
 mutable struct CoevolveJumpAggregation{T, S, F1, F2, RNG, GR, PQ} <:
-               AbstractSSAJumpAggregator
+               AbstractSSAJumpAggregator{T, S, F1, F2, RNG}
     next_jump::Int                    # the next jump to execute
     prev_jump::Int                    # the previous jump that was executed
     next_jump_time::T                 # the time of the next jump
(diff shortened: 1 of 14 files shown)


ChrisRackauckas:
This seems reasonable. Though I do think that we should be moving to a scheme where anything cached is in the `init` phase and when restructuring I think the `Any` can be avoided by directly declaring the type of the integrator. But that may never happen, so this is very good in the meantime.

isaacsas:
I'd like to switch to such a setup too as that would also get rid of @gaurav-arya's issue, however, I'm not in favor of doing so until we have a nice way to rerun the same simulation many times for sampling without a user having to drop into the integrator interface. (i.e. I think `EnsembleProblem`s would need updating to also then use the integrator interface and `reinit` between runs on a thread.) 

But in any case I think we'd still have this issue. Fundamentally, the problem is a circular reference in that the integrator stores the affects, but the wrapped affects need to know the type of the integrator. We'd need some other object to cache the JumpAggregation, which would have to be created after the integrator to avoid this.

PR #310, opened 2023-04-04
Setfields for jump problems

TorkelE:
Two things:

enables `setdinex!` and `getindex` for `JumpProblems. E.g.
```julia
using Catalyst, JumpProcesses
rn = @reaction_network begin
    (p,d), 0 <--> X
end

u0 = [:X => 0]
p = [:p => 1.0, :d => 0.1]
dp = DiscreteProblem(rn,u0,(0.0,10.0),p)
jp = JumpProblem(rn,dp,Direct())

jp[:X]
jp[:p]
jp[:p] = 1.5
jp[:X] = 15
```

It also adds a check so that if `setindex!` is called on a `JumpProblem` then `update_parameters!(prob.massaction_jump, prob.prob.p)` is also called

### DIFF
--- a/src/JumpProcesses.jl
+++ b/src/JumpProcesses.jl
@@ -7,7 +7,7 @@ using RandomNumbers, TreeViews, LinearAlgebra, Markdown, DocStringExtensions
 using DataStructures, PoissonRandom, Random, ArrayInterface
 using FunctionWrappers, UnPack
 using Graphs
-using SciMLBase: SciMLBase
+using SciMLBase: SciMLBase, ___internal_setindex!
 using Base.FastMath: add_fast
 
 import DiffEqBase: DiscreteCallback, init, solve, solve!, plot_indices, initialize!
--- a/src/JumpProcesses.jl
+++ b/src/JumpProcesses.jl
@@ -7,7 +7,7 @@ using RandomNumbers, TreeViews, LinearAlgebra, Markdown, DocStringExtensions
 using DataStructures, PoissonRandom, Random, ArrayInterface
 using FunctionWrappers, UnPack
 using Graphs
-using SciMLBase: SciMLBase, ___internal_setindex!
+using SciMLBase: SciMLBase
 using Base.FastMath: add_fast
 
 import DiffEqBase: DiscreteCallback, init, solve, solve!, plot_indices, initialize!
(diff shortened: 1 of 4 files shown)


TorkelE:
I would add some additional tests, but that would require adding MTK as a test dependency. Would that be preferred?

isaacsas:
I don't think we want to have a test dependency on ModelingToolkit here, the dependencies / tests for this should be comparable to what is done in SciMLBase I'd think.

src/problem.jl
```diff
@@ -119,6 +119,19 @@ function DiffEqBase.remake(thing::JumpProblem; kwargs...)
       thing.kwargs)
 end
 
+# when setindex! is used.
+function Base.setindex!(prob::JumpProblem, args...; kwargs...)
+    ___internal_setindex!(prob.prob, args...; kwargs...)
```
isaacsas, line 124:
```suggestion
   SciMLBase.___internal_setindex!(prob.prob, args...; kwargs...)
```

src/JumpProcesses.jl
```diff
@@ -7,7 +7,7 @@ using RandomNumbers, TreeViews, LinearAlgebra, Markdown, DocStringExtensions
 using DataStructures, PoissonRandom, Random, ArrayInterface
 using FunctionWrappers, UnPack
 using Graphs
-using SciMLBase: SciMLBase
+using SciMLBase: SciMLBase, ___internal_setindex!
```
isaacsas, line 10:
```suggestion
using SciMLBase: SciMLBase
```

isaacsas (changes requested):
I'd avoid the explicit import of the SciMLBase function, but otherwise looks good.

isaacsas:
Maybe just add a simple example that you use to check this is working by directly building a `JumpProblem` instead of going through MTK?

TorkelE:
> I'd avoid the explicit import of the SciMLBase function, but otherwise looks good.

This is a good point

TorkelE:
> Maybe just add a simple test that you use to check this is working by directly building a JumpProblem instead of going through MTK?

Will try to make some simple test not requiring MTK

TorkelE:
Actually, is that possible?

This is only for the cases when a `prob` has a `f` which has a `sys`. However, is there some non-MTK based way of creating systems? Even in SciMLBase we have to make downstream tests that depend on these.

This is only relevant when you have symbols like `:X` and `:p` associated with variables/parameters, but I don't think this is ever the case in JumpProcesses before e.g. MTK gets involved? I will look at it a bit more, but not sure if I will be able to figure it out.

TorkelE:
Thanks a lot! Will use this

PR #318, opened 2023-04-24
Fix rssa allocations

isaacsas:
Closes https://github.com/SciML/JumpProcesses.jl/issues/317

Adds allocations test for all non-spatial aggregators to ensure memory use is independent of the number of time-steps when saving is disabled.

### DIFF
--- a/src/aggregators/bracketing.jl
+++ b/src/aggregators/bracketing.jl
@@ -64,13 +64,27 @@ get brackets for reaction rx by first checking if the reaction is a massaction r
     end
 end
 
-# set up bracketing
-function set_bracketing!(p::AbstractSSAJumpAggregator, u, params, t)
-    # species bracketing interval
+function update_u_brackets!(p::AbstractSSAJumpAggregator, u::AbstractVector)
     @unpack ulow, uhigh = p
     @inbounds for (i, uval) in enumerate(u)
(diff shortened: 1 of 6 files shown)


src/aggregators/bracketing.jl
```diff
@@ -64,13 +64,27 @@ get brackets for reaction rx by first checking if the reaction is a massaction r
     end
 end
 
+function update_u_brackets!(p::AbstractSSAJumpAggregator, u::AbstractVector)
```
isaacsas, line 67:
```suggestion
@inline function update_u_brackets!(p::AbstractSSAJumpAggregator, u::AbstractVector)
```

src/aggregators/bracketing.jl
```diff
@@ -64,13 +64,27 @@ get brackets for reaction rx by first checking if the reaction is a massaction r
...
+    @unpack ulow, uhigh = p
+    @inbounds for (i, uval) in enumerate(u)
+        ulow[i], uhigh[i] = get_spec_brackets(p.bracket_data, i, uval)
+    end
+    nothing
+end
+
+function update_u_brackets!(p::AbstractSSAJumpAggregator, u::SVector)
```
isaacsas, line 75:
```suggestion
@inline function update_u_brackets!(p::AbstractSSAJumpAggregator, u::SVector)
```

PR #328, opened 2023-07-11
Allocations test fix

isaacsas:
It seems that it currently takes a couple runs of RSSACR before the table allocations are fully settled down now. It is expected that there can be some allocation differences here for the algorithms that use PriorityTables (which can dynamically allocate additional buckets, but only ever grow, so should settled down eventually).

### DIFF
--- a/test/allocations.jl
+++ b/test/allocations.jl
@@ -1,10 +1,6 @@
 using Test, JumpProcesses
 using StableRNGs
 
-# note this test appears to be sensitive to the seed for some algorithms (RSSACR)
-# this appears due to dynamic allocations within the priority table as the timescale changes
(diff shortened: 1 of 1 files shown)


isaacsas:
My only confusion is to why this issue has cropped up now all of a sudden.

ChrisRackauckas:
This popped up when v1.9 came out.

isaacsas:
Huh, I never saw that issue before (even with this PR tests passed). 

In any case I did dig into it with the allocation profiler and it seemed to be coming from additional `push!` into groups within the priority table as best I could tell (and not the function wrapper type instability issue we had before). Moving this test to use `StableRNGs` should hopefully eliminated the issue.

PR #329, opened 2023-07-12
format

ChrisRackauckas:


isaacsas:
Can we hold off on this till the update to the formatter style is made? I'd prefer to only have to go through and check the mass changes it makes only once...

PR #331, opened 2023-07-13
Fix Coevolve calling urate beyond tspan

### LINKED ISSUE #330: `Coevolve` can call `urate` with a time beyond the time span
MWE
```julia
using JumpProcesses

test_rate(u, p, t) = 3.0 
test_affect!(integrator) = (integrator.u[1] += 1)
function test_urate(u, p, t) 
    if t > 1.0
        error("test_urate does not handle t > 1.0")
    else
        return 4.0
    end
end

test_jump = VariableRateJump(test_rate, test_affect!; urate=test_urate, rateinterval=(u,p,t)->0.9)

dprob = DiscreteProblem([0], (0.0, 1.0), nothing) 
jprob = JumpProblem(dprob, Coevolve(), test_jump; dep_graph=[[1]])

for i in 1:100
    solve(jprob, SSAStepper())
end 
# sometimes throws ERROR: test_urate does not handle t > 1.0
```
I had a practical case where I ran into the above issue because my supplied `urate` function actually went haywire for `t` beyond my timespan. In general I think it would be a good property to only call the user-supplied functions for `t` in the timespan?

gaurav-arya:
Fix #330


isaacsas:
@gzagatti can you give this a review?

My plan is to merge this, make a bug fix release, and then hopefully merge your PR and make a minor release. It would be good after this is merged to update your PR with the test and ensure it doesn't have the same issue.

src/aggregators/coevolve.jl
```diff
@@ -175,6 +175,7 @@ function next_time(p::CoevolveJumpAggregation{T}, u, params, t, i, tstop::T) whe
             rateinterval = get_rateinterval(p, lidx, u, params, t)
             if s > rateinterval
                 t = t + rateinterval
+                (t >= tstop) && break
```
gzagatti, line 178: This wouldn’t work, because when you break you haven’t updated `_t` which is what is returned by the function.

Alternatively, this should give a correct trajectory:

```suggestion
                if t >= tstop
                    _t = t
                    break
                end
```

gzagatti:
There's a subtle problem with the proposed fix.

Basically, the function `next_time` is designed to re-use rng in [line 168](https://github.com/SciML/JumpProcesses.jl/blob/d3a1e33682d38416a688027c32782add9a4ecb13/src/aggregators/coevolve.jl#L168). When you break before computing the next `urate`, you will be basically re-using the  rng in the next interaction which will give incorrect samples. The correct way to proceed is to re-compute `urate` in those cases rather than recycling the rng. 

This is one of the issues addressed by PR #315. In particular see [line 260](https://github.com/SciML/JumpProcesses.jl/blob/ba5aaf3c88921ff00c0ffafb02cb5e2026f6a91f/src/aggregators/coevolve.jl#L260) and [line 290](https://github.com/SciML/JumpProcesses.jl/blob/ba5aaf3c88921ff00c0ffafb02cb5e2026f6a91f/src/aggregators/coevolve.jl#L290). The new implementation of Coevolve ensures that we only compute `urate`, `lrate` and `rate` for the present time `t`. In other words, it is synced with the model time.

This is so subtle that on my first implementation of PR #315 I could not figure out why some of the samples from Coevolve were slightly different from other algorithms. It was poor recycling of the rng that caused the issue.

With that being said, I think it is valid to add the proposed tests. I would go even further and ensure that `lrate`, `rate` and `urate` are not called beyond `tstop`. Also you could probably improve the test speed by reducing `tstop` (or decresing the rates) and setting `rateinterval` equal to `tstop`.

gaurav-arya:
Okay, I've reverted the fix and applied your suggestions to the test. I think this could be ready to merge as a broken test that your PR would then fix, or we can wait until #315 lands

isaacsas:
I'm going to merge 315 shortly, and then if you want to tweak the version of this test there just update this PR and I'll merge it.

Thanks!

isaacsas:
version of the test "here" I mean...

PR #334, opened 2023-07-19
Added ExtendedJumpArray dispatch for ODE_DEFAULT_NORM

meson800:
This change reduces the number of allocations (to zero!) required to take the norm of an ExtendedJumpArray. Previously, ODE_DEFAULT_NORM fell back to a default, allocating implementation.

I also added tests to check that this norm definition is equivalent to the old one, using (stable) random ExtendedJumpArrays. There's also a test that checks that the new norm dispatch is not allocating.

This should close SciML/DifferentialEquations.jl#969

### DIFF
--- a/src/extended_jump_array.jl
+++ b/src/extended_jump_array.jl
@@ -95,6 +95,11 @@ function Base.similar(A::ExtendedJumpArray, ::Type{S},
     ExtendedJumpArray(similar(A.u, S), similar(A.jump_u, S))
 end
 
+# ODE norm to prevent type-unstable fallback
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

isaacsas:
Thanks @meson800! when you are done with updates could you run JuliaFormatter version 1.0.32 on the PR to pass the formatting check?

meson800:
Formatting and changes done, should be good to go! (but needs another CI run approval)

isaacsas:
Thanks @meson800 for taking care of this!

isaacsas:
9.7.2 with your fix should be out shortly.

PR #336, opened 2023-08-21
removes stackengine dependency

gzagatti:
This PR removes the dependency of the paper on `stackengine.sty` which raised compilation errors by [`whedon`, the compilation bot](https://github.com/JuliaCon/proceedings-review/issues/132#issuecomment-1681752237).

We were using `stackengine.sty` to produce better looking under-bars. We remove the dependency and add a custom function for producing under-bars.


gzagatti:
Thanks!

PR #337, opened 2023-08-22
Adds subfig.sty

gzagatti:
The Julia Proceedings compilation environment does not have `subfig.sty` which is added to the paper repo.

I probably going to make a few more of these types of PRs until we get `whedon` to compile the paper. So I appreciate your patience on that and apologies in advance for the multiples PRs.

### DIFF
--- a/paper/paper.tex
+++ b/paper/paper.tex
@@ -12,7 +12,6 @@
 \usepackage{mathtools}
 \usepackage{soul}
 
-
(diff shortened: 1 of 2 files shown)


isaacsas:
@gzagatti I'll try to keep an eye on notifications, but feel free to ping me on Slack to get these merged quick. They don't require review...

PR #338, opened 2023-08-22
Juliacon2023

gzagatti:
The Julia Proceedings compilation environment does not have multirow.sty which is added to the paper repo.

### DIFF
--- a/paper/algcompatible.sty
+++ b/paper/algcompatible.sty
@@ -1,0 +1,89 @@
+% ALGORITHMIC STYLE -- Released 27 APR 2005
+%    for LaTeX version 2e
+%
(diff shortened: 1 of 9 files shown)


isaacsas:
Maybe add every package style file you use so we don’t have to deal with 100 PRs?

gzagatti:
I've added all the listed dependencies listed in the preamble of the paper.

If this doesn't work I can include the dependencies that come with the latex distribution. For future reference, I found out one can list all dependencies used to compile the paper to `deps.log` using the `-deps-out` flag as below.

```sh
latexmk --lualatex -deps-out=deps.log paper.tex
```

PR #340, opened 2023-08-30
Add non-default broadcasting of ExtendedJumpArray's.

### LINKED ISSUE #335: Efficency of ExtendedJumpArray broadcasting in ode_interpolant
# Background
Continuing to optimize a system with VariableRateJumps and callbacks, I've found that performing ODE interpolations on ExtendedJumpArray's is consuming around ~80% of the total runtime of the solve. About 10% is actually in the various `do_step` backtraces, with most of the runtime is in the `find_first_continuous_callback` call. Of that, the vast majority is happening in `Base.getindex(A:ExtendedJumpArray)`

# Problem
In particular, it looks like the combination of `ExtendedJumpArray`'s, `@muladd`, and `FastBroadcast` is causing unoptimal code generation. For my use case, the problematic function is [here](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/dense/interpolants.jl#L408-L415):
```
@muladd function _ode_interpolant!(out, Θ, dt, y₀, y₁, k,
    cache::Union{Tsit5ConstantCache, Tsit5Cache},
    idxs::Nothing, T::Type{Val{0}})
    @tsit5pre0
    @inbounds @.. broadcast=false out=y₀ +
                                      dt *
                                      (k[1] * b1Θ + k[2] * b2Θ + k[3] * b3Θ + k[4] * b4Θ +
                                       k[5] * b5Θ + k[6] * b6Θ + k[7] * b7Θ)
    out
end
```
but it seems to affect others as well.

Examining this with Cthulhu and friends, looking at the LLVM and native code, etc etc shows an absolute morass of branching, with something like 4800 (!) branches in the assembly. The branches are coming from inlining where it's using the ExtendedJumpArray index check. This churn means that the compiler doesn't seem to be able to loop unroll, since it has to repeatedly check "is index < length(jump_array.u)", over and over and over.

# Solution
I'm guessing that this would be a lot faster and not be churning the branch predictor with a bunch of mostly-useless (<=) calls to switch between the `.u` and `.jump_u` members, if we could somehow turn these calls into separate calls like
```
out.u = y0.u + dt * ...
out.jump_u = y0.jump_u + dt * ...
```

I fixed this by adding this dispatch to my usercode (I unpacked one of the macros because getting the imports right was annoying, but it's just the original ode_interpolant with the `.u` and `.jump_u` parts unrolled):
```
using MuladdMacro, FastBroadcast
using OrdinaryDiffEq: constvalue, Tsit5Interp
function OrdinaryDiffEq._ode_interpolant!(out::ExtendedJumpArray, Θ, dt, y₀::ExtendedJumpArray, y₁, k,
    cache::Union{OrdinaryDiffEq.Tsit5ConstantCache, OrdinaryDiffEq.Tsit5Cache},
    idxs::Nothing, T::Type{Val{0}})
    OrdinaryDiffEq.@tsit5unpack
    Θ² = Θ * Θ
    b1Θ = Θ * @evalpoly(Θ, r11, r12, r13, r14)
    b2Θ = Θ² * @evalpoly(Θ, r22, r23, r24)
    b3Θ = Θ² * @evalpoly(Θ, r32, r33, r34)
    b4Θ = Θ² * @evalpoly(Θ, r42, r43, r44)
    b5Θ = Θ² * @evalpoly(Θ, r52, r53, r54)
    b6Θ = Θ² * @evalpoly(Θ, r62, r63, r64)
    b7Θ = Θ² * @evalpoly(Θ, r72, r73, r74)
    @muladd @inbounds FastBroadcast.@.. broadcast=false out.u=y₀.u +
                                      dt *
                                      (k[1].u * b1Θ + k[2].u * b2Θ + k[3].u * b3Θ + k[4].u * b4Θ +
                                       k[5].u * b5Θ + k[6].u * b6Θ + k[7].u * b7Θ)
    @muladd @inbounds FastBroadcast.@.. broadcast=false out.jump_u=y₀.jump_u +
                                      dt *
                                      (k[1].jump_u * b1Θ + k[2].jump_u * b2Θ + k[3].jump_u * b3Θ + k[4].jump_u * b4Θ +
                                       k[5].jump_u * b5Θ + k[6].jump_u * b6Θ + k[7].jump_u * b7Θ)
    out
end
```
This dramatically reduces the runtime, and inspecting the LLVM code shows "only" 123 branches with good loop unrolling. There's still a boat-load of allocations happening in `handle_callbacks`, but I'll deal with that with a separate issue/PR.

# Questions
- Is it possible to write a dispatch for the `@..` macro to do this type of code generation automatically?
- If not, an annoying approach would be to manually add ExtendedJumpArray dispatches to [interpolants.jl](https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/src/dense/interpolants.jl), but that's a lot more fragile. I'm also not sure where this dispatch would go, as adding this dispatch into JumpProcesses would also come with a dependency on `OrdinaryDiffEq`.
- Is there a better approach?

meson800:
# Summary

The previous ExtendedJumpArray code contains code that implements part of the broadcastable interface, but is not actually used due to the entire interface not being included. This commit replaces that broadcast code, reusing much of the arg packing/repacking. Broadcasting operations that only rely on scalars and `ExtendedJumpArray`'s will now use an efficient implementation that can emit SIMD instructions at the native and LLVM code levels. This closes #335. Benchmarking shows between a two to five fold increase in performance with this broadcasting method.

The previous code uses the fallback indexing method when broadcasting, which requires branching inside the broadcast loop. This branch is likely well-branch-predicted, but the presence of the branch prevents kernel functions from being fused into efficient SIMD instructions.

# Implementation
This PR defines a ExtendedJumpArrayStyle, a broadcast style that includes two sub-styles that represent the styles for broadcasting the `u` and `jump_u` entries. `copyto!` overloads are specified that do the efficient thing and call `copyto!`, using the broadcast computation tree, on `u` and `jump_u` separately. Binary style rules are defined such that broadcasts fall back to the old method if you do something like adding an ExtendedJumpArray to a normal vector.

The arg tuple unpacking is mostly the same as the old partial implementation that was there; I just renamed the functions and unified them with the new broadcasting style.

A Julia >=1.9 extension is added for FastBroadcast.jl, where the equivalent of the `copyto!` call, `fast_materialize!` is similarly overloaded. This only adds a weak dependency, so only users who also use `FastBroadcast` (such as `OrdinaryDiffEq`) will have these methods loaded; those that don't already have FastBroadcast in the environment will not pull it in as a dependency.

Tests are added to check that the new broadcasting behavior is behaving as expected. In addition, a benchmark file is included that can be used to confirm that `ExtendedJumpArray` broadcasting is just as fast as broadcasting of normal `Vector{Float64}`'s. This PR passes all other tests, including e.g. the ones that use auto differentiation; broadcasting works great even in this case :)

# Remaining problems/notes
Inhomogeneous `ExtendedJumpArray`'s still act strangely under broadcasting, but no more strangely than they were already acting. In fact, some of the VariableRateJump tests rely on the behavior that an `ExtendedJumpArray` plus a `Vector` decay to become a `Vector`.

The new broadcasting method means that if you define an _output_ array that is inhomogeneous (e.g. one where the `A.u` is a `Vector{Float64}` but `A.jump_u` is a `Vector{Int64}`), then broadcasted operations will successfully maintain the inhomogeneity. However, if you allow the broadcast machinery to create an output array for you (allocating), it will just allocate a homogenous ExtendedJumpArray with both `u` and `jump_u` being `Vector{Float64}`'s. This is the same behavior as the fallback, so this should not break any existing downstream code.

Finally, I just used an implementation for selecting an output `ExtendedJumpArray` while broadcasting based on the [Julia interface documentation](https://docs.julialang.org/en/v1/manual/interfaces/#Selecting-an-appropriate-output-array); this is a recursive function that is used to unpack the broadcasted tuple in a type-stable way, but there can be a few allocations that happen prior to the broadcast if someone broadcasts a very complicated expression that exceeds a recursion depth of 16. I was trying to replace it with a concise generated function that would always be type stable, but I think it is far less likely to hit this limit than in my other recent PRs. If there is an instance where this happens, I can fix this.


Project.toml
```diff
@@ -39,6 +42,9 @@ TreeViews = "0.3"
 UnPack = "1.0.2"
 julia = "1.6"
 
+[extensions]
+FastExtendedJumpArrayBroadcast = "FastBroadcast"
```
ChrisRackauckas, line 46:
```suggestion
JumpProcessFastBroadcastExt = "FastBroadcast"
```

The naming for the module should follow the standard format.

test/benchmark_extended_jump_array.jl
```diff
@@ -0,0 +1,27 @@
+# This file is not directly included in a test case, but is used to
+# benchmark and compare changes to the broadcsting
+using JumpProcesses, StableRNGs, FastBroadcast, BenchmarkTools
```
ChrisRackauckas, line 3: Move it to a benchmarks folder at the top level like in https://github.com/SciML/LinearSolve.jl

ChrisRackauckas:
This looks great! Just a few minor pieces mentioned above and I think this is good to go. Comments:

> The new broadcasting method means that if you define an output array that is inhomogeneous (e.g. one where the A.u is a Vector{Float64} but A.jump_u is a Vector{Int64}), then broadcasted operations will successfully maintain the inhomogeneity. However, if you allow the broadcast machinery to create an output array for you (allocating), it will just allocate a homogenous ExtendedJumpArray with both u and jump_u being Vector{Float64}'s. This is the same behavior as the fallback, so this should not break any existing downstream code.

👍 that should be fine.

> Finally, I just used an implementation for selecting an output ExtendedJumpArray while broadcasting based on the [Julia interface documentation](https://docs.julialang.org/en/v1/manual/interfaces/#Selecting-an-appropriate-output-array); this is a recursive function that is used to unpack the broadcasted tuple in a type-stable way, but there can be a few allocations that happen prior to the broadcast if someone broadcasts a very complicated expression that exceeds a recursion depth of 16. I was trying to replace it with a concise generated function that would always be type stable, but I think it is far less likely to hit this limit than in my other recent PRs. If there is an instance where this happens, I can fix this.

This is something that FastBroadcast.jl handles, so in practice we're fine.

isaacsas:
Would be interesting to redo the VariableRate benchmarks once this is merged and see the performance vs. Coevolve now.

meson800:
>  Just a few minor pieces mentioned above and I think this is good to go

I added a commit to address those comments, thanks!

> This is something that FastBroadcast.jl handles, so in practice we're fine.

I think the problem could still exist, as [FastBroadcast also calls out to `similar`](https://github.com/YingboMa/FastBroadcast.jl/blob/master/src/FastBroadcast.jl#L27). In the `ExtendedJumpArrayStyle` method of `similar` is where `find_eja` is used (in order to locate the `ExtendedJumpArray` in the arg list to determine output axis sizes); if the argument list is too deep then the compiler might give up type inference for the recursive call, leading to some inefficient boxing/unboxing.

> Would be interesting to redo the VariableRate benchmarks once this is merged and see the performance vs. Coevolve now.

I expect that the broadcasting speedup is potentially less for short `ExtendedJumpArray`'s because they benefit less from SIMD.

ChrisRackauckas (approved):

meson800:
We'll see when the 1.6 CI finishes, but I think that this is still fine on < Julia 1.9. It won't load the extension so the FastBroadcasting will be slower, but the broadcast rules should still load and correctly select the output type, fixing the problem.

ChrisRackauckas:
> We'll see when the 1.6 CI finishes, but I think that this is still fine on < Julia 1.9. It won't load the extension so the FastBroadcasting will be slower, but the broadcast rules should still load and correctly select the output type, fixing the problem in https://github.com/SciML/JumpProcesses.jl/issues/321.

Yes, it should just fallback to being slower, which is fine for earlier versions. We need to just move forward and get the new LTS 😅

isaacsas:
@meson800 thanks for this! I suspect you've actually closed a bunch of other open `ExtendedJumpArray`-related issues with this.

PR #342, opened 2023-09-03
Refactor internal functions in spatial SSAs.

Vilin97:
This is a pure refactoring PR to make adding support for spatial constant rate jumps easier.

- Pass `integrator` around internally.
- Refactor `update_rx_rates!` in `reaction_rates.jl`.

### DIFF
--- a/src/spatial/reaction_rates.jl
+++ b/src/spatial/reaction_rates.jl
@@ -57,17 +57,8 @@ function update_rx_rates!(rx_rates::RxRates{F, M}, rxs, integrator,
     u = integrator.u
     ma_jumps = rx_rates.ma_jumps
     @inbounds for rx in rxs
-        set_rx_rate_at_site!(rx_rates, site, rx,
-                             evalrxrate((@view u[:, site]), rx, ma_jumps))
-    end
-end
-
-function update_rx_rates!(rx_rates::RxRates{F, M}, rxs, integrator,
-                          site) where {F, M <: SpatialMassActionJump}
-    u = integrator.u
-    ma_jumps = rx_rates.ma_jumps
-    @inbounds for rx in rxs
-        set_rx_rate_at_site!(rx_rates, site, rx, evalrxrate(u, rx, ma_jumps, site))
+        rate = eval_massaction_rate(u, rx, ma_jumps, site)
+        set_rx_rate_at_site!(rx_rates, site, rx, rate)
     end
 end
 
@@ -93,3 +84,6 @@ function Base.show(io::IO, ::MIME"text/plain", rx_rates::RxRates)
     num_rxs, num_sites = size(rx_rates.rates)
     println(io, "RxRates with $num_rxs reactions and $num_sites sites")
 end
+
+eval_massaction_rate(u, rx, ma_jumps::M, site) where {M <: SpatialMassActionJump} = evalrxrate(u, rx, ma_jumps, site)
+eval_massaction_rate(u, rx, ma_jumps::M, site) where {M <: MassActionJump} = evalrxrate((@view u[:, site]), rx, ma_jumps)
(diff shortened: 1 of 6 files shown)


Vilin97:
Hmm, I am confused about the build error. It tells me that some files are not formatted, so I ran the formatter in VSCode. But that did not remove the build error. How should I run the formatter to get rid of the build error? @isaacsas 

image

isaacsas:
Don't worry about the formatting. There was a change that was terrible and made a really bad formatting alteration that messes up files in many people's opinion. The proposed fix got bogged down in bikeshedding and so nothing has been implemented yet. For just ignore formatting warnings, and don't try to run the formatter yourself. (If you really feel you need to format, stick with version 1.0.32 of JuliaFormatter.)

isaacsas:
Please revert any formatter created changes please.

Vilin97:
@isaacsas , I reverted the formatting.
What is the timeline for fixing the formatting issue? It's scary to see the red x everywhere because it makes it feel like our CI is broken, which makes it feel like the code health of the package is bad.

isaacsas:
The timeline is for someone to make an executive decision on the formatting change to switch to, and then for someone to update the style at JuliaFormatter. I've pinged the thread once already suggesting we use the option that had the most support and least pushback to get this resolved, but that didn't seem to get traction, so there isn't much more I can do. 

If you look at the individual tests we aren't merging anything that fails CI. It is just the formatter action that is failing. I'd happily remove it until this is settled, but I think @ChrisRackauckas wants to keep it on all SciML libraries as policy. If it really bothers you format with the version I mentioned and the test should then pass. (Or format with that version in a separate PR, and then update this once that is merged to avoid polluting the diff here.)

src/spatial/reaction_rates.jl
```diff
@@ -52,20 +52,13 @@ end
 
 update rates of all reactions in rxs at site
 """
-function update_rx_rates!(rx_rates::RxRates{F, M}, rxs, u,
+function update_rx_rates!(rx_rates::RxRates{F, M}, rxs, integrator,
                           site) where {F, M <: MassActionJump}
```
isaacsas, line 56:
```suggestion
                          site) where {F, M <: AbstractMassActionJump}
```
otherwise this won't apply to `SpatialMassActionJump`s will it?

src/spatial/spatial_massaction_jump.jl
```diff
@@ -89,6 +89,9 @@ function SpatialMassActionJump(ma_jumps::MassActionJump{T, S, U, V}; scale_rates
                           scale_rates = scale_rates, useiszero = useiszero, nocopy = nocopy)
 end
 
+function SpatialMassActionJump(nothing)
+    SpatialMassActionJump([], [], [])
```
isaacsas, line 93: `[]` does not have a concrete type. What's the use case for this?

isaacsas (changes requested):

src/spatial/spatial_massaction_jump.jl
Vilin97: Maybe I should have kept it in the main PR. The use-case is to build RxRates with only constant rate jumps. Like [this](https://github.com/SciML/JumpProcesses.jl/pull/343/files#diff-e6f4dbdb2ddf3c6387947cbbda9235dc28658e3ab4a31933646662e1d1d25712R32). Do you have a suggestion for how to make an empty massaction jump better?

src/spatial/reaction_rates.jl
Vilin97: Yes, thank you.

src/spatial/spatial_massaction_jump.jl
isaacsas: Do something like we do here then to pick reasonable types:

https://github.com/SciML/JumpProcesses.jl/blob/421cb9b42a953ac0330d103933e25aa01008e1d3/src/aggregators/ssajump.jl#L125

isaacsas (approved):

PR #343, opened 2023-09-03
[WIP] Spatial constant rate jump 2

Vilin97:
Add spatial constant rate jump support. 
Builds on https://github.com/SciML/JumpProcesses.jl/pull/342.

### DIFF
--- a/src/spatial/reaction_rates.jl
+++ b/src/spatial/reaction_rates.jl
@@ -3,7 +3,7 @@ A file with structs and functions for sampling reactions and updating reaction r
 """
 
 ### spatial rx rates ###
-struct RxRates{F, M}
+struct RxRates{F, M, C}
     "rx_rates[i,j] is rate of reaction i at site j"
     rates::Matrix{F}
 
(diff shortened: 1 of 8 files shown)


src/spatial/reaction_rates.jl
```diff
@@ -48,16 +54,21 @@ function total_site_rx_rate(rx_rates::RxRates, site)
...
     ma_jumps = rx_rates.ma_jumps
     @inbounds for rx in rxs
-        rate = eval_massaction_rate(u, rx, ma_jumps, site)
-        set_rx_rate_at_site!(rx_rates, site, rx, rate)
+        if is_massaction(rx_rates, rx)
+            rate = eval_massaction_rate(u, rx, ma_jumps, site)
+            set_rx_rate_at_site!(rx_rates, site, rx, rate)
+        else
+            cr_jump = rx_rates.cr_jumps[rx - get_num_majumps(ma_jumps)]
+            set_rx_rate_at_site!(rx_rates, site, rx, cr_jump.rate(u, integrator.p, integrator.t, site))
```
isaacsas, line 70: This isn't type stable is it? That is why we split the ConstantRateJump rates and affects in the non-spatial solvers and wrap them inside `FunctionWrapper`s.

src/spatial/reaction_rates.jl
```diff
@@ -77,6 +88,16 @@ function sample_rx_at_site(rx_rates::RxRates, site, rng)
                   rand(rng) * total_site_rx_rate(rx_rates, site))
 end
 
+function execute_rx_at_site!(integrator, rx_rates::RxRates, rx, site)
+    if is_massaction(rx_rates, rx)
+        @inbounds executerx!((@view integrator.u[:, site]), rx,
+                             rx_rates.ma_jumps)
+    else
+        cr_jump = rx_rates.cr_jumps[rx - get_num_majumps(rx_rates.ma_jumps)]
+        cr_jump.affect!(integrator, site)
```
isaacsas, line 97: This isn't type stable is it? That is why we split the ConstantRateJump rates and affects in the non-spatial solvers and wrap them inside FunctionWrappers.

isaacsas (changes requested):
Instead of passing the site into the `affect` and `rate` functions, we could store it in the aggregator and make an interface for users to access it via the integrator (which stores the aggregator). Something like `get_current_site(integrator)`. Then this would work with old-style constant rate jump signatures. I don't know if we want that though.

The other big issue is I think there are going to be type stability issues without using FunctionWrappers on the rate and affect functions (since each function is its own type, every crj will be a different type).

PR #344, opened 2023-09-09
Bracketing unit tests

Vilin97:
The unit tests will help me extend the bracketing functionality to the spatial RSSACRDirect SSA.

### DIFF
--- a/src/aggregators/bracketing.jl
+++ b/src/aggregators/bracketing.jl
@@ -14,18 +14,18 @@ end
 BracketData{T1, T2}() where {T1, T2} = BracketData(T1(0.1), T2(25), T2(4))
 
 # support either vectors of data for each field, or scalars
-"Get fluctuation rate of species i."
+# Get fluctuation rate of species i.
 @inline getfr(bd::BracketData{AbstractVector{T1}, T2}, i) where {T1, T2} = bd.fluctrate[i]
 @inline getfr(bd::BracketData{T1, T2}, i) where {T1 <: Number, T2} = bd.fluctrate
 
-"Get threshold value of species i."
+# Get threshold value of species i.
 @inline gettv(bd::BracketData{T1, AbstractVector{T2}}, i) where {T1, T2} = bd.threshold[i]
 @inline gettv(bd::BracketData{T1, T2}, i) where {T1, T2 <: Number} = bd.threshold
 
 @inline getΔu(bd::BracketData{T1, AbstractVector{T2}}, i) where {T1, T2} = bd.Δu[i]
 @inline getΔu(bd::BracketData{T1, T2}, i) where {T1, T2 <: Number} = bd.Δu
 
-"Get brackets ((1-fluctrate)*u, (1+fluctrate)*u) for species i."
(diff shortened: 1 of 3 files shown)


src/aggregators/bracketing.jl
```diff
@@ -35,23 +38,23 @@ BracketData{T1, T2}() where {T1, T2} = BracketData(T1(0.1), T2(25), T2(4))
...
+"Get propensity brackets of constant rate jump."
 @inline function get_cjump_brackets(ulow, uhigh, rate, params, t)
     rlow = rate(ulow, params, t)
     rhigh = rate(uhigh, params, t)
-    if rlow <= rhigh
-        return rlow, rhigh
-    else
-        return rhigh, rlow
-    end
+    return min(rlow, rhigh), max(rlow, rhigh)
```
isaacsas, line 53: Doesn't this do extra work?

src/aggregators/bracketing.jl
```diff
@@ -14,15 +14,18 @@ end
 BracketData{T1, T2}() where {T1, T2} = BracketData(T1(0.1), T2(25), T2(4))
 
 # support either vectors of data for each field, or scalars
+"Get fluctuation rate of species i."
```
isaacsas, line 17: Unless you are writing these to be comments that are visible to users via the REPL, in which case they should document using the standard documentation format that also describes what the arguments are and the function signature, I'd suggest using a hash.

test/runtests.jl
```diff
@@ -2,6 +2,7 @@
 using JumpProcesses, DiffEqBase, SafeTestsets
 
 @time begin
+    @time @safetestset "Bracketing Tests" begin include("bracketing.jl") end
```
isaacsas, line 5: I'd put these lower down the list. Maybe by the composition-rejection table tests.

src/aggregators/bracketing.jl
isaacsas: Any strings you use become visible to users who type "? func_name", whereas I don't think hash-based comments do (so we can think of them as internal only documentation).

src/aggregators/bracketing.jl
isaacsas, line 53:
```suggestion
    return (rlow <= rhigh) ? rlow,rhigh : rhigh,rlow
```
if you want it as a one-liner.

src/aggregators/bracketing.jl
Vilin97: Switched to use `#`.

test/runtests.jl
Vilin97: Done. My logic was that unit tests should go first becaues they are faster than integration tests.

isaacsas: That makes sense, but we don't currently separate them (perhaps we should), so I'd prefer to keep the tests ordered by how "fundamental" the component is in some sense. If we want to consider restructuring them in the future though that would be a good consideration!

isaacsas:
LGTM. Feel free to merge when tests finished.

PR #346, opened 2023-09-11
DirectCR-RSSA

Vilin97:
A new spatial SSA called `DirectCRRSSA` uses RSSA-style sampling to sample the site+jump. It is about 20% faster than `DirectCRDirect`, at least on the problem in the `test/spatial/ABC.jl` (5 sites in a line, 3 species, 2 reactions). 

```
using BenchmarkTools
@btime solve(jump_problems[1], SSAStepper()) # NSM with a `CartesianGrid`, 52ms
@btime solve(jump_problems[2], SSAStepper()) # NSM with a `Graph`, 49ms
@btime solve(jump_problems[3], SSAStepper()) # DirectCRDirect with  a `CartesianGrid`, 51ms
@btime solve(jump_problems[4], SSAStepper()) # DirectCRRSSA with a `CartesianGrid`, 42ms
```


src/spatial/directcrdirect.jl
Vilin97, line 1: There was a very hard-to-find bug in the ordering of arguments in `generate_jumps!`. I fixed it here and in `NSM.jl`.

src/spatial/hop_rates.jl
```diff
@@ -57,30 +57,22 @@ function HopRates(p::Pair{SpecHop, SiteHop},
...
+    update_hop_rates!(hop_rates::AbstractHopRates, species_vec, u, site, spatial_system)
 
-update rates of all specs in species at site
+update rates of all species in species_vec at site
 """
-function update_hop_rates!(hop_rates::AbstractHopRates, species::AbstractArray, u, site,
-                           spatial_system)
-    @inbounds for spec in species
-        update_hop_rate!(hop_rates, spec, u, site, spatial_system)
+function update_hop_rates!(hop_rates::AbstractHopRates, species_vec, u, site, spatial_system)
```
Vilin97, line 64: If `species_vec` is a single species, the function reduces to updating a single hop rate. So I deleted `update_hop_rate!`.

src/spatial/directcrdirect.jl
Vilin97, line 1: There was a very hard-to-find bug in the ordering of arguments in generate_jumps!. I fixed it here and in NSM.jl.

src/spatial/hop_rates.jl
isaacsas: Or are you saying this is always called with a vector from higher levels so it would still have the vector overhead?

src/spatial/directcrrssa.jl
```diff
@@ -0,0 +1,265 @@
...
+    time_delta = zero(t)
+    while true
+        site = sample(rt, site_rates_high, rng)
+        jump = sample_jump_direct(rx_rates.high, hop_rates.high, site, spatial_system, rng)
+        time_delta += randexp(rng)
+        if accept_jump(p, u, jump)
+            p.next_jump_time = t + time_delta / groupsum(rt)
+            p.next_jump = jump
+            break
+        end
```
isaacsas, line 145: Does this handle if there is no next jump?

src/spatial/directcrrssa.jl
```diff
@@ -0,0 +1,265 @@
...
+end
+
+function accept_hop(p, u, jump)
+    @unpack hop_rates, spatial_system, rng = p
+    species, site = jump.jidx, jump.src
+    acceptance_threshold = rand(rng) * hop_rate(hop_rates.high, species, site)
+    if hop_rate(hop_rates.low, species, site) > acceptance_threshold
+        return true
+    else
+        # compute the real rate. Could have used hop_rates.high as well.
```
isaacsas, line 175: Does this mean we are storing two copies of the hopping constants for the low vs high (or is it aliased)?

src/spatial/directcrrssa.jl
```diff
@@ -0,0 +1,265 @@
...
+end
+
+function accept_rx(p, u, jump)
+    @unpack rx_rates, rng = p
+    rx, site = reaction_id_from_jump(p, jump), jump.src
+    acceptance_threshold = rand(rng) * rx_rate(rx_rates.high, rx, site)
+    if rx_rate(rx_rates.low, rx, site) > acceptance_threshold
+        return true
+    else
+        # compute the real rate. Could have used rx_rates.high as well.
```
isaacsas, line 188: Similar question to above

src/spatial/directcrrssa.jl
```diff
@@ -0,0 +1,265 @@
...
+        update_brackets!(p, integrator, jump.jidx, (jump.src, jump.dst))
+    else
+        update_brackets!(p, integrator, p.jumptovars_map[reaction_id_from_jump(p, jump)], jump.src)
+    end
+end
+
+function update_brackets!(p, integrator, species_to_update, sites_to_update)
+    @unpack rx_rates, hop_rates, site_rates_high, u_low_high, bracket_data, vartojumps_map, spatial_system = p
+    u = integrator.u
+    for site in sites_to_update, species in species_to_update
```
isaacsas, line 242: I always forget, but make sure this is the correct memory order to update over.

Vilin97:
@isaacsas , I merged `master` into this branch but it still shows that there are merge conflicts. What is going on?

src/spatial/bracketing.jl
```diff
@@ -5,9 +5,15 @@ struct LowHigh{T}
...
-    LowHigh(low_and_high::T) where {T} = new{T}(low_and_high, deepcopy(low_and_high))
+    function LowHigh(low::T, high::T; do_copy = true) where {T} 
+        if do_copy
+            return new{T}(deepcopy(low), deepcopy(high))
+        else
+            return new{T}(low, high)
+        end
+    end
+    LowHigh(pair::Tuple{T,T}; kwargs...) where {T} = LowHigh(pair[1], pair[2]; kwargs...)
+    LowHigh(low_and_high::T; kwargs...) where {T} = LowHigh(low_and_high, low_and_high; kwargs...)
```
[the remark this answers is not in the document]
Vilin97: Even if it's not optimized away, it shouldn't be expensive, right? I am not good at reading low level code but it seems that it does get optimized away in the end. 

```julia
# Old struct without branching
struct LowHighOld{T}
    low::T
    high::T

    function LowHighOld(low::T, high::T) where {T}
        new{T}(deepcopy(low), deepcopy(high))
    end
end

# New struct with branching
struct LowHighNew{T}
    low::T
    high::T

    function LowHighNew(low::T, high::T; do_copy=true) where {T}
        if do_copy
            return new{T}(deepcopy(low), deepcopy(high))
        else
            return new{T}(low, high)
        end
    end
end

# Test values
low_value = 10
high_value = 20

# Code introspection to see if the branch is removed in scalar cases
@code_llvm debuginfo=:none LowHighOld(low_value, high_value)

@code_llvm debuginfo=:none LowHighNew(low_value, high_value; do_copy=true)

```

The output of the code introspection is as follows:

```llvm
define void @julia_LowHighOld_2364([2 x i64]* noalias nocapture noundef nonnull sret([2 x i64]) 
align 8 dereferenceable(16) %0, {}* noundef nonnull readonly %1, i64 signext %2, i64 signext %3) #0 {
top:
  %newstruct.sroa.0.0..sroa_idx = getelementptr inbounds [2 x i64], [2 x i64]* %0, i64 0, i64 0 
  store i64 %2, i64* %newstruct.sroa.0.0..sroa_idx, align 8
  %newstruct.sroa.2.0..sroa_idx1 = getelementptr inbounds [2 x i64], [2 x i64]* %0, i64 0, i64 1  store i64 %3, i64* %newstruct.sroa.2.0..sroa_idx1, align 8
  ret void
}

define void @julia_LowHighNew_2366([2 x i64]* noalias nocapture noundef nonnull sret([2 x i64]) 
align 8 dereferenceable(16) %0, [1 x i8]* nocapture noundef nonnull readonly align 1 dereferenceable(1) %1, {}* noundef nonnull readonly %2, i64 signext %3, i64 signext %4) #0 {
top:
  %.sroa.025.0..sroa_idx = getelementptr inbounds [2 x i64], [2 x i64]* %0, i64 0, i64 0        
  store i64 %3, i64* %.sroa.025.0..sroa_idx, align 8
  %.sroa.2.0..sroa_idx26 = getelementptr inbounds [2 x i64], [2 x i64]* %0, i64 0, i64 1
  store i64 %4, i64* %.sroa.2.0..sroa_idx26, align 8
  ret void
}
```

src/spatial/directcrrssa.jl
Vilin97: To be honest, I do not rembmer any of the logic by now. Do we have a test for what you are asking? If not, we should, for all SSAs.

src/spatial/directcrrssa.jl
Vilin97: Aliased because `do_copy = false` when initializing `LowHigh(hopping_rates)`. Same with `rx_rates`.

src/spatial/directcrrssa.jl
Vilin97: Same answer -- aliased.

src/spatial/directcrrssa.jl
Vilin97: `u_low_high[species, site]` indicates that `species` should be in the inner-most loop. So this is correct.

src/spatial/hop_rates.jl
Vilin97: I don't remember my reasoning now. I put it back in.

src/spatial/hop_rates.jl
```diff
@@ -57,30 +57,22 @@ function HopRates(p::Pair{SpecHop, SiteHop},
...
+    @inbounds for species in species_vec
+        rates = hop_rates.rates
+        old_rate = rates[species, site]
+        rates[species, site] = evalhoprate(hop_rates, u, species, site,
+                                                    spatial_system)
+        hop_rates.sum_rates[site] += rates[species, site] - old_rate
+        old_rate
```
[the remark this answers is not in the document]
Vilin97: The output is not new. I don't know if it's ever used in practice. But it's not a new addition.

Vilin97 (commented):
Addressed comments.

Vilin97:
What do I need to do to get rid of the format and the spell check CI failures?

isaacsas:
Spell check fix the suggested spelling errors I guess. Haven’t ever used it or seen it before. 

Format use JuliaFormatter to format the repo and add to the PR.

isaacsas:
Just FYI, I'm traveling for the next week, but will try to review once I am back in Boston.

isaacsas:
@Vilin97 sorry again for the long delays with this. I’ll try to put a test for the case I was worried about into the general tests in a followup. Assuming tests now pass I’ll merge. 

I do think there is a chance the removal of the dispatch for HighLow could have performance implications, as I’m not sure Julia can always remove the branch in this situation (what it can do from the REPL may not be the same as what it can do here). But in any case we can merge and that can be examined later as a performance optimization. 

Thanks again for your efforts on this — I look forward to updating the benchmarks to use it to see how it does.

isaacsas:
Ahh, I didn’t realize that the updates to address my earlier comments actually never passed tests.

PR #348, opened 2023-09-12
Improve spatial tests

Vilin97:
Add stable RNGs to the spatial integration tests and reduce number of simulations. This reduces the runtime of spatial tests by over 50%.


isaacsas (commented):
Have you checked these are enough tests to avoid occasional failures from undersampling? Those can be annoying to deal with / understand.

isaacsas:
Nevermind, with StableRNGs and a set seed that obviously shouldn’t happen.

isaacsas (approved):

PR #351, opened 2023-09-22
Choose an SSA if no SSA is passed in `JumpProblem`.

Vilin97:
If `JumpProblem(prob, jumps)` is called, an SSA is selected via the function below. The assumption is that `Direct` is fastest for a small network, while `RSSACR` is fastest for other networks. If `RSSACR` cannot be used because no species-to-jumps graph is given, then `DirectCR` is used instead.

```
# return the fastest aggregator out of the available ones
function select_aggregator(jumps::JumpSet; vartojumps_map=nothing, jumptovars_map=nothing, dep_graph=nothing, spatial_system=nothing, hopping_constants=nothing)
    
    # detect if a spatial SSA should be used
    !isnothing(spatial_system) && !isnothing(hopping_constants) && return DirectCRDirect

    # if variable rate jumps are present, return the only SSA that supports them
    num_vrjs(jumps) != 0 && return Coevolve
    
    # if the number of jumps is small, return the Direct
    num_majumps(jumps) + num_crjs(jumps) < 10 && return Direct

    # if there are only massaction jumps, we can any build dependency graph, so return the fastest aggregator
    num_crjs(jumps) == 0 && num_vrjs(jumps) == 0 && return RSSACR

    # in the presence of constant rate jumps, we rely on the given dependency graphs
    if !isnothing(vartojumps_map) && !isnothing(jumptovars_map)
        return RSSACR
    elseif !isnothing(dep_graph)
        return DirectCR
    else
        return Direct
    end
end
```


test/spatial/ABC.jl
```diff
@@ -56,6 +56,9 @@ jump_problems = JumpProblem[JumpProblem(prob, NSM(), majumps,
 push!(jump_problems,
       JumpProblem(prob, DirectCRDirect(), majumps, hopping_constants = hopping_constants,
                   spatial_system = grids[1], save_positions = (false, false), rng = rng))
+push!(jump_problems,
+JumpProblem(prob, majumps, hopping_constants = hopping_constants,
```
isaacsas, line 60:
```suggestion
      JumpProblem(prob, majumps, hopping_constants = hopping_constants,
```

src/problem.jl
```diff
@@ -171,9 +171,12 @@ function JumpProblem(prob, aggregator::AbstractAggregatorAlgorithm, jumps::Abstr
                      kwargs...)
     JumpProblem(prob, aggregator, JumpSet(jumps...); kwargs...)
 end
-function JumpProblem(prob, jumps::JumpSet; kwargs...)
-    JumpProblem(prob, NullAggregator(), jumps; kwargs...)
+function JumpProblem(prob, jumps::JumpSet; vartojumps_map=nothing, jumptovars_map=nothing, dep_graph=nothing, spatial_system=nothing, hopping_constants=nothing, kwargs...)
```
isaacsas, line 174: Don't you need versions for all the jump types too? i.e. like
```function JumpProblem(prob, jumps::ConstantRateJump; kwargs...)```
and
```function JumpProblem(prob, jumps::AbstractJump...; kwargs...)```

Comment (isaacsas) [reply src/problem.jl:174]: See the dispatches above your new one.

Comment (isaacsas) [reply src/problem.jl:174]: These should probably get associated tests too.

File: src/aggregators/aggregators.jl [hunk 3]
```diff
@@ -184,6 +184,32 @@ needs_vartojumps_map(aggregator::RSSACR) = true
...
 is_spatial(aggregator::DirectCRDirect) = true
+
+# return the fastest aggregator out of the available ones
+function select_aggregator(jumps::JumpSet; vartojumps_map=nothing, jumptovars_map=nothing, dep_graph=nothing, spatial_system=nothing, hopping_constants=nothing)
+    
+    # detect if a spatial SSA should be used
+    !isnothing(spatial_system) && !isnothing(hopping_constants) && return DirectCRDirect
+
+    # if variable rate jumps are present, return the only SSA that supports them
+    num_vrjs(jumps) != 0 && return Coevolve
```
isaacsas, line 199: We should only pick `Coevolve` if `any(isbounded, vrjs)`, otherwise we should probably select `Direct`.

src/aggregators/aggregators.jl
```diff
@@ -184,6 +184,32 @@ needs_vartojumps_map(aggregator::RSSACR) = true
...
+    # if there are only massaction jumps, we can any build dependency graph, so return the fastest aggregator
+    num_crjs(jumps) == 0 && num_vrjs(jumps) == 0 && return RSSACR
+
+    # in the presence of constant rate jumps, we rely on the given dependency graphs
+    if !isnothing(vartojumps_map) && !isnothing(jumptovars_map)
+        return RSSACR
+    elseif !isnothing(dep_graph)
+        return DirectCR
+    else
+        return Direct
```
isaacsas, line 213: I think we are missing a middle regime where `SortingDirect` or `RSSA` might be better. Have you played around with this at all / looked over the SciMLBenchmarks and the Catalyst paper results?

isaacsas (changes requested):

src/aggregators/aggregators.jl
TorkelE: When I did the Catalyst tests, `RSSA` was best for small systems. However, we didn't test very small systems, so not sure about these. But there definitely was a regime where `RSSA` as the best choice.

Vilin97: image
This image suggests that RSSA is indeed better on small models but only marginally. I did not see a comparison with Direct and SortingDirect there. @TorkelE , do you have a table of times from when you did the benchmarking? I did not find such a table for SSA benchmarks in the paper.

Vilin97: image
image
These two are from models with 6 and 18 reactions, respectively. Indeed, it looks like between ~6 and ~66 reactions we might want to use RSSA. The number 66 comes from the benchmark from the Catalyst paper.

TorkelE: Think the smallest model (Multistate) is 18 reactions. It is kind of hard to see the Direct and Sorting DIrect as RSSA (and RSSACR) is more or less on top of them (also suggesting that the choice of method is not that important).

There is not good table, but the results are in JSON files here: https://github.com/SciML/Catalyst_PLOS_COMPBIO_2023/tree/master/Benchmarks/Benchmarking_results/Threads_1

Look for e.g. RSSA_multistate and you find the multistate results. The data is saved in `Dict`s with the benchmark median times and simulation lengs as separate entries.

ChrisRackauckas: Does that hold with static array states?

ChrisRackauckas:
Perfect is always the enemy of good for this kind of thing. I can tell you that 90% of users just use Gillespie's method if we document `Direct` as Gillespie's method. Even the simplest defaulting scheme can be merged and do better than the vast majority of users, so the details can always evolve but I think the biggest thing is to just get something with the right broad strokes and also update all of the tutorials to not hardcode a method. The latter is BTW very important, because there's still people who use DiffEq who just throw Tsit5 everywhere instead of using the default, and you really want to stop that habit via tutorials otherwise people will overuse Gillespie.

src/aggregators/aggregators.jl
Vilin97: What is `isbounded`?
I did make the change.

src/problem.jl
Vilin97: The dispatches on lines 138-152 above already exist and have the signature you specified. Why they already exist is unclear to me but I did not have to add them.

src/aggregators/aggregators.jl
Vilin97: I added `RSSA` if the number of jumps is between 10 and 100. This schema is bound to be just a heuristic so I don't think we should spend a ton of time finding the exact threshold when e.g. RSSA is better than e.g. SortingDirect.

src/aggregators/aggregators.jl
```diff
@@ -184,6 +184,38 @@ needs_vartojumps_map(aggregator::RSSACR) = true
...
+    # if variable rate jumps are present, return one of the two SSAs that support them
+    if num_vrjs(jumps) != 0
+        any(isbounded, vrjs) && return Coevolve
+        return Direct
+    end
+    
+    # if the number of jumps is small, return the Direct
+    num_jumps(jumps) < 10 && return Direct
+
+    # if there are only massaction jumps, we can any build dependency graph
```
TorkelE, line 207: "we can any build dependency graph" 
Sounds like a weird sentence structure?

src/aggregators/aggregators.jl
```diff
@@ -184,6 +184,38 @@ needs_vartojumps_map(aggregator::RSSACR) = true
...
+    can_build_dependency_graphs = num_crjs(jumps) == 0 && num_vrjs(jumps) == 0
+    have_dependency_graphs = !isnothing(vartojumps_map) && !isnothing(jumptovars_map)
+
+    # if we have the species-jumps dependency graphs or can build them, use one of the Rejection-based methods
+    if can_build_dependency_graphs || have_dependency_graphs
+        num_jumps(jumps) < 100 && return RSSA
+        return RSSACR
+    # if we have the jumps-jumps dependency graph, use the Composition-Rejection Direct method
+    elseif !isnothing(dep_graph)
```
TorkelE, line 216: I don't really understand the difference between `!isnothing(dep_graph)` and `have_dependency_graphs`, so not sure what is going on here (not asking to change anything, just pointing out that I would not be able to tell fi something should be changed).

test/geneexpr_test.jl
```diff
@@ -12,7 +12,7 @@ dotestmean = true
 doprintmeans = false
 
 # SSAs to test
-SSAalgs = (RDirect(), RSSACR(), Direct(), DirectFW(), FRM(), FRMFW(), SortingDirect(),
+SSAalgs = (JumpProcesses.NullAggregator(), RDirect(), RSSACR(), Direct(), DirectFW(), FRM(), FRMFW(), SortingDirect(),
```
TorkelE, line 15: Would it make sense to, for completeness, have a test calling `JumpProblem(prob, jumpset)`, and not always going via `JumpProcesses.NullAggregator()`? Like, it totally should be fine, but this is probably how most users will en up calling `JumpProblem`s, so it might be prudent to have that version in the tests as well.

TorkelE (commented):
Looks good to me.

src/aggregators/aggregators.jl
Vilin97: I rewrote it as "we can any build the species-jumps dependency graphs".

src/aggregators/aggregators.jl
Vilin97: I reworded it. `dep_graph` is the jumps-jumps dependency graph, as opposed to species-jumps and jumps-species.

src/aggregators/aggregators.jl
TorkelE: I still think "..., we can any build ..." sounds like strange English. I'm not a native speaker though (and it is only a comment), so no need to dwell on it.

src/aggregators/aggregators.jl
TorkelE: Got it, yes, it makes more sense now.

test/geneexpr_test.jl
Vilin97: Added at the end of this file.

src/aggregators/aggregators.jl
Vilin97: That was a genuine typo. I fixed it now.

isaacsas (approved):

isaacsas:
@Vilin97 will merge when I get tests passing.

@TorkelE expanded this a bit to allow not needing to pass `SSAStepper` too, i.e. with this PR this should now work
```julia
jprob = JumpProblem(rn, dprob)
sol = solve(jprob)
```

isaacsas:
@TorkelE actually MTK and Catalyst need appropriate dispatches for JumpProblem to make this work, so not quite there yet for being usable in Catalyst.

src/solve.jl
```diff
@@ -7,6 +7,11 @@ function DiffEqBase.__solve(jump_prob::DiffEqBase.AbstractJumpProblem{P},
     integrator.sol
 end
 
+# if passed a JumpProblem and no aggregator is selected use SSAStepper
+function DiffEqBase.__solve(jump_prob::DiffEqBase.AbstractJumpProblem; kwargs...)
+    DiffEqBase.__solve(jump_prob, SSAStepper(); kwargs...)
+end
```
isaacsas, line 13: @ChrisRackauckas can you confirm this is an OK dispatch for me to add to enable 
```julia
sol = solve(jprob; kwargs...)  # equivalent to solve(jprob, SSAStepper(); kwargs...)
```

isaacsas: Hmm, I guess this ignores when the JumpProblem is over an ODEProblem or such. Is there a way I can explicitly select the default auto-solver for a given problem type?

isaacsas:
@ChrisRackauckas if you could give a quick look to ensure I haven't added a dispatch better handled elsewhere (or shadowing one elsewhere) that would be helpful.

HISTORY.md
```diff
@@ -2,17 +2,45 @@
...
+  sol = solve(jprob, SSAStepper())
+  ```
+- For `JumpProblem`s over `DiscreteProblem`s that only have `MassActionJump`s,
+  `ConstantRateJump`s, and bounded `VariableRateJump`s, one no longer needs to
+  specify `SSAStepper()` when calling `solve`, i.e. the following now works for
+  the previous example and is equivalent to manually passing `SSAStepper()`:
+  ```julia
+  sol = solve(jprob)
+  ```
+- Plotting a solution generated with `save_positions = (false, false)` no uses
```
isaacsas, line 25:
```suggestion
- Plotting a solution generated with `save_positions = (false, false)` now uses
```

src/aggregators/aggregators.jl
```diff
@@ -188,6 +188,40 @@ needs_vartojumps_map(aggregator::RSSACR) = true
...
+function select_aggregator(jumps::JumpSet; vartojumps_map = nothing,
+        jumptovars_map = nothing, dep_graph = nothing, spatial_system = nothing,
+        hopping_constants = nothing)
+
+    # detect if a spatial SSA should be used
+    !isnothing(spatial_system) && !isnothing(hopping_constants) && return DirectCRDirect
+
+    # if variable rate jumps are present, return one of the two SSAs that support them
+    if num_vrjs(jumps) > 0
+        (num_bndvrjs(jumps) > 0) && return Coevolve
```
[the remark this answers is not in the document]
isaacsas: In theory we support mixing bounded and non-bounded vrjs (but if there are any non-bounded vrjs you need to use a JumpProblem over an ODEProblem or such). JumpProblem has code to check these cases, but I'm not sure how well this is actually tested currently.

isaacsas:
@Vilin97 thanks for your work on this and sorry for the long long delay!

Vilin97:
Thank you for cleaning up and adding a default solver too. Now we need to update the docs, right?

isaacsas:
Yes, the docs need updates in places now. Using the defaults in the most introductory places, but explaining how to select solvers and such as they currently do in more advanced places.

PR #352, opened 2023-09-23
Documenter 1.0 upgrade

ArnoStrouwen:


ArnoStrouwen:
Is this package going to keep using an older formatting style? Or has it just not been updated.

ChrisRackauckas:
update it later

isaacsas:
Why did this get merged? SciMLStyle is still not fixed to my knowledge, and this now breaks the formatting here even worse by formatting with the broken style version?

Please revert this and open a clean PR with just the doc updates.

isaacsas:
A mass restyle also shouldn’t be included as part of anything but a separate PR.

src/spatial/spatial_massaction_jump.jl
```diff
@@ -44,89 +44,113 @@ end
...
-                          ma_jumps.net_stoch, ma_jumps.param_mapper;
-                          scale_rates = scale_rates, useiszero = useiszero, nocopy = nocopy)
+        ma_jumps.net_stoch, ma_jumps.param_mapper;
+        scale_rates = scale_rates, useiszero = useiszero, nocopy = nocopy)
 end
 
 ##############################################
 
-function get_num_majumps(smaj::SpatialMassActionJump{Nothing, Nothing, S, U, V}) where
+function get_num_majumps(smaj::SpatialMassActionJump{
```
isaacsas, line 94: @ChrisRackauckas this kind of reformatting is exactly what we discussed in that thread, which break VSCode functionality, and has readability issues. It is why we haven't reformatted while awaiting the fix to that thread, since it seems like this is a non-reversible format change (i.e. the formatter itself won't fix this in the future).

ChrisRackauckas:
Sorry I thought this was just another one of the Documenter v1.0 upgrades. I didn't note the formatter changes slipped in.

isaacsas:
Can we revert this then and just put through the doc changes as a small PR without format changes?

ChrisRackauckas:
yes

PR #354, opened 2023-09-23
Update FormatCheck.yml

ArnoStrouwen:


isaacsas:
Why is this change being made? https://github.com/domluna/JuliaFormatter.jl/issues/741 Is still open and a new style has not been made available to fix the broken formatting for which we had to introduce this cap?

PR #362, opened 2023-10-29
format attempt using JuliaFormatter master

isaacsas:


docs/src/faq.md
```diff
@@ -62,7 +62,7 @@ argument. Continuing the previous example:
 #] add RandomNumbers
 using RandomNumbers
 jprob = JumpProblem(dprob, Direct(), maj,
-                    rng = Xorshifts.Xoroshiro128Star(rand(UInt64)))
+    rng = Xorshifts.Xoroshiro128Star(rand(UInt64)))
```
isaacsas, line 65: @ChrisRackauckas @YingboMa is this really what the style is supposed to do? This reformat to a single tab spacing makes it look like an incorrectly indented independent line instead of a function argument...

docs/src/index.md
```diff
@@ -107,9 +107,9 @@ projtoml = joinpath("..", "..", "Project.toml")
 version = TOML.parse(read(projtoml, String))["version"]
 name = TOML.parse(read(projtoml, String))["name"]
 link_manifest = "https://github.com/SciML/" * name * ".jl/tree/gh-pages/v" * version *
-       "/assets/Manifest.toml"
+                "/assets/Manifest.toml"
```
isaacsas, line 110: This is what I would have thought/wanted the function argument above to get aligned as... But if the above is desirable then shouldn't this get indented just one tab spacing too?

src/aggregators/coevolve.jl
```diff
@@ -84,8 +84,13 @@ function (p::CoevolveJumpAggregation)(integrator::I) where {I <:
...
 
-function (p::CoevolveJumpAggregation{T, S, F1, F2})(integrator::AbstractSSAIntegrator) where
-    {T, S, F1, F2 <: Union{Tuple, Nothing}}
+function (p::CoevolveJumpAggregation{
+        T,
+        S,
+        F1,
+        F2,
+    })(integrator::AbstractSSAIntegrator) where
+        {T, S, F1, F2 <: Union{Tuple, Nothing}}
```
isaacsas, line 93: @ChrisRackauckas @YingboMa this still seems like the expanding things unnecessarily to one per line bug is around... This is adding a lot of unnecessary lines with all the new line breaks for something that can fit on one line within the col width (i.e. the parametric types).

src/aggregators/ssajump.jl
```diff
@@ -88,8 +88,13 @@ function (p::AbstractSSAJumpAggregator)(integrator::I) where {I <: DiffEqBase.DE
...
 
-function (p::AbstractSSAJumpAggregator{T, S, F1, F2})(integrator::DiffEqBase.DEIntegrator) where
-    {T, S, F1, F2 <: Union{Tuple, Nothing}}
+function (p::AbstractSSAJumpAggregator{
+        T,
+        S,
+        F1,
+        F2,
+    })(integrator::DiffEqBase.DEIntegrator) where
+        {T, S, F1, F2 <: Union{Tuple, Nothing}}
```
isaacsas, line 97: Same issue.

src/spatial/directcrdirect.jl
```diff
@@ -137,7 +137,9 @@ end
 
 reset all stucts, reevaluate all rates, repopulate the priority table
 """
-function fill_rates_and_get_times!(aggregation::DirectCRDirectJumpAggregation, integrator, t)
+function fill_rates_and_get_times!(aggregation::DirectCRDirectJumpAggregation,
+        integrator,
+        t)
     @unpack spatial_system, rx_rates, hop_rates, site_rates, rt = aggregation
```
isaacsas, line 143: Another case of expanding to more lines than needed for function arguments.

src/spatial/spatial_massaction_jump.jl
```diff
@@ -44,89 +44,113 @@ end
...
-function get_num_majumps(smaj::SpatialMassActionJump{Nothing, Nothing, S, U, V}) where
-    {S, U, V}
+function get_num_majumps(smaj::SpatialMassActionJump{
+        Nothing,
+        Nothing,
+        S,
+        U,
+        V,
+    }) where
+        {S, U, V}
```
isaacsas, line 101: multline format bug again.

src/spatial/spatial_massaction_jump.jl
```diff
@@ -44,89 +44,113 @@ end
...
-function get_num_majumps(smaj::SpatialMassActionJump{Nothing, B, S, U, V}) where
-    {B, S, U, V}
+function get_num_majumps(smaj::SpatialMassActionJump{
+        Nothing,
+        B,
+        S,
+        U,
+        V,
+    }) where
+        {B, S, U, V}
```
isaacsas, line 111: multline format bug again.

src/spatial/spatial_massaction_jump.jl
```diff
@@ -44,89 +44,113 @@ end
...
-function get_num_majumps(smaj::SpatialMassActionJump{A, Nothing, S, U, V}) where
-    {A, S, U, V}
+function get_num_majumps(smaj::SpatialMassActionJump{
+        A,
+        Nothing,
+        S,
+        U,
+        V,
+    }) where
+        {A, S, U, V}
```
isaacsas, line 121: multline format bug again.

src/spatial/spatial_massaction_jump.jl
```diff
@@ -44,89 +44,113 @@ end
...
-function get_num_majumps(smaj::SpatialMassActionJump{A, B, S, U, V}) where
-    {A <: AbstractVector, B <: AbstractMatrix, S, U, V}
+function get_num_majumps(smaj::SpatialMassActionJump{
+        A,
+        B,
+        S,
+        U,
+        V,
+    }) where
+        {A <: AbstractVector, B <: AbstractMatrix, S, U, V}
```
isaacsas, line 131: multline format bug again.

isaacsas (commented):
@ChrisRackauckas @YingboMa it doesn't appear to me that the bug that was reported a while ago about expanding multiple function arguments or type parameters to one per line instead of maintaining compactness has been fixed. See the examples I've highlighted.

src/aggregators/coevolve.jl
ChrisRackauckas: Yeah this looks incorrect indeed.

domluna: the bug here is `})` should be 4 spaces dedented.

the line expansion is due to the margin ruling and so it'll break the line based on the nesting style which is delegating to either Default and YAS styles in SciML depending on whether yas_style_nesting is true or false. Blue style tries the args on a line indented before splitting them up so perhaps replacing default style with blue style for nesting would be what you guys want. otherwise you can increase the margin.

docs/src/faq.md
domluna: if yas_style_nesting is set to true it'll indent to the function argument

src/aggregators/coevolve.jl
ChrisRackauckas: Defaulting to blue style's approach there sounds like a good approach.

YingboMa: Doesn't blue style fully expand into 1-arg per line if all args don't fit in one line? I'd much prefer to not do that and try to minimize newly introduced lines by splitting.

domluna: yes it does. 

https://github.com/domluna/JuliaFormatter.jl/blob/master/src/styles/blue/nest.jl#L1-L133

you could alter that to try a incremental line splitting thing. Try 1 line, if not 1, then 2, if not 2, then 3, etc.

in what situations do you want line splitting though? you can set the margin to be high if it's a global want, or alter some nesting functions to not split.

domluna: fix for the bug here https://github.com/domluna/JuliaFormatter.jl/pull/791

isaacsas: Using JuliaFormatter v1.0.45 I'm still getting a similar reformat to above:

Screenshot 2023-12-12 at 8 25 52 AM

domluna:
the fix was for the extra bracket indent

src/aggregators/coevolve.jl
domluna: my email didn't send but the fix was for the extra indent

Issue #365, opened 2023-11-02
Remarks about docs [JuliaCon Proceedings review]


gdalle:
Hey there!
I'm reviewing the package and associated paper for the JuliaCon Proceedings, so I'll put some remarks about the documentation here. This issue will be updated as I make my way through the docs.

https://github.com/JuliaCon/proceedings-review/issues/133

**Home page**

- I think a few sentences might be warranted about what a jump / point process is and to whom it is useful ("statement of need" section of the review). I know people landing on this page will probably know already, but it doesn't cost much to make it accessible for people who don't.

**Simple Poisson Processes in JumpProcesses**

- I don't understand the specific roles of the aggregator and the stepper
- For the process with varying birth rate, you state that "death does not change `u[1]`", which seems weird to me. Isn't it rather birth that doesn't change `u[2]`?

**Continuous-Time Jump Processes and Gillespie Methods**

- Is there a way to implement the variable rate case of time-dependent infection rate without a global variable `H`? Same question for other models on the page
- When it comes to controlling saving behavior, what exactly is saved by the callback mechanism? During the first read it was unclear to me that the saving referred to the solution values.

**Piecewise Deterministic Markov Processes and Jump Diffusion Equations**

- The jump diffusion equation may be a bit unclear for those who (like me) don't know SDEs. I assume $dW$ is a Brownian motion?
- Why does the SDE version look so different from the PDMPs on the last plot?

**Spatial SSAs with JumpProcesses.jl**

(nothing)

**Mathematical Specification of a problem with jumps**

- I would put the blurb at the beginning much earlier in the docs, perhaps even before the tutorials
- An example of `DiscreteProblem` definition would be welcome before stating that `JumpProblem` must wrap it

**Jump solvers**

- The distinct roles of the aggregator and stepper are still unclear to me

**Small typos**

- https://github.com/SciML/JumpProcesses.jl/pull/366

gdalle:
https://github.com/SciML/JumpProcesses.jl/pull/366

ChrisRackauckas:
> I don't understand the specific roles of the aggregator and the stepper

Everyone likes to talk about the jumps as being a stochastic process. They are not. The stochastic process is only defined by knowing all of the jumps together and choosing some method to aggregate them. The aggregator generates the representation of the stochastic process from the jump descriptions. But then you have to say how you're going to step: do you step directly to the jumps? Constant dt? Etc. This matters more in the context of connecting to ODEs.

[...]

Issue #370, opened 2023-11-02
Remarks about package [JuliaCon Proceedings review]


gdalle:
You know the drill by now. Here are a few general questions I had while reading the paper and docs:

https://github.com/JuliaCon/proceedings-review/issues/133

**Functionalities**

- The main reason I wrote [PointProcesses.jl](https://github.com/gdalle/PointProcesses.jl) was to work with processes that generate arbitrary marks (not just multivariate counting processes). I think it is possible to do that with JumpProcesses.jl but I'd like some pointers
- Other interesting aspects of my library include loglikelihood computation and fitting. I think fitting is not in your scope, but is there any chance you would consider implementing a loglikelihood computation given a process and a sample?

isaacsas:
Couldn't mark sampling be encoded in the `affect!` function? You can always setup your parameter object to have a place to store the marks (if you don't want to store them in the solution `u`). We should probably add a doc example showing how to do this (and how to get at the `rng` we are using so user samplers can also use it).

We don't have any explicit higher level fitting functionality here, but we should compose with higher level packages in SciML and such for fitting (modulo that AD can't really be used through the aggregation methods -- but this is something we are working on). We could certainly consider adding such computations here if it makes sense. Do you have any specific references for the case you are interested in?

ChrisRackauckas:
> Couldn't mark sampling be encoded in the affect! function? You can always setup your parameter object to have a place to store the marks (if you don't want to store them in the solution u). We should probably add a doc example showing how to do this (and how to get at the rng we are using so user samplers can also use it).

Yes, the marks are just handled in the `affect!` functions. The only thing that needs specialization there are RegularJumps, and I believe there's stuff written on that aspect already?

> Other interesting aspects of my library include loglikelihood computation and fitting. I think fitting is not in your scope, but is there any chance you would consider implementing a loglikelihood computation given a process and a sample?

The general loglikelihoods can only be defined through sampling. Indeed the simple cases can have analytical solution and it can make sense to write that down, but for as they become general there's no analytical solution and sampling 10,000 times is the way to estimate the likelihood.

isaacsas:
Yes, RegularJumps should support marks with the newer interface (I think the person who added that didn't add any docs unfortunately).

gzagatti:
Since a marked point process is just a ground point process with a mark distribution. It should be pretty straightforward to implement marked point processes by drawing from an arbitrary distribution inside of the `affect!` function. 

[...]

PR #372, opened 2023-11-21
Update JumpProcesses documentation

gzagatti:
Following @gdalle review of our JuliaCon paper (thank you very much!), I have modified the documentation to address all his concerns.

The goal was to make the documentation more accessible to beginners and those coming from point process theory. I have added two new sections "Getting Started with JumpProcesses in Julia" and "Temporal Point Processes (TPP) with JumpProcesses".

Please see my comments to all the points raised by the reviewer.

From Issue #365.

> **Home page**
> 
>     * I think a few sentences might be warranted about what a jump / point process is and to whom it is useful ("statement of need" section of the review). I know people landing on this page will probably know already, but it doesn't cost much to make it accessible for people who don't.

I added a paragraph to situate a new user. I have updated the list of links and split it into two lists, one for tutorials and one for references and APIs.

The front page also guides the user to a new page "Getting Started with JumpProcesses in Julia" to align with the DifferentialEquations documentation which also contains a similar page.

I have also aligned the sidecar with the DifferentialEquations documentation to ensure that a user familiar with that library documentation will also be familiar with JumpProcesses documentation.

> 
> 
> **Simple Poisson Processes in JumpProcesses**
> 
>     * I don't understand the specific roles of the aggregator and the stepper

I have rephrased the paragraph where we introduce the aggregator to provide more intuition on this type of object. Also, see "Getting Started with JumpProcesses in Julia" for more accessible discussion on aggregators and steppers.

> 
>     * For the process with varying birth rate, you state that "death does not change `u[1]`", which seems weird to me. Isn't it rather birth that doesn't change `u[2]`?

This paragraph was indeed very confusing. I have re-written it to make it more clear. The main point is that birth and death processes are orthogonal to the state variable `u`. The death process does not affect the birth process, because the later only obeys seasonal fluctuation. In other words, it is not because a lot of people are dying that agents will change their fertility behavior.

> 
> 
> **Continuous-Time Jump Processes and Gillespie Methods**
> 
>     * Is there a way to implement the variable rate case of time-dependent infection rate without a global variable `H`? Same question for other models on the page

I have removed the dependency on the global variable. `H` is now part of the parameters `p` that are used to initialize the problem.

> 
>     * When it comes to controlling saving behavior, what exactly is saved by the callback mechanism? During the first read it was unclear to me that the saving referred to the solution values.

I have added some notes to clarify that we are only saving the state variable `u`  at end of section "Reducing Memory Use: Controlling Saving Behavior"

> 
> 
> **Piecewise Deterministic Markov Processes and Jump Diffusion Equations**
> 
>     * The jump diffusion equation may be a bit unclear for those who (like me) don't know SDEs. I assume dW is a Brownian motion?

I have added a beginner-level discussion below the equation.

> 
>     * Why does the SDE version look so different from the PDMPs on the last plot?

This is just a coincidence, since all paths are stochastic they might actually look very different from simulation to simulation. In any case, I have added a discussion below the plots to explain what the different problem definitions bring to the simulated paths. 

I have also reduced the complexity of the problems by removing the copied jump. It doesn't really add much to the overall goal of the tutorial and might be confusing. The main goal of the tutorial is to show how JumpProcesses can be used with discrete problems, ODEs, or SDEs.

> 
> 
> **Spatial SSAs with JumpProcesses.jl**
> 
> (nothing)
> 
> **Mathematical Specification of a problem with jumps**
> 
>     * I would put the blurb at the beginning much earlier in the docs, perhaps even before the tutorials

I have added a "Getting Started with JumpProcesses in Julia" to align with the DifferentialEquations documentation which also contains a similar page. A concise tutorial in the same style as the parent library should make JumpProcesses more accessible.

>
>     * An example of `DiscreteProblem` definition would be welcome before stating that `JumpProblem` must wrap it
> 

That is now discussed in "Getting Started with JumpProcesses".

> 
> **Jump solvers**
> 
>     * The distinct roles of the aggregator and stepper are still unclear to me
> 

The distinction should be made clear and accessible in "Getting Started with JumpProcesses". 

> 
> **Small typos**
> 
>     * [Fix typos in docs #366](https://github.com/SciML/JumpProcesses.jl/pull/366)

Thanks!

From Issue #370.

> **Functionalities**
> 
>     * The main reason I wrote [PointProcesses.jl](https://github.com/gdalle/PointProcesses.jl) was to work with processes that generate arbitrary marks (not just multivariate counting processes). I think it is possible to do that with JumpProcesses.jl but I'd like some pointers

I have added a complete new tutorial on temporal point processes (TPP) with JumpProcesses. It guides someone familiar with TPP theory on how to use JumpProcesses to simulate TPPs.

As an added bonus, I am using PointProcess to define `SciMLPointProcess` which inherits `AbstractPointProcess` whose API contained many TPP constructs.

The tutorial builds `SciMLPointProcess` while apply it to a concrete case the marked Hawkes process.

> 
>     * Other interesting aspects of my library include loglikelihood computation and fitting. I think fitting is not in your scope, but is there any chance you would consider implementing a loglikelihood computation given a process and a sample?

I have implemented all the API from PointProcess except for fitting. 

Regarding TPP parameter fitting, we cannot always derive an analytical estimator via maximum likelihood for TPPs. Therefore, it is beyond the scope of the tutorial to implement such method. I have added a discussion of this fact at the end of the tutorial.

Fix issues #365, #370 and JuliaCon/proceedings-review#133.


isaacsas:
@gzagatti thanks! I will give this a look through so we can hopefully get it merged very soon. (Hopefully tonight or tomorrow.)

docs/make.jl
```diff
@@ -27,5 +27,5 @@ makedocs(sitename = "JumpProcesses.jl",
                                   mathengine),
          pages = pages)
 
-deploydocs(repo = "github.com/SciML/JumpProcesses.jl.git";
```
isaacsas, line 30: Why did you change this?

isaacsas: Now we can't preview the build on the PR.

gzagatti: I shouldn't have committed this. I just commented out while developing it on my machine.

docs/Project.toml
```diff
@@ -1,11 +1,16 @@
...
 DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+GR_jll = "d2c73de3-f751-5644-a686-071e5b155ba9"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 JumpProcesses = "ccbc3e58-028d-4f4c-8cd5-9ae44345cda5"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+PointProcesses = "af0b7596-9bb0-472a-a012-63904f2b4c55"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
```
isaacsas, line 13: These all need compat entries.

docs/src/assets/Project.toml
```diff
@@ -1,11 +1,16 @@
...
 DifferentialEquations = "0c46a032-eb83-5123-abaf-570d42b7fbaa"
+Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
+GR_jll = "d2c73de3-f751-5644-a686-071e5b155ba9"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 JumpProcesses = "ccbc3e58-028d-4f4c-8cd5-9ae44345cda5"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+PointProcesses = "af0b7596-9bb0-472a-a012-63904f2b4c55"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
```
isaacsas, line 13: Needs compat entries.

docs/src/getting_started.md
```diff
@@ -0,0 +1,189 @@
...
+share many things in common which make JumpProcesses ideal for both.
+
+Processes involving multiple jumps are known as compound jump (or point)
+processes.
+
+```math
+du = \sum_{j} h_j(u,p,t) dN_j(t)
+```
+
+where ``N_j`` is a jump process with rate ``\lambda_i(u,p,t)``.
```
isaacsas, line 18: This is too vague. You need to fully explain the equation and the meaning of all the terms in it if you are putting this forward as the new opening introduction to the package. What is the "jump process" $N_j$, what is $h_j$? What is a "rate"?

docs/src/getting_started.md
```diff
@@ -0,0 +1,189 @@
...
+
+The homogeneous Poisson process is the canonical point process with a constant
+rate of change. A compound Poisson process is a continuous-time Markov Chain
+where the time to the next jump is exponentially distributed as determined by
+the aggregate rate. In the statistics literature, the composition of Poisson
+processes is described by the superposition theorem. Simulation algorithms for
+these types of processes are known in biology and chemistry as Gillespie methods
+or Stochastic Simulation Algorithms (SSA), with the time evolution that the
+probability these processes are in a given state at a given time satisfying the
+Chemical Master Equation (CME).
```
isaacsas, line 28: This is likely not going to make much sense to anyone who doesn't know point-processes, which would be many of our users.

isaacsas (commented):
While I appreciate your effort with the getting started tutorial, I don't think it improves the new user experience overall. I'd prefer if the getting started tutorial just points users to the appropriate tutorial based on their background (e.g. point-processes, jump processes in physics or chemistry, etc). Otherwise it needs to explain the equations it is using, and explain what is appearing in them for each of these different communities so they can understand what it is doing. 

Perhaps we should instead add a brief but fully explained mathematical discussion as requested in the review as the "getting started tutorial", and then link there to the concrete tutorials for people with different backgrounds. That would avoid needing to try to write a single tutorial that is accessible to all fields, while providing a single point that sends users to the tutorial that might work best for them.

docs/src/tutorials/jump_diffusion.md
```diff
@@ -4,17 +4,25 @@
...
-tutorial, we will show how to solve problems with even more general jumps. In the
-special case that ``g_j = 0`` for all ``j``, we'll call the resulting jump-ODE a
-[piecewise deterministic Markov
+The change in state vector ``u`` depends on three concurrent dynamics, (1)
+deterministic ``dt`` with size ``f(u, p, t)``, (2) stochastic diffusion
+``dW_j(t)`` with size ``g_j(u, p, t)``, and, (3) stochastic jump
+``dN_i(t)`` with size ``h_i(u, p, t)``. By diffusion we mean a continuous
+stochastic process which is usually represented as Gaussian white noise.
+By jump we mean a discrete stochastic process which is usually represented
+by a Poisson process.
```
isaacsas, line 20:
```suggestion
The change in state vector ``u`` depends on three concurrent dynamics, (1)
deterministic drift ``dt`` with size ``f(u, p, t)``, (2) stochastic diffusions
``dW_j(t)`` with sizes ``g_j(u, p, t)``, and, (3) stochastic jumps
``dN_i(t)`` with sizes ``h_i(u, p, t)``. By diffusion we mean a continuous
stochastic process which is usually represented as Gaussian white noise (i.e. ``W_j(t)`` is a Brownian Motion).
By jump we mean a discrete stochastic process which is usually represented
by a Poisson process.
```

docs/src/tutorials/jump_diffusion.md
```diff
@@ -124,17 +135,10 @@ Using the same `affect!` we build a [`VariableRateJump`](@ref):
...
-```@example tut3
-jump2 = deepcopy(jump)
-```
-
-so that way we have two independent jump processes. We now couple these jumps
-to the `ODEProblem`:
+We now couple the jump to the `ODEProblem`:
 
 ```@example tut3
-jump_prob = JumpProblem(prob, Direct(), jump, jump2)
```
isaacsas, line 137: Why drop this? Without it the tutorial only shows examples that use one jump, which may give a misleading impression that one can't have more than one jump?

docs/src/tutorials/simple_poisson_process.md
```diff
@@ -6,6 +6,12 @@ primarily in chemical or population process models, where several types of jumps
 may occur, can skip directly to the [second tutorial](@ref ssa_tutorial) for a
 tutorial covering similar material but focused on the SIR model.
 
+!!! note
+    
+    If you have a background in point process theory, we recommend that
+    you read our example on [modelling temporal point processes with
+    JumpProcesses](@ref tpp_tutorial)
+
```
isaacsas, line 14: I would just merge this into the first paragraph for parallel construction and say something as the new last sentence like "Readers with a background in point process theory can skip directly to [your new tutorial]."

docs/src/tutorials/simple_poisson_process.md
```diff
@@ -128,11 +134,13 @@ dprob = DiscreteProblem(u₀, tspan, p)
...
-jump type). Here we use the classical `Direct` method, proposed by Gillespie in
-the chemical reaction context, but going back to earlier work by Doob and others
-(and also known as Kinetic Monte Carlo in the physics literature)
+specifies which sampling algorithm, called an aggregator in JumpProcesses, to
+use for determining next jump times (and in the case of multiple possible jumps
+the next jump type). We call them aggregators because these algorithms are
+methods for aggregating a set of jumps to determine the next jump time.  Here
+we use the classical `Direct` method, proposed by Gillespie in the chemical
+reaction context, but going back to earlier work by Doob and others (and also
+known as Kinetic Monte Carlo in the physics literature)
```
isaacsas, line 143:
```suggestion
specifies which sampling algorithm, called an aggregator in JumpProcesses, to
use for determining next jump times (and in the case of multiple possible jumps
the next jump type). We call them aggregators because these algorithms are
methods for aggregating a set of jumps to determine the next jump time and type.  Here
we use the classical `Direct` method, proposed by Gillespie in the chemical
reaction context, but going back to earlier work by Doob and others (and also
known as Kinetic Monte Carlo in the physics literature)
```

docs/src/tutorials/simple_poisson_process.md
```diff
@@ -179,8 +187,11 @@ deathcrj = ConstantRateJump(deathrate, deathaffect!)
 ```
 
 As the death rate is constant *between* jumps, we can encode this process as a
-second `ConstantRateJump`. We then construct the corresponding problems, passing
-both jumps to `JumpProblem`, and can solve as before
+second `ConstantRateJump`. Overall, we keep track of four different processes,
+the population size ``N``, total deaths ``D``, births ``Y_b`` and deaths
+``Y_d``. ``N`` and ``D`` are represented by ``u`` in our code, ``Y_b`` by `crj`
+and ``Y_d`` by `death_crj`. We then construct the corresponding problems,
+passing both jumps to `JumpProblem`, and can solve as before
```
isaacsas, line 194:
```suggestion
second `ConstantRateJump`. Overall, we keep track of four different processes,
the population size ``N(t)``, total deaths ``D(t)``, births ``Y_b(t)`` and deaths
``Y_d(t)``. The vector ``(N(t), D(t))`` is represented by ``u`` in our code, with ``Y_b(t)`` encoded by `crj`
and ``Y_d(t)`` by `death_crj`. We then construct the corresponding problems,
passing both jumps to `JumpProblem`, and can solve as before
```

docs/src/tutorials/simple_poisson_process.md
```diff
@@ -252,20 +263,25 @@ lrate(u, p, t) = p.λ
...
+Finally, to efficiently simulate the new jump process, we must also
+specify a dependency graph. This indicates when a given jump occurs, which
+jumps in the system need to have their rates and/or rate bounds
+recalculated (for example, due to depending on changed components in `u`).
+Internally, JumpProcesses preserves the relative ordering of jumps of each
+distinct type, but always reorders all `ConstantRateJump`s to appear
+before any `VariableRateJump`s. Irrespective of how `JumpProblem` is
+initialized, we can think that internally the jumps will get ordered as
+vector `[deathcrj, vrj1]`. This vector of jumps is not the same as our
+state variable `u`.
```
isaacsas, line 275:
```suggestion
before any `VariableRateJump`s. Irrespective of how `JumpProblem` is
initialized, internally the jumps will be ordered as the vector `[deathcrj, vrj1]`. Note, this vector of the jumps is distinct from our state variable vector, `u`.
```

isaacsas (changes requested):

isaacsas:
@gzagatti left you some comments.

docs/Project.toml
gzagatti: Fixed. I have also updated all the dependencies.

docs/src/assets/Project.toml
gzagatti: Fixed. I have also updated all the dependencies.

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -580,14 +582,15 @@ required for bounded `VariableRateJump`s, we have for any `s` in `[t,t + rateint
...
+removed from `p[5]` for every recovery.
 
 ```@example tut2
 rate4(u, p, t) = p[2] * u[2]         # ν*I
 function affect4!(integrator)
     integrator.u[2] -= 1
     integrator.u[3] += 1
-    length(H) > 0 && deleteat!(H, rand(1:length(H)))
+    length(integrator.p[5]) > 0 &&
+        deleteat!(integrator.p[5], rand(1:length(integrator.p[5])))
```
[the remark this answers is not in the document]
gzagatti: I have replace the array with a dictionary. As I understand, dictionaries in Julia are implemented as Hash maps so insertions and deletions should be O(1).

We don't use random removal on the MultivariateHawkes benchmark. The brute-force version of the benchmark only pushes to the array.

docs/src/tutorials/jump_diffusion.md
gzagatti: I dropped because I felt the goal of this section (as the title suggests) was to illustrate variable jump in the context of SDEs. Therefore, adding another jump in this section can be confusing to the reader. My first reaction was "Why not just double the rate to achieve the same effect?" It got me thinking that there was some sort of trick that I was not getting.

I think if the intention is to demonstrate the case of multivariate jumps  in SDEs. Why don't we just create another section appropriately titled to illustrate such problems? Then we can just mix the variable and constant jumps from the previous two sections.

I will adjust according to your final say on that.

docs/src/tutorials/simple_poisson_process.md
gzagatti: Fixed.

docs/src/tutorials/jump_diffusion.md
gzagatti: I have pushed a suggestion with this extra section.

docs/src/index.md
```diff
@@ -7,19 +7,32 @@ methods, or Kinetic Monte Carlo methods. It also enables the incorporation of
 jump processes into hybrid jump-ODE and jump-SDE models, including jump
 diffusions.
 
+Historically, jump processes have been developed in the context of dynamical
+systems to describe dynamics with sudden changes — the jumps — in a
+system's value at random times. In contrast, the development of point processes
+has been more focused on describing the occurrence of random events — the
+points — over a support. In reality, jump and point processes share many
+things in common which make JumpProcesses ideal for both.
+
```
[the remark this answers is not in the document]
gzagatti: Fixed.

docs/src/getting_started.md
gzagatti: Explanation added.

gzagatti:
@isaacsas I have addressed most of your feedback and thanks for taking the time.

Regarding the "Getting Started", I got a bit stuck. I made a few changes to keep it simpler. But I think removing all the code will sort of make it redundant with the index since the index already contains some brief explanation of the maths and links to all tutorials.

My goal with "Getting Started" was to go through the motion of setting the problem up and solving with little else. It even breaks it down on the same 3 steps as in DifferentialEquations and discuss the very basic constructs in the library. For instance, the reviewer was confused with terms like aggregator and stepper. This page provides brief explanation of them in context.

I feel that a beginner would get a bit lost with the remaining tutorials because they are much longer and address more specific models.

Feel free to remove the "Getting Started"  tutorial if you still not convinced. Or even to refactor it to your liking.

docs/src/tutorials/discrete_stochastic_example.md
```diff
@@ -677,6 +683,10 @@ will no longer be exact for a pure jump process, as the solution values at jump
 times have not been stored. i.e for `t` a time we did not save at `sol(t)` will
 no longer give the exact value of the solution at `t`.*
 
+In summary, the jump callback will save the state variable `u` before and
+after the jump whenever `save_positions = (true, true)`. All other saving
+behaviour is delegated to the integrator and controlled via `saveat`.
```
isaacsas, line 688:
```suggestion
behaviour is delegated to the integrator and controlled via keyword arguments to `solve` such as `saveat`.
```

PR #376, opened 2023-12-06
WIP: Doc update V2

isaacsas:
Revision of https://github.com/SciML/JumpProcesses.jl/pull/372


docs/src/tutorials/point_process.md
```diff
@@ -0,0 +1,826 @@
...
+```@example tpp
+using JumpProcesses
+using PointProcesses
+
+struct SciMLPointProcess{M, J <: JumpProcesses.AbstractJump, G, D, T <: Real} <:
+       AbstractPointProcess{M}
+    jumps::Vector{J}
+    mark_dist::Vector{D}
+    g::G
+    p::Any
```
isaacsas, line 90: @gzagatti why `Any`?

gzagatti: No particular reason. I just meant to say the parameters are up to the user. Should we remove `Any` or use `Tuple` instead?

gzagatti: In Julia, I never know when to use `Any` and when to leave it blank.

PR #387, opened 2023-12-30
drop tests for indexing problem with a parameter

isaacsas:
Since indexing problems by a parameter has been dropped by SciMLBase we shouldn't test for this anymore.


TorkelE:
For reference, not sure whether this is implemented or not, but I think the plan is to make something like;
```julia
@parameters k
...
prob.ps[k]
```
work for parameter indexing. Basically, due to parameters having some property, allowing
```julia
prob[k]
```
directly (like what is done for initial conditions) becomes type unstable. hence the indexing has to be done on the `.ps` field instead. When it is clear that this is ready, we should move to that instead.

ChrisRackauckas:
> For reference, not sure whether this is implemented or not, but I think the plan is to make something like;

Set `getp`/`setp`.

But again this is separate from whether these tests should be here. The symbolic indexing interface has all of those things, but we are dropping specifically definition of names by defining `f.syms`

isaacsas:
For some reason 1.10 CI seems bugged. Not sure why it is taking over an hour (I've already restarted it once).

isaacsas:
OK, so we should just drop these tests completely then? Since otherwise there is no real way to create a system here (i.e. we don't want to have to use MTK to run the tests...)

ChrisRackauckas:
You can define a SymbolCache `f.sys`

src/problem.jl
```diff
@@ -122,9 +122,12 @@ end
...
     SciMLBase.___internal_setindex!(prob.prob, args...; kwargs...)
-    if using_params(prob.massaction_jump)
-        update_parameters!(prob.massaction_jump, prob.prob.p)
-    end
+
+    # since parameters are no longer allowed to be mutated and the preceding will error this
+    # isn't needed
+    # if using_params(prob.massaction_jump)
+    #     update_parameters!(prob.massaction_jump, prob.prob.p)
+    # end
```
ChrisRackauckas, line 130: `setp` will still allow for issues, along with manual mutation.

isaacsas: We probably need a custom dispatch on it then to update MassActionJumps?

ChrisRackauckas: Yes, @AayushSabharwal can probably help with that.

isaacsas: That would be great, thanks!

isaacsas:
> You can define a SymbolCache `f.sys`

Ok, that is very helpful to know. I was confused from looking at the example test file in SII and thought I’d have to completely build a custom struct satisfying the interface.

ChrisRackauckas:
We'll make it a tutorial, https://docs.sciml.ai/SymbolicIndexingInterface/dev/simple_sii_sys/ it's just missing right now.

test/jprob_symbol_indexing.jl
```diff
@@ -6,14 +6,14 @@ affect1!(integ) = (integ.u[1] += 1)
 affect2!(integ) = (integ.u[2] += 1)
 crj1 = ConstantRateJump(rate1, affect1!)
 crj2 = ConstantRateJump(rate2, affect2!)
-g = DiscreteFunction((du, u, p, t) -> nothing; syms = [:a, :b], paramsyms = [:p1, :p2])
+g = DiscreteFunction((du, u, p, t) -> nothing; sys = SymbolicIndexingInterface.SymbolCache([:a, :b], [:p1, :p2], :t)
```
ChrisRackauckas, line 9:
```suggestion
g = DiscreteFunction((du, u, p, t) -> nothing; sys = SymbolicIndexingInterface.SymbolCache([:a, :b], [:p1, :p2], :t))
```

src/problem.jl
AayushSabharwal: We now have `set_parameter!` which is used by `setp` internally, and can be implemented to do other bookkeeping actions when `setp` is used.

AayushSabharwal:
SciMLBase master supports `prob.ps[p]` syntax now, so parameter indexing tests can use that instead

test/jprob_symbol_indexing.jl
```diff
@@ -6,14 +6,14 @@ affect1!(integ) = (integ.u[1] += 1)
...
 # runs the tests
 @test jprob[:a] == 0
 @test jprob[:b] == 10
-@test jprob[:p1] == 1.0
-@test jprob[:p2] == 2.0
 
-# tests for setindex (e.g. `jprob[:a] = 10`) not possible, this requires the problem to have a .f.sys filed.,
+# these are no longer supported by SciMLBase
+@test getp(jprob,:p1)(jprob) == 1.0
+@test getp(jprob,:p2)(jprob) == 2.0
```
AayushSabharwal, line 19:
```suggestion
@test getp(jprob,:p2)(jprob) == 2.0
@test jprob.ps[:p1] == 1.0
@test jprob.ps[:p2] == 2.0
```

AayushSabharwal:
Master isn't tagged yet, so this won't work

test/jprob_symbol_indexing.jl
```diff
@@ -6,14 +6,16 @@ affect1!(integ) = (integ.u[1] += 1)
...
+@test jprob.ps[:p1] == 1.0
+@test jprob.ps[:p2] == 2.0
```
isaacsas, line 21: Where has the dispatch been added to ensure `update_parameters!(prob.massaction_jump, prob.prob.p)` is called when these parameter changes are made?

src/problem.jl
isaacsas: @AayushSabharwal, can you add the needed dispatches to that when parameters get modified, i.e. something like what we had before when modifying a problem:

```julia
if using_params(prob.massaction_jump)
    update_parameters!(prob.massaction_jump, prob.prob.p)
end
```

Alternatively, @ChrisRackauckas do we want to put this on users (since this could be expensive, so a faster workflow is to only call it once all parameter modifications are finished)? That does leave a gotcha someone could encounter unless they carefully read the docs (which is why the previous behavior here always updated directly after modifying a parameter).

isaacsas: But we do put calling the jump update function on users currently when modifying integrators within callbacks, so perhaps it isn't unreasonable to do that.

isaacsas:
Handled in https://github.com/SciML/JumpProcesses.jl/pull/408 now.

PR #404, opened 2024-03-04
Complete 2nd rebuttal, fix #402

### LINKED ISSUE #402: Remarks about paper - episode 2 [JuliaCon Proceedings review]
https://github.com/JuliaCon/proceedings-review/issues/133

I have re-read the paper, overall the presentation has improved indeed! In particular, this time around I have finally understood the syncing aspect which makes your algorithm different. Thus I was able to focus a bit more on the mathematical part at the beginning.

My main remaining issues are:

- The introduction of `Coevolve` still doesn't make it clear what is yours and what is theirs. I suggest introducing a new name, maybe CoevolveSync, since from what I understand this is your major contribution to an existing algorithm
- The section on inverse methods is rather confusing and could be shortened
- In Algorithm 5 it would be nice to be able to compare with a naive queueing algorithm, to see the benefits of syncing
- On the easiest benchmark I don't understand why queueing methods perform poorly: are these instances 1-dimensional (in the mark space)?
- The comparison with Tick is great, I would also add brief references to other point process libraries in Python and R

You can find all my remarks as annotations on the [PDF file](https://github.com/SciML/JumpProcesses.jl/files/14267579/Zagatti.et.al.-.2023.-.Extending.JumpProcess.jl.for.fast.point.process.si.pdf). Is it okay as a format for exchanging remarks? Can you read them?

gzagatti:
This PR is the second round of rebuttal. It fixes #402.

Please also see my responses to the reviewer in the original issue.

This time I have made all my changes using the `\comment` macro to make it easier to spot all my changes in magenta in the compiled PDF. Once we are satisfied with all the changes. I will remove the macro and leave only the text. I hope that it makes it easier to review it that way.

cc: @gdalle


gzagatti:
Find the compiled paper [here](https://github.com/SciML/JumpProcesses.jl/files/14487046/paper.pdf) for easier reference.

isaacsas:
LGTM. Should I merge?

One minor comment, if you wanted to add a sentence speculating on why NRM/Coevolve don't offer the top performance on constant rate problems, you could say that it is due to the cost of updating their underlying indexed priority queue data structure which stores the rates. A table-based data-structure would be expected to be more competitive [this is the Sanft-Othmer paper](https://arxiv.org/abs/1503.05832), and a real cache-optimized indexed priority queue would give very good performance (but perhaps not beating RSSACR on the largest problems). The latter would be a great project actually that if as performant as I expect could be a J. Chem. Phys. paper...

gzagatti:
Thanks for the quick turnaround! Let me add your minor comment and remove the comment markup before you merge.

gzagatti:
[Draft](https://github.com/SciML/JumpProcesses.jl/files/14487315/paper.pdf) with minor comment added.

I will now remove the comment markup and let you know once it is good to merge.

isaacsas:
The highlighting makes it much easier to quickly see what has changed!

gzagatti:
Now, it's good to merge.

PR #408, opened 2024-03-20
Indexing fixes

isaacsas:


isaacsas:
@AayushSabharwal can you check I've got the correct dispatches now for mutating parameters within JumpProblems ensuring `update_parameters!(...)` is called to update the associated MassActionJump?

isaacsas:
If this looks ok I'll add some more symbolic tests for integrators and for remake.

src/problem.jl
```diff
@@ -122,9 +122,17 @@ end
...
+
+# for updating parameters in JumpProblems to update MassActionJumps
+function SII.set_parameter!(prob::JumpProblem, val, idx)
+    ans = SII.set_parameter!(SII.parameter_values(prob), val, idx)
+
     if using_params(prob.massaction_jump)
         update_parameters!(prob.massaction_jump, prob.prob.p)
     end
+
+    ans
```
AayushSabharwal, line 135: While this is technically correct, `set_parameter!` will be called once for each parameter which would make this _very_ expensive. I'll PR to SII to add a callback that runs at the end of `setp` (`finalize_parameters_hook!(prob, ps)`?)

AayushSabharwal: https://github.com/SciML/SymbolicIndexingInterface.jl/pull/58

isaacsas: @AayushSabharwal please ping me when a SII release with that callback is available so I can get this finished -- we need this for making updates to Catalyst for MTK9. Thanks!

AayushSabharwal: It's been released now @isaacsas

AayushSabharwal: Oops, I thought it was merged. Sorry 😅

AayushSabharwal: _Now_ it's out in 0.3.13

isaacsas:
@ChrisRackauckas this hadn't yet been updated for https://github.com/SciML/SymbolicIndexingInterface.jl/pull/58, which is the interface we are supposed to use (and was only merged/released today).

I'll open another PR to update this to the correct interface, hopefully later today or tomorrow.

PR #412, opened 2024-04-06
Update simple_regular_solve.jl with new function BinomialTauLeaping

c-louhichi:
Add a function for binomial tau leaping :)

## Checklist

- [ ] Appropriate tests were added
- [ ] The new code follows the

### DIFF
--- a/src/simple_regular_solve.jl
+++ b/src/simple_regular_solve.jl
@@ -51,3 +51,105 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
 end
 
 export SimpleTauLeaping
+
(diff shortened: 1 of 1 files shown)


isaacsas:
@ChrisRackauckas maybe it would make sense to allow passing non-regular jump types to regular jump methods too so they can specialize their code? This would allow @c-louhichi to specialize on mass action jump input for now.

isaacsas:
@c-louhichi can you also update https://github.com/SciML/JumpProcesses.jl/blob/master/test/regular_jumps.jl with an example using / testing the output of your new solver?

isaacsas:
We probably don't want to have an expression-based analysis here, but let's get a test in and then I can leave more detailed feedback.

PR #424, opened 2024-06-13
Test master

ChrisRackauckas:


ChrisRackauckas:
@oscardssmith you've been looking at this recently, is this the same FastBroadcast change?

ChrisRackauckas:
This array type acts like `x.u` but then has extra `x.jump_u` that are kept along for broadcast. It's a very weird array. Maybe its FastBroadcast extension should just opt-out so it can use its broadcast overloads.

oscardssmith:
oh this is fun. It turns out that FastBroadcast@0.3 just works without the extension, but the extension breaks it.

PR #425, opened 2024-06-13
restrict FastBroadcast to 0.3 and delete the extension

oscardssmith:
Surprisingly enough, with the new FastBroadcast version, everything just seems to work out of the box (and the extension was broken)


isaacsas:
Everything just worked before the extension too, but without the extension it was falling back in very slow default versions. The extension was added because the performance was terrible with these fallbacks.

isaacsas:
What performance impact does dropping the extension have on systems with VariableRateJumps?

isaacsas:
https://github.com/SciML/JumpProcesses.jl/issues/335

It would be unifortunate if the result of this is undoing all the work done to solve that issue

oscardssmith:
Looks like the performance is still good. Using the benchmark from the end of that issue, I see
```
julia> function test_single_dot(out, array)
            @inbounds  @. out = array + 1.0 * array + 1.2 * array
       end
test_single_dot (generic function with 1 method)

julia> @benchmark test_single_dot(bench_out_array, bench_in_array)
BenchmarkTools.Trial: 7308 samples with 1 evaluation.
 Range (min … max):  600.089 μs …  1.456 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     651.551 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   676.034 μs ± 80.437 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

    ▄█▅▂                                                        
  ▂▆█████▇▇▇▆▆▅▅▄▄▃▃▃▃▃▂▂▂▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  600 μs          Histogram: frequency by time         1.01 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark test_single_dot(base_out_array, base_in_array)
BenchmarkTools.Trial: 7592 samples with 1 evaluation.
 Range (min … max):  604.956 μs …  1.456 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     632.637 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   650.833 μs ± 56.660 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▄▇█▆▃                                                      
  ▁▄███████▆▅▄▄▄▄▄▄▄▄▃▃▂▃▂▂▂▂▂▂▂▂▂▂▂▁▁▁▂▁▁▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  605 μs          Histogram: frequency by time          829 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

isaacsas:
Ok, thanks for checking.

PR #430, opened 2024-07-25
Fix saveat plotting

isaacsas:
This preserves using piecewise constant plotting when saving every jump, but switches to less misleading linear interpolation when not saving every jump.

Requires https://github.com/SciML/SciMLBase.jl/pull/749

Closes https://github.com/SciML/JumpProcesses.jl/issues/206


test/save_positions.jl
```diff
@@ -1,28 +1,59 @@
...
+    end
+
+    # for mixed problems sol.dense currently ignores save_positions
+    oprob = ODEProblem((du,u,p,t) -> du[1] = .1, u0, tspan)
+    for sp in sps
+        jprob = JumpProblem(oprob, Direct(), crj; save_positions = sp, rng)
+        sol = solve(jprob, Tsit5())
+        @test sol.dense == true
+        @test SciMLBase.isdenseplot(sol) == true
+    end
```
isaacsas, line 57: Currently with PDMPs or jump-diffusions `sol.dense == true` always, irregardless of `save_positions`. I'm not sure if this should be considered an issue or not.

isaacsas:
Closes https://github.com/SciML/JumpProcesses.jl/pull/430

PR #443, opened 2024-08-12
periodic callback and tstop aliasing fixes

isaacsas:
Closes https://github.com/SciML/JumpProcesses.jl/issues/417.
Closes https://github.com/SciML/JumpProcesses.jl/issues/442

This also adds an `alias_tstops` keyword to `solve` allowing the `tstop` vector to be aliased. By default it is `false`.

Should be good to go once
- https://github.com/SciML/DiffEqBase.jl/pull/1071 and a DiffEqBase release
- https://github.com/SciML/OrdinaryDiffEq.jl/pull/2377 and OrdinaryDiffEq release
- https://github.com/SciML/StochasticDiffEq.jl/pull/580 and StochasticDiffEq release
- https://github.com/SciML/Sundials.jl/pull/459 and Sundials release
- https://github.com/SciML/DiffEqCallbacks.jl/pull/229 and a DiffEqCallbacks release

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -110,7 +110,7 @@ DiffEqBase.get_tstops_array(integrator::SSAIntegrator) = DiffEqBase.get_tstops(i
 
 # ODE integrators seem to add tf into tstops which SSAIntegrator does not do
 # so must account for it here
-function DiffEqBase.get_max_tstops(integrator::SSAIntegrator)
+function DiffEqBase.get_tstops_max(integrator::SSAIntegrator)
     tstops = DiffEqBase.get_tstops_array(integrator)
     tf = integrator.sol.prob.tspan[2]
     if !isempty(tstops)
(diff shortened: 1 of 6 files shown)


src/SSA_stepper.jl
```diff
@@ -120,6 +124,12 @@ function DiffEqBase.solve!(integrator::SSAIntegrator)
     end
     integrator.t = end_time
 
+    # check callbacks one last time
+    if !(integrator.opts.callback.discrete_callbacks isa Tuple{})
+        DiffEqBase.apply_discrete_callback!(integrator,
+            integrator.opts.callback.discrete_callbacks...)
+    end
+
```
isaacsas, line 132: @ChrisRackauckas wanted to check this one with you. As far as I can tell SSAStepper was not calling callbacks at the final time, `tspan[2]`. For the periodic callback this resulted in `affect_final = true` being ignored. I added this last callback check at the final time once time-stepping is done, but it would be good for you to check this is the right approach to handle this / consistent with other integrators. 

Note that `should_continue_solve`'s setup is to continue stepping if the next `tstop` or jump time are *less* than `end_time`, hence why the `final_affect` is not hit via `step!`.

ChrisRackauckas: That's a pretty reasonable way to fix it.

PR #450, opened 2024-09-02
fix vrj reinitialization bug

isaacsas:
It seems that the code to reset variable rate jump problems was not correctly broadcasting calls to `randexp` resulting in the **same** samples being used for all jumps.

cc: @ChrisRackauckas 

Closes https://github.com/SciML/JumpProcesses.jl/issues/320

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -58,7 +58,8 @@ function resetted_jump_problem(_jump_prob, seed)
 
     if !isempty(jump_prob.variable_jumps)
         @assert jump_prob.prob.u0 isa ExtendedJumpArray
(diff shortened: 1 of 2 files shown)


isaacsas:
Now that we have a Catalyst VRJ interface I think this is indicative that we really need to add more VRJ correctness tests. 

Even with this fix https://github.com/SciML/JumpProcesses.jl/issues/320 still doesn't seem to work.

isaacsas:
If tests pass this should be good to go (and we should then make an immediate release).

ChrisRackauckas (approved):

PR #451, opened 2024-09-02
make terminate! really terminate

isaacsas:
The post-stepping call to `apply_discrete_callback!` and all post-stepping saving is now discarded if a user has called `terminate!`. This seems like the desired behavior, but let me know if I should modify this.

### DIFF
--- a/src/SSA_stepper.jl
+++ b/src/SSA_stepper.jl
@@ -122,27 +122,31 @@ function DiffEqBase.solve!(integrator::SSAIntegrator)
     while should_continue_solve(integrator) # It stops before adding a tstop over
         step!(integrator)
     end
-    integrator.t = end_time
 
-    # check callbacks one last time
-    if !(integrator.opts.callback.discrete_callbacks isa Tuple{})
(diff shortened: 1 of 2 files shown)


src/SSA_stepper.jl
```diff
@@ -122,27 +122,31 @@ function DiffEqBase.solve!(integrator::SSAIntegrator)
...
+        # check callbacks one last time
+        if !(integrator.opts.callback.discrete_callbacks isa Tuple{})
+            DiffEqBase.apply_discrete_callback!(integrator,
+                integrator.opts.callback.discrete_callbacks...)
         end
-    end
 
-    if integrator.save_end && integrator.sol.t[end] != end_time
-        push!(integrator.sol.t, end_time)
-        push!(integrator.sol.u, copy(integrator.u))
```
isaacsas, line 145: None of this code was actually changed, it is just now inside the `if` statement and only called if the return code does not indicate termination occurred.

ChrisRackauckas (approved):

PR #455, opened 2024-09-19
[Ready for Review] - Constant-complexity NRM implementation

vyudu:
An implementation of the method in [Sanft/Othmer](https://www-users.cse.umn.edu/~othmer/papers/Sanft15.pdf).


Vilin97:
Please correct me if I'm wrong but isn't this `Direct-CR`?

Vilin97:
The paper you link and the citation you give in the code are different. Which one is correct?

src/aggregators/ccnrm.jl
```diff
@@ -0,0 +1,158 @@
+# Implementation the original Next Reaction Method
```
Vilin97, line 1: This citation is wrong, no?

src/aggregators/prioritytable.jl
```diff
@@ -1,16 +1,16 @@
...
+    Implementation
+    Stores ids that identify the priorities. Each group stores a range of priority ids.
+    The basic design assumes a lower-most group of "zero" priorities, and a second
+    group storing all non-zero priorities that are < a `minpriority`. All other groups
+    store priorities within consecutive ranges. The `minpriority` is taken fixed at creation,
+    but the `maxpriority` will increase dynamically based on inserted / updated priorities.
+    The ranges are assumed to be powers of two:
+       bin 1 = {0},
+       bin 2 = (0,`minpriority`),
+       bin 3 = [`minpriority`,`2*minpriority`)...
```
Vilin97, line 13: No need to change the formatting

test/runtests.jl
```diff
@@ -16,7 +16,7 @@ using JumpProcesses, DiffEqBase, SafeTestsets
     @time @safetestset "Mass Action Jump Tests; Gene Expr Model" begin include("geneexpr_test.jl") end
     @time @safetestset "Mass Action Jump Tests; Nonlinear Rx Model" begin include("bimolerx_test.jl") end
     @time @safetestset "Mass Action Jump Tests; Special Cases" begin include("degenerate_rx_cases.jl") end
-    @time @safetestset "Direct allocations test" begin include("allocations.jl") end
+    # @time @safetestset "Direct allocations test" begin include("allocations.jl") end
```
Vilin97, line 19: Why comment out this test?

src/aggregators/ccnrm.jl
```diff
@@ -0,0 +1,158 @@
...
+    rates = p.rates
+    idx = get_num_majumps(majumps) + 1
+    @inbounds for rate in rates
+        cur_rates[idx] = rate(u, params, t)
+        pttdata[idx] = t + randexp(p.rng) / cur_rates[idx]
+        idx += 1
+    end
+
+    # Build the priority time table with the times and bin width. 
+    binwidth = 16 / sum(cur_rates)
```
Vilin97, line 154: Do not hardcode `16` everywhere. Make it a global constant or a field in the aggregator

Vilin97 (commented):
Thank you for the PR!

1. Can you please explain what is the difference between this SSA and `DirectCRDirect` (same Sanft 2015 paper is cited there) and also `DirectCR` (the binning strategy looks similar)? Thank you.

2. It would be helpful if you reported the relative performance on the new SSA on one or two problems, e.g. from tests, compared to other SSAs in the library. If it's much slower (>2x) than similar binning SSAs like `DirectCR`, we will need to look into it more.

3. Also, I left some more specific comments about the code.

vyudu:
Hi Vasily, the main difference with DirectCR is that the priority table in this method stores next reaction times rather than reaction rates, and this implementation doesn't handle spatial jumps like DirectCRDirect. But admittedly I am a little confused about the citation for DirectCRDirect being the Sanft/Othmer paper -- they do discuss a spatial generalization of CCNRM, but it does not seem equivalent to DirectCRDirect (since it involves storing reaction times instead of rates). 

Performance-wise this is definitely a WIP (will convert it to a draft). Like you say it is slower and allocates much more memory than DirectCR or NRM (which is why I currently have the allocations test commented out, since it fails), but am submitting it in this state to hopefully get some feedback on how to improve this aspect, cc @isaacsas. Working on rewriting the methods for the PriorityTimeTable to allocate less.

src/aggregators/ccnrm.jl
```diff
@@ -81,17 +83,16 @@ end
 # calculate the next jump / jump time
 # just the first reaction in the first non-empty bin in the priority table
 function generate_jumps!(p::CCNRMJumpAggregation, integrator, u, params, t)
-    next_jump, next_time = getfirst(p.ptt)
+    p.next_jump, p.next_jump_time = getfirst(p.ptt)
 
     # Rebuild the table if no next jump is found. 
-    if next_jump === nothing
+    if p.next_jump === nothing
```
isaacsas, line 89: I don't think you want to return `nothing` so that the return type has to be a `Union` that can potentially have performance implications. Why not just return `0` which keeps the same type but is an invalid index for a reaction?

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
+    ptt.steps += 1
+    min_time = typemax(eltype(times))
+    min_idx = 0
+    for i in 1:(groups[minbin].numpids)
+        pid = groups[minbin].pids[i]
+        times[pid] < min_time && begin
+            min_time = times[pid]
+            min_idx = pid
+        end
+    end
```
isaacsas, line 429:
```suggestion
    for pid in groups[minbin].pids
        times[pid] < min_time && begin
            min_time = times[pid]
            min_idx = pid
        end
    end
```

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
+
+    # Reinsert the times into the groups. 
+    for (id, time) in enumerate(times)
+        time > ptt.maxtime && continue
+        insert!(ptt, id, time)
+    end
+    ptt.minbin = findfirst(g -> g.numpids > (0), groups)
+    ptt.minbin === nothing && (ptt.minbin = 0)
+    ptt.steps = 0
```
isaacsas, line 404:
```suggestion
    ptt.steps = 0
    
    return nothing
```

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
+
+    return min_idx, min_time
+end
+
+function insert!(ptt::PriorityTimeTable, pid, time)
+    @unpack timegrouper, pidtogroup, groups = ptt
+    gid = timegrouper(time)
+    @inbounds pididx = insert!(groups[gid], pid)
+    @inbounds pidtogroup[pid] = (gid, pididx)
+end
```
isaacsas, line 439:
```suggestion
    @inbounds pidtogroup[pid] = (gid, pididx)
    
    return nothing
end
```

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
+# Update the priority table when a reaction time gets updated. We only shift
+# between bins if the new time is within the current time window; otherwise
+# we remove the reaction and wait until rebuild. 
+function update!(ptt::PriorityTimeTable{T, F}, pid, oldtime, newtime) where {T, F}
+    @unpack times, timegrouper, maxtime, pidtogroup, groups = ptt
+
+    times[pid] = newtime
+    if oldtime >= maxtime
+        # If a reaction comes back into the time window, insert it. 
+        newtime < maxtime ? insert!(ptt, pid, newtime) : return
```
isaacsas, line 450:
```suggestion
        newtime < maxtime ? insert!(ptt, pid, newtime) : return nothing
```

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
+        oldgid == newgid && return
+        @inbounds begin
+            pidx = pidtogroup[pid][2]
+            movedpid = remove!(groups[oldgid], pidx)
+            pidtogroup[movedpid] = (oldgid, pidx)
+            newpidx = insert!(groups[newgid], pid)
+            pidtogroup[pid] = (newgid, newpidx)
+        end
+    end
+end
```
isaacsas, line 472:
```suggestion
    end
    return nothing
end
```

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
+            gid, pidx = pidtogroup[pid]
+            movedpid = remove!(groups[gid], pidx)
+            pidtogroup[movedpid] = (gid, pidx)
+            pidtogroup[pid] = (zero(F), zero(F))
+        end
+    else
+        # Move bins if the reaction was already inside. 
+        oldgid = timegrouper(oldtime)
+        newgid = timegrouper(newtime)
+        oldgid == newgid && return
```
isaacsas, line 463:
```suggestion
        oldgid == newgid && return nothing
```

isaacsas (changes requested):
Make sure to `@inbounds` hot loops with indexing when this is ready to merge.

src/aggregators/prioritytable.jl
vyudu: A bit uncertain if this works - since the number of active PIDs may be fewer than the total number of PIDs you could be accessing PIDs that are no longer in the bin if you iterate over all of them.

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
 end
+
+##########################
+### Routines for CCNRM ###
+##########################
+
+mutable struct TimeGrouper{T <: Number}
+    mintime::T
+    timestep::T
+end
```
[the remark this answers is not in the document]
vyudu: Can definitely do this, though I think I may misunderstand how memory works in this case - I'd thought that if it's a mutable struct and you're just changing the fields, then no memory actually gets allocated? Whereas if you're creating an immutable struct repeatedly each one is a separate allocation that adds up.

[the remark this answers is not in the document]
isaacsas: Accessing a mutable struct is usually via a pointer though (sometimes Julia can avoid this but I don't know how easy it is to figure this out a priori).

src/aggregators/ccnrm.jl
```diff
@@ -0,0 +1,161 @@
...
+        # update the jump rate
+        @inbounds cur_rates[rx] = calculate_jump_rate(ma_jumps, num_majumps, rates, u,
+            params, t, rx)
+
+        # Calculate new jump times for dependent jumps
+        if rx != p.next_jump && oldrate > zero(oldrate)
+            if cur_rates[rx] > zero(eltype(cur_rates))
+                update!(ptt, rx, oldtime, t + oldrate / cur_rates[rx] * (times[rx] - t))
+            else
+                update!(ptt, rx, oldtime, 2 * end_time)
```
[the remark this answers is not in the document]
isaacsas: The problem is that in theory we should support a user changing `end_time` via a callback (or at least may want to at some point). So it would be better not to couple into it in that way.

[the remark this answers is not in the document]
vyudu: Maybe we could use floatmax then?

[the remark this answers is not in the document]
isaacsas: Also, if the rate is zero though shouldn't it not get binned? Such reactions can't ever occur until they are updated to a non-zero rate.

src/aggregators/prioritytable.jl
```diff
@@ -312,3 +318,155 @@ function sample(pt::PriorityTable, priorities, rng = DEFAULT_RNG)
...
+    ptt = PriorityTimeTable(
+        groups, pidtogroup, times, ttgdata, zero(pidtype), zero(pidtype), maxtime)
+    # Insert priority ids into the groups
+    for (pid, time) in enumerate(times)
+        if time > maxtime
+            pidtogroup[pid] = (0, 0)
+            continue
+        end
+        insert!(ptt, pid, time)
+    end
```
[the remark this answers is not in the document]
isaacsas: Yeah, maybe it would be better to uniformly distribute them among bins.

But for now you can leave as is -- we do the same in DirectCR I think anyways, so it should probably also get reexamined.

src/aggregators/ccnrm.jl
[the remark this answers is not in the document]
vyudu: Right, anything that falls outside the current time window of the priority time table doesn't get binned, just gets passed over.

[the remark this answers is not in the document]
isaacsas: `floatmax` is fine. Just make sure it has the same type as `t`.

isaacsas:
Please revert the format commits. They are changing lots of other stuff and there is apparently a bug in JuliaFormatter anyways that is causing formats to break semantics.

isaacsas:
@vyudu is this all done / updated now?

vyudu:
Should be after these tests run.

PR #473, opened 2025-02-10
don't access callbacks till checking they exist

isaacsas:
Closes https://github.com/SciML/JumpProcesses.jl/issues/472

### DIFF
--- a/test/thread_safety.jl
+++ b/test/thread_safety.jl
@@ -1,5 +1,5 @@
 using DiffEqBase
-using JumpProcesses, OrdinaryDiffEqTsit5 
+using JumpProcesses, OrdinaryDiffEq
 using StableRNGs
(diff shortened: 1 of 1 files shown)


isaacsas:
@ChrisRackauckas I wonder if we shouldn't be doing something different with the seeding in `resetted_jump_problem`. Right now if there is a constant rate jump we reseed the `rng` stored in the aggregator. But when reinitializing the variable rate jump propensities we just use the `rng` from the `JumpProblem` (which should be the same as what is in the aggregator). So reseeding only occurs if there are constant rate jumps present. I'm not sure about how this is working at the higher level with threading, but should we instead always be reseeding just using the `rng` from in the `JumpProblem`?

ChrisRackauckas (approved):

PR #475, opened 2025-02-11
add test initial variable rate propensities are unique in ensembles

isaacsas:


ChrisRackauckas (approved):

isaacsas:
This was my worry. Everything passes with the default rng, probably because it handles different streams on different threads, but this shows there may be issues when using a custom rng with the current approach.

isaacsas:
(As here we use StableRNGs.)

isaacsas:
I'm going to merge this since this shows everything is ok with `default_rng()`, which I assume most people use, but we probably should modify this to work more generally.

PR #488, opened 2025-06-02
VR_DirectFW aggregator

sivasathyaseeelan:
## Checklist

- [x] Appropriate tests were added
- [ ] The new code follows the


src/jumps.jl
```diff
@@ -741,3 +741,19 @@ function get_jump_info_fwrappers(u, p, t, constant_jumps)
...
+    if (vjumps !== nothing) && !isempty(vjumps)
+        rates = [RateWrapper(c.rate) for c in vjumps]
+        affects! = Any[(x -> (c.affect!(x); nothing)) for c in vjumps]
+    else
+        rates = Vector{RateWrapper}()
+        affects! = Any[]
+    end
+
+    rates, affects!
+end
```
isaacsas, line 759: Why can't this reuse the previous version?

src/variable_rate.jl
```diff
@@ -402,7 +429,7 @@ function (cache::VR_DirectEventCache)(u, t, integrator)
     return cache.current_threshold
 end
 
-@generated function execute_affect!(cache::VR_DirectEventCache{T, RNG, F1, F2}, integrator, idx) where {T, RNG, F1, F2 <: Tuple}
+@generated function execute_affect!(cache::VR_DirectEventCache{T, RNG, F1, F2}, integrator, idx) where {T, RNG, F1, F2}
```
isaacsas, line 432:
```suggestion
@generated function execute_affect!(cache::VR_DirectEventCache{T, RNG, F1, F2}, integrator, idx) where {T, RNG, F1, F2}
```
No, the previous version was correct as it is only for tuples and shouldn't be used for function wrappers. For function wrappers just evaluate the `idx` entry in the affect function array directly (and add a new dispatch here for that case).

test/save_positions.jl
```diff
@@ -22,15 +22,13 @@ let
...
             urate = (u, p, t) -> 1.0, rateinterval = (u, p, t) -> 5.0)
-        jumpproblem = JumpProblem(oprob, alg, jump; vr_aggregator = VR_Direct(), dep_graph = [[1]],
-            save_positions = (false, true), rng)
-        sol = solve(jumpproblem, Tsit5(); save_everystep = false)
-        @test sol.t == [0.0, 30.0]
-
-        jumpproblem = JumpProblem(oprob, alg, jump; vr_aggregator = VR_FRM(), dep_graph = [[1]],
-            save_positions = (false, true), rng)
-        sol = solve(jumpproblem, Tsit5(); save_everystep = false)
-        @test sol.t == [0.0, 30.0]
```
isaacsas, line 33: Why are you removing existing tests?

src/variable_rate.jl
```diff
@@ -281,20 +286,34 @@ mutable struct VR_DirectEventCache{T, RNG <: AbstractRNG, F1, F2}
...
+function VR_DirectEventCache(jumps::JumpSet, ::VR_DirectFW, prob, ::Type{T}; rng = DEFAULT_RNG) where T
+    initial_threshold = randexp(rng, T)
+    vjumps = jumps.variable_jumps
+
+    # handle vjumps using tuples
+    rate_funcs, affect_funcs = get_jump_info_vr_fwrappers(vjumps, prob)
+        
+    cum_rate_sum = Vector{T}(undef, length(vjumps))
+        
+    VR_DirectEventCache{T, typeof(rng), typeof(rate_funcs), typeof(affect_funcs)}(zero(T),
```
isaacsas, line 314:
```suggestion
    VR_DirectEventCache{T, typeof(rng), typeof(rate_funcs), Any}(zero(T), 
```
The type of the affect vector  in the cache needs to be `Any`. Later we will concretize it to the specific type when we want to actually call an affect.

test/save_positions.jl
isaacsas: Ahh, I missed that. OK, sounds good!

src/variable_rate.jl
```diff
@@ -408,6 +453,10 @@ end
     end
 end
 
+@inline function execute_affect!(cache::VR_DirectEventCache{T, RNG, F1, F2}, integrator, idx) where {T, RNG, F1, F2}
```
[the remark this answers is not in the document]
sivasathyaseeelan: These are required to distinguish between Direct and DirectFW modes, else getting error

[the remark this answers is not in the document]
sivasathyaseeelan: We can't do this. Then this will be ambiguous

src/variable_rate.jl
```diff
@@ -308,8 +329,24 @@ function initialize_vr_direct_cache!(cache::VR_DirectEventCache, u, t, integrato
     nothing
 end
 
+@inline function concretize_vr_direct_affects!(cache::VR_DirectEventCache{T, RNG, F1, F2}, 
+        ::I) where {T, RNG, F1, F2, I <: DiffEqBase.DEIntegrator}
```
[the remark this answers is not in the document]
sivasathyaseeelan: I don't know why I am getting the error?
I also tried restarted

```bash
ERROR: MethodError: concretize_vr_direct_affects!(::JumpProcesses.VR_DirectEventCache{…}, ::OrdinaryDiffEqCore.ODEIntegrator{…}) is ambiguous.

Candidates:
  concretize_vr_direct_affects!(cache::JumpProcesses.VR_DirectEventCache{T, RNG, F1, F2}, ::I) where {T, RNG, F1, F2<:Tuple, I<:SciMLBase.DEIntegrator}
    @ JumpProcesses ~/Desktop/julia/JumpProcesses.jl/src/variable_rate.jl:342
  concretize_vr_direct_affects!(cache::JumpProcesses.VR_DirectEventCache, ::I) where I<:SciMLBase.DEIntegrator
    @ JumpProcesses ~/Desktop/julia/JumpProcesses.jl/src/variable_rate.jl:332

Possible fix, define
  concretize_vr_direct_affects!(::JumpProcesses.VR_DirectEventCache{…}, ::I) where {…}
```

sivasathyaseeelan:
> I suggest you just add a
> 
> ```julia
> const VR_JUMP_AGGREGATORS = (VR_FRM(), VR_Direct(), VR_DirectFW())
> ```
> 
> global constant where the types are defined. Then in the tests you can just loop over the elements of this instead of having to manually type out the tuple.

I think we can do this when we make VR_Direct as default. So we can do a final refactor

src/variable_rate.jl
```diff
@@ -402,9 +447,21 @@ function (cache::VR_DirectEventCache)(u, t, integrator)
     return cache.current_threshold
 end
 
-@generated function execute_affect!(cache::VR_DirectEventCache{T, RNG, F1, F2}, integrator, idx) where {T, RNG, F1, F2 <: Tuple}
+@generated function execute_affect!(cache::VR_DirectEventCache{T, RNG, F1, F2},
```
ChrisRackauckas, line 450: you'd always hit this one instead?

ChrisRackauckas:
> I think we can do this when we make VR_Direct as default. So we can do a final refactor

I see, that's fine. Let's merge to get benchmarks kicked off and include this in the bump that changes the default.

PR #491, opened 2025-06-22
Add buildkite runner

ChrisRackauckas:
Needs to be in master before it will launch


isaacsas:
As long as this is getting setup could we add a “correctness” group on Buildkite for longer CI tests for correctness (by running many samples)? I think this is something we’ve discussed before that is needed but hard to have via GitHub CI.

ChrisRackauckas:
I setup an exclusive queue runner for that, which for now will do nothing.

isaacsas:
Thanks!

PR #499, opened 2025-07-12
Add a GPU kernel for SSAStepper

sivasathyaseeelan:
Adds a GPU kernel implementation of Gillespie's Direct method for `SSAStepper`, so an ensemble of mass action `JumpProblem`s can be solved with one trajectory per GPU thread through `EnsembleGPUKernel`. This complements the existing `SimpleTauLeaping` kernel in the same extension.

```julia
using JumpProcesses, KernelAbstractions, Adapt, CUDA, Statistics

# SIR model as mass action jumps
maj = MassActionJump([0.1 / 1000.0, 0.01],
                     [[1 => 1, 2 => 1], [2 => 1]],
                     [[1 => -1, 2 => 1], [2 => -1, 3 => 1]])
prob = DiscreteProblem([999, 10, 0], (0.0, 250.0))
jump_prob = JumpProblem(prob, Direct(), maj)

sol = solve(EnsembleProblem(jump_prob), SSAStepper(),
            EnsembleGPUKernel(CUDABackend()); trajectories = 100_000, saveat = 10.0)

mean(sol.u[i].u[end][3] for i in 1:100_000)   # mean final recovered count
```

## Scope

Only `MassActionJump`s are supported. A mass action propensity is fully described by its stoichiometry and rate constant, so it can be evaluated on the device from data alone. A `ConstantRateJump` cannot: its rate is an arbitrary Julia closure and its `affect!` mutates an integrator, so there is nothing a kernel can reconstruct. Rather than guessing at those functions, unsupported problems are rejected with an error that explains what to do instead.

Also currently unsupported, each with an explicit error: callbacks, and a `prob_func` (the reaction data is uploaded once and shared by every thread).

## How it works

* **Stoichiometry transfer.** `MassActionJump` stores stoichiometry as a `Vector{Vector{Pair{Int,Int}}}`, which has no device representation. Both the reactant and net stoichiometries are flattened into species/coefficient arrays indexed by CSR-style offsets.
* **Propensities.** Evaluated exactly as `evalrxrate` does, including the falling factorial `x(x-1)…(x-k+1)` for higher order reactions and the reactant depletion check, with the combinatoric prefactor taken from `scaled_rates`.
* **No scratch memory.** Trajectory state lives in registers as an `SVector`, and reaction selection recomputes propensities instead of caching them, so the kernel allocates no per-thread buffers in global memory.
* **RNG.** Each thread advances its own `splitmix64` stream, seeded by mixing the trajectory index, so draws within a trajectory are sequential and streams do not overlap between trajectories.
* **Saving.** `saveat` is required, and the solution is sampled on that grid. The number of jumps in an SSA trajectory is random, so there is no trajectory length to allocate up front; a grid gives every trajectory the same known output size. The grid is shared by all trajectories and doubles as the `t` vector of each solution. Writes are laid out trajectory-major so they coalesce across threads.

Because sampling happens only on the grid, each returned solution contains exactly the grid points, as if the problem had been built with `save_positions = (false, false)`. The aggregator in the `JumpProblem` is ignored, since the kernel always runs the Direct method, and randomness comes from the device generator seeded by `seed` rather than the problem's `rng`.

## Testing

`test/gpu/ssa.jl` runs in the existing `CUDA` test group. It checks the kernel against exact analytic results where they are available (linear death, zero order, and the binomial law for the number of survivors, the last of which also confirms the per-trajectory streams are independent — a shared stream would collapse the sample variance) and against the serial SSA otherwise. It also covers second and third order propensities, conservation laws, extinction into an absorbing state, `saveat`/`save_start`/`save_end` handling, seed reproducibility, and each rejected-input path.

The suite was additionally run on real GPU hardware through Metal (69 assertions, all passing) as a cross-check that the kernel compiles and behaves correctly off the CPU backend.

## Checklist


test/gpu/ssa_test.jl
```diff
@@ -0,0 +1,29 @@
...
+prob = DiscreteProblem([10.0], (0.0, 3.0))
+jump_prob = JumpProblem(prob, Direct(), jump, jump2; rng = rng)
+
+integrator = init(jump_prob, SSAStepper())
+step!(integrator)
+integrator.u[1]
+
+# test different saving behaviors
+
+sol = solve(EnsembleProblem(jump_prob), SSAStepper(), EnsembleGPUKernel(CUDABackend()),
```
ChrisRackauckas, line 27: Make the same test in the CPU tests just using the CPU backend to kernel abstractions.

isaacsas:
@sivasathyaseeelan do you plan to update per @ChrisRackauckas comments?

ext/ssa_stepper.jl
```diff
@@ -0,0 +1,416 @@
...
+# the device from data alone. `ConstantRateJump`s cannot: their rates are
+# arbitrary Julia closures and their `affect!`s mutate an integrator, neither of
+# which can be reconstructed from a device kernel.
+#
+# Two properties of mass action jumps are what make the kernel simple: the
+# combinatoric prefactors are already folded into `scaled_rates`, and the
+# propensities depend on neither `p` nor `t`. A trajectory can therefore be
+# advanced using nothing but the stoichiometry tables.
+###############################################################################
+
```
ChrisRackauckas:
```suggestion
```

ext/ssa_stepper.jl
```diff
@@ -0,0 +1,416 @@
...
+end
+
+###############################################################################
+# Device RNG.
+#
+# splitmix64: a small counter-based generator with a full 2^64 period. Each
+# thread keeps its state in a register and advances it on every draw, so a
+# trajectory consumes one long stream rather than re-deriving values from a
+# fixed seed (which correlates successive draws).
+###############################################################################
```
ChrisRackauckas: this should use an actual device rng

ChrisRackauckas:
Almost good just fix the rng, i.e. use the device rng

test/gpu/ssa_test.jl
sivasathyaseeelan: Done. The test body is now in `test/ssa_kernel_tests.jl` as one function, `run_ssa_kernel_tests(backend, nsims)`.

Two files call it:

* `test/gpu/ssa.jl` calls it with `CUDABackend()`, in the `CUDA` group.
* `test/kernelabstractions_ssa.jl` calls it with the KernelAbstractions `CPU()` backend, in the `InterfaceII` group.

The two runs use the same assertions. `KernelAbstractions` and `Adapt` are now in the test target. The CPU run uses fewer trajectories, because it has no GPU parallelism.

ext/ssa_stepper.jl
```diff
@@ -0,0 +1,324 @@
...
+@inline function exponential_rand(lambda::T, seed::UInt64, idx::Int64) where T
+    seed = (1103515245 * (seed ⊻ UInt64(idx)) + 12345) % 2^31
+    u = Float64(seed) / 2^31
+    return -log(u) / lambda
+end
+
+@inline function uniform_rand(seed::UInt64, idx::Int64)
+    seed = (1103515245 * (seed ⊻ UInt64(idx)) + 12345) % 2^31
+    return Float64(seed) / 2^31
+end
```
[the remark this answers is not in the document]
sivasathyaseeelan: Done. The hand-written generator is removed.

The kernel now calls `rand` and gets the generator of the backend. Please see my reply on the RNG comment for the details.

ext/ssa_stepper.jl
```diff
@@ -0,0 +1,324 @@
...
+                         depend_indices, depend_starts, depend_counts, 
+                         us, ts, current_u_buf, rate_cache_buf, seed_val, max_steps;
+                         ndrange=trajectories)
+    synchronize(backend)
+
+    return ts, us
+end
+
+# Modified Gillespie Direct kernel for arbitrary dependencies
+@kernel function gillespie_direct_kernel(@Const(prob_data), @Const(num_jumps),
```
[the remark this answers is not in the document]
sivasathyaseeelan: That code is removed in the rewrite.

The old version put a `vectorized_gillespie_direct` wrapper between `__solve` and the kernel. The new version dispatches on `alg::SSAStepper` in `SciMLBase.__solve` and calls the kernel directly.

`SSAIntegrator` itself cannot run on a device: it is mutable, it holds callbacks, and it pushes to its solution arrays. The kernel therefore does the Direct method with the state in registers as an `SVector`. It keeps the `saveat` behaviour of `SSAStepper`: a grid point strictly before the next jump gets the pre-jump state, so the sampled path agrees with the serial solver.

ext/ssa_stepper.jl
sivasathyaseeelan: Done. The header block is removed.

ext/ssa_stepper.jl
sivasathyaseeelan: Done. The kernel now calls `rand(RT)`, which resolves to the generator of the backend. This is the same generator that `PoissonRandom.PassthroughRNG` forwards to for the tau-leaping kernel.

I ask for the value in the working precision. A `Float32` problem then does not pull `Float64` math onto the device. This is necessary: `randexp(PassthroughRNG())` returns `Float64`, and a first attempt with it failed to compile for a 32-bit backend with `unsupported use of double value`.

The waiting time is `-log1p(-rand(RT))`. This gives `0` and not `Inf` if `rand` returns `0`, and it needs no branch.

`seed` now seeds the ambient generator, so it makes a run repeatable on `CPU()`. To seed a GPU backend, use the `seed!` of that backend. I removed the two cross-run reproducibility assertions. Stream independence is still covered: the test compares the variance of the survivor count against the exact `Binomial(n, exp(-ct))` value, which would collapse if the trajectories shared one stream.

sivasathyaseeelan:
@ChrisRackauckas addressed the review comments (the kernel now draws from the backend device `RNG` via rand, and the same tests run on the KernelAbstractions `CPU()` backend as well as `CUDA`)

PR #513, opened 2025-08-09
SimpleAdaptiveTauLeaping and SimpleImplicitTauLeaping

sivasathyaseeelan:
```julia
using JumpProcesses, DiffEqBase
using Test, LinearAlgebra
using StableRNGs, Plots
rng = StableRNG(12345)

function regular_rate(out, u, p, t)
    out[1] = (0.1 / 1000.0) * u[1] * u[2]
    out[2] = 0.01u[2]
end

function regular_c(dc, u, p, t, mark)
    dc[1, 1] = -1
    dc[2, 1] = 1
    dc[2, 2] = -1
    dc[3, 2] = 1
end

dc = zeros(3, 2)

rj = RegularJump(regular_rate, regular_c, dc; constant_c = true)
jumps = JumpSet(rj)

prob = DiscreteProblem([999.0, 1.0, 0.0], (0.0, 250.0))
jump_prob = JumpProblem(prob, Direct(), rj; rng = rng)
sol = solve(jump_prob, ImplicitTauLeaping())

const _dc = zeros(3, 2)
dc[1, 1] = -1
dc[2, 1] = 1
dc[2, 2] = -1
dc[3, 2] = 1

function regular_c(du, u, p, t, counts, mark)
    mul!(du, dc, counts)
end

rj = RegularJump(regular_rate, regular_c, 2)
jumps = JumpSet(rj)
prob = DiscreteProblem([999, 1, 0], (0.0, 250.0))
jump_prob = JumpProblem(prob, Direct(), rj; rng = rng)
sol = solve(jump_prob, ImplicitTauLeaping())
plot(sol)
```

Screenshot from 2025-07-28 18-25-31

```julia
using JumpProcesses, DiffEqBase
using Test, LinearAlgebra
using StableRNGs, Plots
rng = StableRNG(12345)

# Parameters
c1 = 1.0      # S1 -> 0
c2 = 10.0     # S1 + S1 <- S2
c3 = 1000.0   # S1 + S1 -> S2
c4 = 0.1      # S2 -> S3
p = (c1, c2, c3, c4)

regular_rate_implicit = (out, u, p, t) -> begin
    out[1] = p[1] * u[1]          # S1 -> 0
    out[2] = p[2] * u[2]          # S1 + S1 <- S2
    out[3] = p[3] * u[1] * (u[1] - 1) / 2  # S1 + S1 -> S2
    out[4] = p[4] * u[2]          # S2 -> S3
end

regular_c_implicit = (dc, u, p, t, counts, mark) -> begin
    dc .= 0.0
    dc[1] = -counts[1] - 2 * counts[3] + 2 * counts[2]  # S1: -decay - 2*forward + 2*backward
    dc[2] = counts[3] - counts[2] - counts[4]           # S2: +forward - backward - decay
    dc[3] = counts[4]                                   # S3: +decay
end

u0 = [10000.0, 0.0, 0.0]  # S1, S2, S3
tspan = (0.0, 4.0)

# Create JumpProblem with proper parameter passing
prob_disc = DiscreteProblem(u0, tspan, p)
rj = RegularJump(regular_rate_implicit, regular_c_implicit, 4)
jump_prob = JumpProblem(prob_disc, Direct(), rj; rng=StableRNG(12345))
sol = solve(jump_prob, ImplicitTauLeaping())
plot(sol)
```

Screenshot from 2025-07-28 18-13-49

## Checklist


Project.toml
```diff
@@ -13,6 +13,7 @@ FunctionWrappers = "069b7b12-0de2-55c6-9aab-29f3d0a68a2e"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
+NonlinearSolve = "8913a72c-1f9b-4ce2-8d82-65094dcecaec"
```
isaacsas, line 16: This is a pretty heavy addition. Can the code that needs it be moved to an extension? For most jump problems this shouldn’t be needed.

ChrisRackauckas: It should just be SimpleNonlinearSolve...

isaacsas:
@sivasathyaseeelan can you make the correctness test more strict. Compare against `Direct`, which will probably require a smaller timestep for the leaping methods, and calculate the maximum error over time (say every second) instead of just at the final time. 

I’m reading through the references to refresh my memory on how the adaptive stepping works, and will try to finish that tomorrow and give you some more feedback afterwards.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,376 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+            mu[i] += nu[i, j] * rate_cache[j]
+            sigma2[i] += nu[i, j]^2 * rate_cache[j]
+        end
+        gi = compute_gi(u, nu, i, rate, rate_cache, p, t)
+        bound = max(epsilon * u[i] / gi, 1.0)
+        mu_term = abs(mu[i]) > 0 ? bound / abs(mu[i]) : Inf
+        sigma_term = sigma2[i] > 0 ? bound^2 / sigma2[i] : Inf
+        tau = min(tau, mu_term, sigma_term)
+    end
+    return max(tau, 1e-10)
```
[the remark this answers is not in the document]
sivasathyaseeelan: yep dtmin make sense. I will add that

Project.toml
```diff
@@ -19,6 +19,7 @@ RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
 SciMLBase = "0bca4576-84f4-4d90-8ffe-ffa030f20462"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
+SimpleNonlinearSolve = "727e6d20-b764-4bd8-a329-72de5adea6c7"
```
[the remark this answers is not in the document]
sivasathyaseeelan: we dont require SimpleNonlinearSolve for SimpleAdaptiveTauLeaping. So removed that for this PR

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,376 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+
+    u = [copy(u0)]
+    t = [tspan[1]]
+    rate_cache = zeros(Float64, numjumps)
+    counts = zeros(Int, numjumps)
+    du = similar(u0)
+    t_end = tspan[2]
+    epsilon = alg.epsilon
+
+    nu = compute_stoichiometry(c, u0, numjumps, p, t[1])
```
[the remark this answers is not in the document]
sivasathyaseeelan: I made it a function so that we can use it for both SimpleAdaptiveTauLeaping and SimpleImplicitTauLeaping

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,110 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+            c(du, u_prev, p, t_prev, counts_temp, nothing)
+            nu[:, j] = du
+        end
+        rate(rate_cache, u_prev, p, t_prev)
+        tau = compute_tau_explicit(u_prev, rate_cache, nu, hor, p, t_prev, epsilon, rate, dtmin)
+        tau = min(tau, t_end - t_prev)
+        counts .= pois_rand.(rng, max.(rate_cache * tau, 0.0))
+        c(du, u_prev, p, t_prev, counts, nothing)
+        u_new = max.(u_prev + du, 0)
+        if any(u_new .< 0)
```
ChrisRackauckas, line 143: flip these

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,110 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+            tau /= 2
```
ChrisRackauckas, line 144: give a reference that this is okay

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
 end
 
+struct SimpleAdaptiveTauLeaping <: DiffEqBase.DEAlgorithm
+    epsilon::Float64  # Error control parameter
+end
+
+SimpleAdaptiveTauLeaping(; epsilon=0.05) = SimpleAdaptiveTauLeaping(epsilon)
+
+function compute_gi(u, nu, hor, i)
+    max_order = 1.0
```
isaacsas, line 60: Why not just make this an integer? It should just be an integer right given that `hor` is integer?

Generally try not to assume floating point types in your code. They should be inferred from the types of `t` or the type of the elements of `u` as appropriate.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+
+    # Compute initial stoichiometry and HOR
+    nu = zeros(Int, length(u0), numjumps)
+    counts_temp = zeros(Int, numjumps)
+    for j in 1:numjumps
+        fill!(counts_temp, 0)
+        counts_temp[j] = 1
+        c(du, u0, p, t[1], counts_temp, nothing)
+        nu[:, j] = du
+    end
```
isaacsas, line 122: As I mentioned in the other PR, I suggest switching to the input being `MassActionJump`s. Then you can get rid of most of this stoichiometry-related code. It will also have a big performance benefit since you won't need to recalculate `nu` every timestep (which is likely to be very expensive).

isaacsas: Once you've made the switch to `MassActionJump`s you can get the stoichiometry for the `i`th reaction from `net_stoch[i]`. It only needs to be extracted once as it doesn't change during a simulation.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+    hor = zeros(Int, size(nu, 2))
+    for j in 1:size(nu, 2)
+        hor[j] = sum(abs.(nu[:, j])) > maximum(abs.(nu[:, j])) ? 2 : 1
```
isaacsas, line 125: Doesn't this need to be calculated based on the substrate stoichiometry, not the net stoichiometry? (The order of a reaction is determined by the substrates coefficients, i.e. for $m_1 S_1 + m_2 S_2 --> n_1 S_1 + n_2 S_2$ the order is $m_1 + m_2$.) If you use `MassActionJump`s you can get these from the `reactant_stoch` field.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+    end
+
+    saveat_times = isnothing(saveat) ? Float64[] : saveat isa Number ? collect(range(tspan[1], tspan[2], step=saveat)) : collect(saveat)
```
isaacsas, line 128:
```suggestion
    saveat_times = isnothing(saveat) ? Vector{typeof(t))() : saveat isa Number ? collect(range(tspan[1], tspan[2], step=saveat)) : collect(saveat)
```
This is too long to read / parse, expand it out with an outer if/else block. Try not to nest ternary operators as it gets hard to read the code.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+    rj = jump_prob.regular_jump
+    rate = rj.rate
+    numjumps = rj.numjumps
+    c = rj.c
+    u0 = copy(prob.u0)
+    tspan = prob.tspan
+    p = prob.p
+
+    u = [copy(u0)]
+    t = [tspan[1]]
```
isaacsas, line 107: Make these the current `u` and current `t`, and have separate vectors `usave` and `tsave` or such for the history. That is more consistent with SciML naming.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+    while t[end] < t_end
+        u_prev = u[end]
+        t_prev = t[end]
+        # Recompute stoichiometry
+        for j in 1:numjumps
+            fill!(counts_temp, 0)
+            counts_temp[j] = 1
+            c(du, u_prev, p, t_prev, counts_temp, nothing)
+            nu[:, j] = du
+        end
```
isaacsas, line 140: See above. This should no longer be necessary if you switch to `MassActionJump`s. Currently this is a very expensive operation that I would expect makes the method too slow to be practically useful.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+        tau = min(tau, t_end - t_prev)
+        if !isempty(saveat_times)
+            if save_idx <= length(saveat_times) && t_prev + tau > saveat_times[save_idx]
+                tau = saveat_times[save_idx] - t_prev
+            end
+        end
+        counts .= pois_rand.(rng, max.(rate_cache * tau, 0.0))
+        c(du, u_prev, p, t_prev, counts, nothing)
+        u_new = u_prev + du
+        if any(u_new .< 0)
```
isaacsas, line 152:
```suggestion
        if any(<(0), u_new)
```
This avoids allocations.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+            # Halve tau to avoid negative populations, as per Cao et al. (2006, J. Chem. Phys., DOI: 10.1063/1.2159468)
+            tau /= 2
+            continue
+        end
+        u_new = max.(u_new, 0)  # Ensure non-negative states
```
isaacsas, line 157: Don't allocate a new vector. Using a working vector defined outside the loop.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+        push!(u, u_new)
+        push!(t, t_prev + tau)
```
isaacsas, line 159: If there is a `saveat` shouldn't you only save at those times and not other times? More generally try to decouple the current `u` and `t` in the timesteping from the vectors of saved output.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+    if !isempty(saveat_times)
+        t_out = saveat_times
+        u_out = [u[end]]
+        for t_save in saveat_times
+            idx = findlast(ti -> ti <= t_save, t)
+            push!(u_out, u[idx])
+        end
+        t = t_out
+        u = u_out[2:end]
+    end
```
isaacsas, line 175: Why is this needed? Aren't you directly stepping to `saveat` times?

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+SimpleAdaptiveTauLeaping(; epsilon=0.05) = SimpleAdaptiveTauLeaping(epsilon)
+
+function compute_gi(u, nu, hor, i)
+    max_order = 1.0
+    for j in 1:size(nu, 2)
+        if abs(nu[i, j]) > 0
+            max_order = max(max_order, Float64(hor[j]))
+        end
+    end
+    return max_order
```
isaacsas, line 66: This isn't necessarily the correct value for $g_i$ though when `hor[i] > 1`. If `hor[i] == 2` you need to precalculate if any reaction for which species `i` is a substrate has the substrate stoichoimetry of 2 or more (using `reactant_stoch` for example). Then you need to use the formulate for the updated `g_i` in case (II). Similar for if `hor[i] ==3`. 

We don't need to support if `hor[i] > 3`, but the solver should probably error if it detects this is the case.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+end
+
+function compute_tau_explicit(u, rate_cache, nu, hor, p, t, epsilon, rate, dtmin)
+    rate(rate_cache, u, p, t)
+    mu = zeros(length(u))
+    sigma2 = zeros(length(u))
```
isaacsas, line 72: Avoid allocations. Can't these just be scalar variables within the loop over `i`?

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+    tau = Inf
+    for i in 1:length(u)
+        for j in 1:size(nu, 2)
+            mu[i] += nu[i, j] * rate_cache[j]
+            sigma2[i] += nu[i, j]^2 * rate_cache[j]
```
isaacsas, line 77: These can just be written as dot products using views of `nu`.

test/regular_jumps.jl
```diff
@@ -1,41 +1,121 @@
...
+
+    # Compute mean trajectories at t = 0, 1, ..., 250
+    t_points = 0:1.0:250.0
+    mean_direct_S = [mean(sol_direct[i](t)[3] for i in 1:Nsims) for t in t_points]
+    mean_simple_S = [mean(sol_simple[i](t)[3] for i in 1:Nsims) for t in t_points]
+    mean_adaptive_S = [mean(sol_adaptive[i](t)[3] for i in 1:Nsims) for t in t_points]
+
+    for i in 1:251
+        @test isapprox(mean_direct_S[i], mean_simple_S[i], rtol=0.10)
+        @test isapprox(mean_direct_S[i], mean_adaptive_S[i], rtol=0.10)
```
isaacsas, line 119: This is too loose a bound. At least decrease it to 5%.

isaacsas (commented):
General comments:
1. Please add comments citing the formulas you are using from the original paper for the leap selection.
2. Please revise with an eye towards making this non-allocating during the solution timestepping, see the examples I've highlighted but keep in mind they are not all the cases of extraneous allocations.

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+    epsilon::Float64  # Error control parameter
+end
+
+SimpleAdaptiveTauLeaping(; epsilon=0.05) = SimpleAdaptiveTauLeaping(epsilon)
+
+function compute_gi(u, nu, hor, i)
+    max_order = 1.0
+    for j in 1:size(nu, 2)
+        if abs(nu[i, j]) > 0
+            max_order = max(max_order, Float64(hor[j]))
```
ChrisRackauckas, line 63:
```suggestion
            max_order = max(max_order, float(hor[j]))
```

src/simple_regular_solve.jl
```diff
@@ -50,6 +50,136 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
         interp = DiffEqBase.ConstantInterpolation(t, u))
 end
 
+struct SimpleAdaptiveTauLeaping <: DiffEqBase.DEAlgorithm
+    epsilon::Float64  # Error control parameter
```
ChrisRackauckas, line 54:
```suggestion
struct SimpleAdaptiveTauLeaping{T <: AbstractFloat} <: DiffEqBase.DEAlgorithm
    epsilon::T  # Error control parameter
```

src/simple_regular_solve.jl
sivasathyaseeelan: maj is used now

src/simple_regular_solve.jl
sivasathyaseeelan: this can be done

src/simple_regular_solve.jl
sivasathyaseeelan: yes this can be done

src/simple_regular_solve.jl
sivasathyaseeelan: yes can be done

src/simple_regular_solve.jl
```diff
@@ -62,6 +81,164 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+                    gi = 2 / stoch + 1 / (stoch - 1)
+                    max_gi = max(max_gi, ceil(Int, gi))
+                elseif stoch == 1
+                    max_gi = max(max_gi, hor[j])
+                end
+            end
+        end
+    end
+    return max_gi
+end
```
isaacsas, line 114: I haven't reviewed the full PR carefully, but this function still looks wrong to me. Look at Section IV of the paper at the text between (27) and (28).  `x_i` there is not `stoch` it is the `i`th species (i.e. `u_i`).

Also, you should be precalculating if **any** reaction with `HOR[i] == 2` has species `i` with substrate stoichiometry of 2 (and the related conditions when `HOR[i] == 3`). This is determined from `reactant_stoch` not the net stoichiometry matrix `nu`. 

If I can make a suggestion, perhaps it makes sense for you to go back and work on finishing the GPU method PR. It seems like you are having some conceptual issues with the paper here, whereas that PR just involves some basic code modifications to get finished and so should be an easier task. Once that is done and merged perhaps you can more carefully read this paper, particularly Section IV, and then return to this PR?

src/simple_regular_solve.jl
```diff
@@ -61,6 +80,215 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+                elseif nu_ij == 3
+                    if u[i] > 2  # Ensure x_i - 2 > 0
+                        gi = 3 + 1 / (u[i] - 1) + 2 / (u[i] - 2)
+                        max_gi = max(max_gi, ceil(Int64, gi))
+                    end
+                end
+            end
+        end
+    end
+    return max_gi
```
isaacsas, line 157: This is getting better, but there are still some issues. 
1. Just initialize `max_gi = zero(typeof(t))` and get rid of all these `ceil`'s and conversions to integers. WIth the correction formulas `g_i` can be a floating point type so I think it makes sense to just give it the same type as the time variable.
2. You are currently doing `num_species * num_rxs` work per timestep in computing all these `g_i`, when it should only be `num_species` work. You need to *precalculate* when initializing the solver `hor_order`, where `hor_order[i]` is the maximum substrate order of species `i` over all reactions. Then you use this to determine which `g_i` case you are in. For example, if `HOR(i) == 2` and `hor_order[i] == 1` you just set `g_i = 2`, but if `HOR(i) == 2` and `hor_order[i] == 2` you use the formula that involves `u[i]`.

src/simple_regular_solve.jl
```diff
@@ -61,6 +80,215 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
...
+            mu += nu[i, j] * rate_cache[j] # Equation (9a)
+            sigma2 += nu[i, j]^2 * rate_cache[j] # Equation (9b)
+        end
+        gi = compute_gi(u, reaction_conditions, i)
+        bound = max(epsilon * u[i] / gi, 1.0) # max(epsilon * x_i / g_i, 1)
+        mu_term = abs(mu) > 0 ? bound / abs(mu) : Inf # First term in equation (8)
+        sigma_term = sigma2 > 0 ? bound^2 / sigma2 : Inf # Second term in equation (8)
+        tau = min(tau, mu_term, sigma_term) # Equation (8)
+    end
+    return max(tau, dtmin)
```
isaacsas, line 181: We need a test for the case that all the rates are zero to make sure the returned `tau` is reasonable. I suspect your code isn't quite robust when rates start becoming zero.

isaacsas (changes requested):
@sivasathyaseeelan getting better. I've only reviewed the tau-selection code -- once that seems to be the right algorithm / approach I'll give you a more thorough review.

src/simple_regular_solve.jl
sivasathyaseeelan: Now `compute_gi` is in O(1) time complexity

src/simple_regular_solve.jl
sivasathyaseeelan: Resolved this case

src/simple_regular_solve.jl
```diff
@@ -61,6 +80,213 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleTauLeaping;
         interp = DiffEqBase.ConstantInterpolation(t, u))
 end
 
+function compute_hor(reactant_stoch, numjumps)
+    # Compute the highest order of reaction (HOR) for each reaction j, as per Cao et al. (2006), Section IV.
+    # HOR is the sum of stoichiometric coefficients of reactants in reaction j.
+    hor = zeros(Int, numjumps)
+    for j in 1:numjumps
+        order = sum(stoch for (spec_idx, stoch) in reactant_stoch[j]; init=0)
```
isaacsas, line 88:
```suggestion
        order = sum(stoch for (spec_idx, stoch) in reactant_stoch[j])
```
`sum` can figure this out and avoid type assumptions.

isaacsas (changes requested):

PR #518, opened 2025-08-22
drop Markdown dep

isaacsas:
I'm not sure when this got pulled in, but it seems extraneous and unused. (Note it is used in the docs.)


ChrisRackauckas:
Probably the `@doc` macro at some point. But I don't see it in here anymore

PR #519, opened 2025-08-22
Cleanup exports and add constant rate jumps to JumpProblem

isaacsas:
This 

1. Cleans up the exports to live right after their associated include.
2. Adds a field to store `ConstantRateJump`s within `JumpProblem`. This will allow us to avoid instantiating constant rate aggregators, like `Direct`, even when one is using a tau-leaping method. 

As a followup I plan to add a `Leaping()` aggregator that can be selected when one wants to use a pure leaping solver. (Later we can add `HybridLeaping()` for mixed cases.) Specifically, I want to make it so that if you pass `Leaping()` as the aggregator when you want a pure leaping `JumpProblem`, then even if you have non-`RegularJump`s it will still allow `JumpProblem` construction. The leaping timestepper will then check if there are jump types it can/can't support and error if appropriate. For example, in `SimpleTauLeaping` we can just error if non-`RegularJump`s are present, but ultimately we could generate a `RegularJump` that wraps the individual jumps (but maybe warn that it will be less performant). The advantage of this approach is we then can have access to individual jumps for leaping methods where this is potentially useful / needed (for example, adaptive time-stepping based on the Petzold methods, or splitting approaches).

*NOTE* this is technically breaking.


isaacsas:
Ultimately we could also add `HybridLeaping` type aggregators for static and/or dynamic mixing of jumps.

isaacsas:
@ChrisRackauckas ok if I merge this with you? It means we need to hold off on a release till the WIP PR is done / merged (https://github.com/SciML/JumpProcesses.jl/pull/520).

isaacsas:
Actually, adding a new type parameter isn't breaking right, since users should be using the outer constructor?

ChrisRackauckas (approved):
Seems reasonable enough

ChrisRackauckas:
> Actually, adding a new type parameter isn't breaking right, since users should be using the outer constructor?

Yes, and if you only supply a subset of type parameters dispatching still works (that was made so it would be less breaking to change this stuff).

> For example, in SimpleTauLeaping we can just error if non-RegularJumps are present, but ultimately we could generate a RegularJump that wraps the individual jumps (but maybe warn that it will be less performant). The advantage of this approach is we then can have access to individual jumps for leaping methods where this is potentially useful / needed (for example, adaptive time-stepping based on the Petzold methods, or splitting approaches).

Yeah... I think the real thing in the end is just to require the redundant information in different forms of functions and have the solvers use them effectively, and just point the users to Catalyst.jl so everything is generated. Otherwise there's tons of little "we need the aggregated function here, the individual ones here, ..." and to do this efficiently you kind of want both without forcing the user to give you both, which just has to have a symbolic interface.

PR #524, opened 2025-09-05
SimpleAdaptiveTauLeaping solver

sivasathyaseeelan:
Adds `SimpleAdaptiveTauLeaping`, which switches between an explicit and an implicit tau-leaping step according to whether the system currently looks stiff.

Explicit tau-leaping is efficient on non-stiff systems but is limited by the fastest reaction, while an implicit step lifts that limit at the cost of a nonlinear solve per step. This solver measures stiffness at each step and pays for the implicit step only when it is needed, following Cao et al. (2007).

```julia
using JumpProcesses

maj = MassActionJump([100.0, 100.0, 0.1], [[1 => 1], [2 => 1], [1 => 1]],
                     [[1 => -1, 2 => 1], [1 => 1, 2 => -1], [1 => -1, 3 => 1]])
prob = DiscreteProblem([100.0, 100.0, 0.0], (0.0, 1.0))
jprob = JumpProblem(prob, PureLeaping(), maj)

sol = solve(jprob, SimpleAdaptiveTauLeaping())                                       # propensity-ratio test
sol = solve(jprob, SimpleAdaptiveTauLeaping(eigenvalue_check = true))                # eigenvalue test
sol = solve(jprob, SimpleAdaptiveTauLeaping(implicit_alg = SimpleTrapezoidalLeaping()))
```

Stiffness is judged from the spread of the propensities by default, or from the eigenvalue ratio of the drift Jacobian when `eigenvalue_check` is set. When a step is taken implicitly the tau-selection tolerance is relaxed by `implicit_epsilon_factor`, since that step is not restricted by the fast timescale it damps.

## Rebased onto the implicit solvers from #643

The implicit step is no longer defined here. It is whichever algorithm is passed as `implicit_alg`, either `SimpleImplicitTauLeaping` or `SimpleTrapezoidalLeaping`, so the nonlinear solve and its residual come from #643 rather than being duplicated. The earlier revision of this PR carried its own `AbstractImplicitSolver` / `NewtonImplicitSolver` / `TrapezoidalImplicitSolver` hierarchy; those are gone, and this PR now adds no new dependencies, since master already has `ADTypes` and `SimpleNonlinearSolve`.

Step-size selection, mass action rate construction and `saveat` handling are shared with `SimpleExplicitTauLeaping`, and `validate_pure_leaping_inputs` is widened to cover the new algorithm rather than gaining a second identical method.

## Two bugs fixed from the earlier revision

* **The eigenvalue-based stiffness check took eigenvalues of a non-square matrix.** It built the Jacobian of the propensities, which is `numjumps × numspecies`, and passed it straight to `eigvals`. That only works when a model happens to have as many reactions as species; an SEIR model (3 reactions, 4 species) raises `DimensionMismatch: matrix is not square`. The stiffness of the process is a property of the drift `dX/dt = Σⱼ νⱼ aⱼ(X)`, so the check now uses the square Jacobian `nu * da/dX`.
* **A rejected implicit step could spin forever.** The retry path halved `tau` and then `continue`d, but `tau` is recomputed at the top of each iteration, so the halved value was discarded. An explicit retry survives this because it redraws its Poisson counts; the implicit solve is deterministic in `(u_current, tau)` and would fail identically on every pass. The bound is now carried across iterations, and a step that still fails at `dtmin` is reported as an error.

## Testing

The SIR and SEIR mean-trajectory comparisons are extended to `SimpleAdaptiveTauLeaping` with both implicit formulations, checked against the exact SSA solution at every saved timepoint. A dedicated test covers the eigenvalue-based stiffness check on an SEIR model — the case with more species than reactions that the old Jacobian could not handle — over both implicit algorithms and both stiffness tests, asserting the trajectory stays non-negative and conserves total population.

## Checklist


src/simple_regular_solve.jl
```diff
@@ -475,6 +571,290 @@ function DiffEqBase.solve(jump_prob::JumpProblem, alg::SimpleExplicitTauLeaping;
...
+    (; u_current, rate_new, rate_current, nu, p, t, tau, rate, numjumps, solver) = params
+
+    rate(rate_new, u_new, p, t + tau)
+    resid .= u_new .- u_current
+
+    if solver isa NewtonImplicitSolver
+        for j in 1:numjumps
+            for spec_idx in axes(nu, 1)
+                resid[spec_idx] -= nu[spec_idx, j] * rate_new[j] * tau
+            end
```
ChrisRackauckas: why does the adaptive explicit method have newton solves?

ChrisRackauckas:
Rebase this onto master.

src/simple_regular_solve.jl
sivasathyaseeelan: It should not, and it no longer does. This PR now defines no implicit machinery of its own.

You pointed at the copy of `implicit_equation!` that this branch carried, together with its own `AbstractImplicitSolver`, `NewtonImplicitSolver` and `TrapezoidalImplicitSolver` types. That duplication is removed. The branch is rebased on master, and the implicit step now comes from #643. A diff against master adds no `implicit_equation!`, no `solve_implicit` and none of those solver types.

On the name: this is not an explicit method. `SimpleExplicitTauLeaping` already adapts `tau`, so "adaptive" is the wrong word for what is new here. This algorithm decides at each step whether the system is stiff. If it is not stiff, it takes the explicit step. If it is stiff, it takes an implicit step, and that step is the algorithm given as `implicit_alg`, either `SimpleImplicitTauLeaping` or `SimpleTrapezoidalLeaping`. The Newton solve is therefore the one from #643, and only stiff steps pay for it. The method is Cao et al. (2007).

If you prefer a clearer name, I am happy to rename it. `SimpleAutoTauLeaping` would match the `Auto*` convention that OrdinaryDiffEq uses for stiffness switching, for example `AutoTsit5(Rosenbrock23())`. Please tell me which you want.

The PR now adds these definitions only:

* `SimpleAdaptiveTauLeaping` and its keyword constructor
* `compute_tau_implicit`, which relaxes the tau tolerance for a stiff step
* `compute_drift_jacobian` and `is_stiff`, for the stiffness test
* the step loop and its `solve` method

sivasathyaseeelan:
Done, rebased on master.

The rebase removes the duplicated implicit machinery. The adaptive solver now holds an `implicit_alg` field and calls the `solve_implicit` from #643, so the residual and the nonlinear solve are shared, not copied. The diff drops from 382 to 323 added lines, and it no longer touches `Project.toml`, because master already has `ADTypes` and `SimpleNonlinearSolve`.

Two bugs from the earlier revision are also fixed:

* The eigenvalue stiffness test called `eigvals` on the Jacobian of the propensities. That matrix is `numjumps x numspecies`, so it is not square unless a model has as many reactions as species, and an SEIR model raises `DimensionMismatch: matrix is not square`. The test now uses the square drift Jacobian `nu * da/dX`.
* A rejected implicit step halved `tau` and then continued, but `tau` was recomputed at the top of the loop, so the smaller value was lost. An explicit retry redraws its Poisson counts and recovers; the implicit solve is deterministic in `(u_current, tau)` and repeated the same failure. The loop now carries that bound, and a step that still fails at `dtmin` raises an error.

ChrisRackauckas (approved):

PR #580, opened 2026-04-22
Bump DiffEqBase / OrdinaryDiffEq / OrdinaryDiffEqCore compat to include v7/v4

ChrisRackauckas-Claude:
## Summary

Widen compat so JumpProcesses resolves alongside the in-development v7 OrdinaryDiffEq stack (`lib/DiffEqBase` 7.0.0 + `lib/OrdinaryDiffEqCore` 4.0.0 in the OrdinaryDiffEq monorepo):

- `DiffEqBase: "6.192"` → `"6.192, 7"`
- `OrdinaryDiffEq: "6"` → `"6, 7"`
- `OrdinaryDiffEqCore: "3"` → `"3, 4"`

Version bump 9.25.1 → 9.26.0.

## Motivation

On SciML/OrdinaryDiffEq.jl#3488 (v7 release branch), the `test (ModelingToolkit, …)` matrix and various indirect paths fail at resolve time. Even after DiffEqCallbacks.jl#303 + DiffEqNoiseProcess.jl#271 + the problem-library bumps, resolution still trips over JumpProcesses, which is pulled in transitively via `ModelingToolkitBase` → `JumpProcesses` and via `StochasticDiffEqImplicit` → `JumpProcesses`:

```
Unsatisfiable requirements detected for package JumpProcesses [ccbc3e58]:
 JumpProcesses [ccbc3e58] log:
 ├─possible versions are: 9.0.0 - 9.25.1 or uninstalled
 ├─restricted to versions 9.13.1 - 9 by StochasticDiffEqImplicit [5080b986]
 ├─restricted to versions 9.23.0 - 9 by ModelingToolkitBase [7771a370]
 └─restricted by compatibility requirements with DiffEqBase [2b5f629d] to versions: uninstalled
   └─DiffEqBase [2b5f629d] is fixed to version 7.0.0
```

This widening unblocks that edge.

## Why this is source-level safe

The SciMLBase v3 / DiffEqBase v7 rename work was already landed in #579 ("Prepare source for SciMLBase v3 and widen compat"):

- `src/SSA_stepper.jl` defines both `DiffEqBase.u_modified!(::SSAIntegrator, ::Bool)` and, under `@static if isdefined(SciMLBase, :derivative_discontinuity!)`, `SciMLBase.derivative_discontinuity!(::SSAIntegrator, ::Bool)`. So SciMLBase v3's callback machinery (which now calls the renamed `derivative_discontinuity!`) dispatches correctly to the SSA integrator path.
- The generic `u_modified!(integrator, true/false)` call sites in `src/variable_rate.jl` / `src/aggregators/ssajump.jl` / `src/aggregators/coevolve.jl` still resolve on v3: SciMLBase keeps `u_modified!` as `@deprecate u_modified!(i::DEIntegrator, bool) derivative_discontinuity!(i, bool)` (scheduled removal 2028, per the SciMLBase comment). Runtime behavior is unchanged; expect only deprecation warnings, which can be cleaned up as a follow-up by mechanically swapping the remaining generic call sites to `derivative_discontinuity!`.
- No call sites for any of the other removed v3/v7 symbols (`has_destats`, `.destats`, `concrete_solve`, `fastpow`, `RECOMPILE_BY_DEFAULT`, `DEStats`, `QuadratureProblem`, `tuples()`/`intervals()`, standalone `DEAlgorithm`/`DEProblem`/`DESolution`).

## Test plan

- [x] Source grep for removed symbols — clean except for the already-handled `u_modified!` case (covered by the existing v3 guard + the scheduled-2028 SciMLBase deprecate forwarder).
- [ ] CI on this PR — source unchanged, so the currently-registered DiffEqBase 6.x / OrdinaryDiffEq 6.x test path should stay green.
- [ ] OrdinaryDiffEq.jl#3488: `Unsatisfiable requirements detected for package JumpProcesses` should disappear from MTK / Stoch-Impl test groups once this is registered.

## Intentionally out of scope

- `[compat] StochasticDiffEq = "6.82"` left alone. It's a test dep of JumpProcesses; doesn't propagate. The monorepo StochasticDiffEq v6.102.0 still satisfies the pin.
- Mechanical `u_modified!` → `derivative_discontinuity!` sweep in `variable_rate.jl` / `aggregators/*.jl` — belongs in a follow-up once the compat widening is in users' hands.

Co-Authored-By: Chris Rackauckas <accounts@chrisrackauckas.com>


Project.toml
```diff
@@ -45,8 +45,8 @@ Graphs = "1.11"
 KernelAbstractions = "0.9"
 LinearAlgebra = "1"
 LinearSolve = "3"
-OrdinaryDiffEq = "6"
-OrdinaryDiffEqCore = "3"
+OrdinaryDiffEq = "6, 7"
```
ChrisRackauckas:
```suggestion
OrdinaryDiffEq = "6"
```

PR #581, opened 2026-04-26
Bump compat for OrdinaryDiffEq v7 / StochasticDiffEq v7 ecosystem

ChrisRackauckas-Claude:
## Summary

This PR widens compat entries so JumpProcesses works with the upcoming OrdinaryDiffEq v7 / StochasticDiffEq v7 ecosystem, and migrates test code to the SciMLBase v3 / RecursiveArrayTools v4 APIs that shipped in prior PRs.

### Compat changes
- `OrdinaryDiffEq`: `"6"` → `"6, 7"` (main, docs, docs/src/assets Project.toml)
- `StochasticDiffEq`: `"6.82"` → `"6.82, 7"` (same files)
- `SciMLBase` was already `"2.115, 3.1"` and `DiffEqBase` was already `"6.192, 7"` from PRs #579 and #580; no changes needed.

### Source / test migrations
- **`prob_func` signature** (SciMLBase v3 breaking change): The old 3-arg `prob_func(prob, i, repeat)` is gone in SciMLBase v3.6.0, which calls `prob_func(prob, ctx)` where `ctx` is an `EnsembleContext`. Updated in `test/ensemble_uniqueness.jl` and `test/thread_safety.jl`.
- **`length(sol)`** (RecursiveArrayTools v4 breaking change): `length(AbstractVectorOfArray)` now returns total scalar element count (`size` product), not the number of trajectories/timesteps. `length(sol.u)` is unambiguous. Fixed in `test/ensemble_problems.jl`, `test/saveat_regression.jl`, `test/thread_safety.jl`, and `test/variable_rate.jl`.

### No source changes needed
- `src/SSA_stepper.jl` already handles `u_modified!` → `derivative_discontinuity!` rename via an `@static if isdefined(SciMLBase, :derivative_discontinuity!)` guard (added in PR #579).
- `u_modified!` calls in `src/variable_rate.jl` and aggregators are still valid: SciMLBase v3 keeps `u_modified!` as a deprecated re-export through DiffEqBase v7, so these generate deprecation warnings but continue to work.

## Related
- OrdinaryDiffEq v7 tracking: SciML/OrdinaryDiffEq.jl#3562, SciML/OrdinaryDiffEq.jl#3565


ChrisRackauckas-Claude:
Pushed a follow-up commit (4dbce91) that finishes the SciMLBase v3 rename properly and bumps this to a breaking v10.0.0.

## What the new commit does

- **Field rename**: `SSAIntegrator.u_modified::Bool` → `SSAIntegrator.derivative_discontinuity::Bool`. This is the actual fix for the FunctionWrapper test failure — DiffEqBase v7's `initialize!` (callbacks.jl:25) accesses `integrator.derivative_discontinuity` as a field, so a method/getproperty shim can't help. The field has to be renamed.
- **Drop the pre-v7 ecosystem entirely** (the question from the dig was whether to keep both — answer was no, renaming forces this anyway since DiffEqBase v6 callbacks read `integrator.u_modified` directly and would break the moment we rename). Compats:
  - `DiffEqBase`: `"6.192, 7"` → `"7"`
  - `OrdinaryDiffEq`: `"6, 7"` → `"7"`
  - `OrdinaryDiffEqCore`: `"3, 4"` → `"4"`
  - `OrdinaryDiffEqFunctionMap`: `"1, 2"` → `"2"`
  - `RecursiveArrayTools`: `"3.35, 4"` → `"4"`
  - `SciMLBase`: `"2.115, 3.1"` → `"3.1"`
  - `StochasticDiffEq`: `"6.82, 7"` → `"7"`
- **Version bump 9.26.0 → 10.0.0** (breaking ABI on the integrator struct).
- Internal `DiffEqBase.u_modified!` method removed; `SciMLBase.derivative_discontinuity!(::SSAIntegrator, ::Bool)` replaces it (no more `@static if isdefined` guard since SciMLBase v3 is now mandatory).
- Internal `u_modified!(integrator, ...)` calls in `variable_rate.jl`, `aggregators/coevolve.jl`, `aggregators/ssajump.jl` rewritten to `derivative_discontinuity!(integrator, ...)`. `using ... u_modified!` → `using SciMLBase: derivative_discontinuity!`.
- `docs/Project.toml` and `docs/src/assets/Project.toml` updated to drop v6 lower bounds.

## Companion work

- **#583** — patch release v9.26.1 on the v9 line that just tightens `OrdinaryDiffEqCore` to `"3.17, 4"` (fixes the precompile bug that blocks SBMLToolkit#199 / SBMLToolkitTestSuite#63 right now without forcing those repos onto v10).
- **JuliaRegistries/General#154093** — retroactive registry compat tightening for already-released v9.24.0 / v9.25.0 / v9.25.1 / v9.26.0, so installs in the wild stop pulling the broken combination.

This PR (now → v10) completes the v7-ecosystem migration; the other two clean up the v9 line.

isaacsas:
V10 also requires merging https://github.com/SciML/JumpProcesses.jl/pull/570, so if these changes are merged they should not be released at this time as a new release.

ChrisRackauckas:
This shouldn't require a v10 of JumpProcesses: nothing about the API here is changed? We can just keep a deprecation path for now.

ChrisRackauckas-Claude:
Reworked the field-rename approach into a non-breaking alias (force-pushed; previous v10 commit dropped). New plan: this PR stays a minor bump → **v9.27.0**, no API breakage.

## What the new commit does

- **Override `Base.getproperty` / `setproperty!` / `propertynames` on `SSAIntegrator`** so `:derivative_discontinuity` aliases the existing `:u_modified` field. Both names hit the same underlying storage. v6-era callbacks (which read `integrator.u_modified`) and v7-era callbacks (which read `integrator.derivative_discontinuity`) both keep working with no struct rename.
- **Field name unchanged** — `u_modified::Bool` stays. No ABI break for any external code that touches the integrator directly.
- **Kept both methods**: `DiffEqBase.u_modified!(::SSAIntegrator, ::Bool)` and the `@static if isdefined(SciMLBase, :derivative_discontinuity!)`-guarded `SciMLBase.derivative_discontinuity!(::SSAIntegrator, ::Bool)`. The SciMLBase v3 `@deprecate u_modified! → derivative_discontinuity!` alias is bypassed by JumpProcesses' more specific dispatch, so internal `u_modified!(integrator, ...)` calls don't trigger deprecation warnings.
- **Tightened the existing widening** from `OrdinaryDiffEqCore = "3, 4"` to `"3.17, 4"` so the resolver can't pick the broken v3.0–v3.16 versions (same lower-bound fix as #583).
- **Version bump 9.26.0 → 9.27.0** (minor, fully backwards compatible).

## Why this works for both ecosystems

DiffEqBase v6 callbacks: `integrator.u_modified` → hits the real field directly via default `getproperty` (sym ≠ `:derivative_discontinuity`).

DiffEqBase v7 callbacks: `integrator.derivative_discontinuity` → hits the alias branch in our `getproperty`, returns `getfield(integrator, :u_modified)`.

Same for writes via `setproperty!`. `propertynames` reports both for tab-completion / `show`.

The rest of the PR (compat widening, ensemble `prob_func` migration, `length(sol)` fix, OrdinaryDiffEqFunctionMap test dep) is unchanged.

ChrisRackauckas:
Okay this form is one that I am fine with. There is no breaking update in the source code. But I will note that the tests are using the SciMLBase v3 ensemble code now, so actually on earlier SciMLBase the tests will be broken. I think it will be very good for the ecosystem at least for now to have some backwards compatible versions, so because this is only in the test source code I think it would be good to allow this for now, and then bump the lower bound later. This should unblock a whole lot and the tests confirm the actual source code is good so I think it's best to make that judgement call there.

ChrisRackauckas:
@isaacsas on the v10 I assume you'll just want to drop the backwards compat path and just use the new derivative_discontinuity naming?

isaacsas:
Yes, V10 will be SciMLBase 3 only as it makes changes to rng handling that would be unsafe prior to v3.

isaacsas:
You can see the PR for what is changing here -- I just need to update it to master and check everything is passing. I think it should otherwise be good to go, but I won't have time to play with it and give it a last review for a few weeks.

PR #596, opened 2026-06-03
Add optional StochasticAD extension for differentiating jump processes

RomanSahakyan03:
Adds JumpProcessesStochasticADExt, an optional package extension (loaded only when StochasticAD and Distributions are present) that lets StochasticAD's derivative_estimate compute gradients of expectations over variable-rate jump processes.

  Differentiating through the solver directly --
  derivative_estimate(g -> solve(jprob(g), Tsit5())) -- does not compose: the adaptive OrdinaryDiffEq internals require Number-interface methods that StochasticTriple intentionally omits, and VR_Direct's ContinuousCallback rootfind is a boolean predicate on a triple. The extension instead provides a triple-generic, fixed-grid PDMP simulator (generic RK4 + per-channel Bernoulli with first-fire priority + multiplicative-select update) that composes by design.

  - src/JumpProcesses.jl: fixedgrid_simulate / fixedgrid_jump_observable stubs and exports (no StochasticAD code in src/)
  - ext/JumpProcessesStochasticADExt.jl: the simulator and a JumpProblem adapter
  - Project.toml: StochasticAD and Distributions as weakdeps/extension + compat
  - test/stochasticad_tests.jl: analytic-gradient and MCWF tests (registered)

### DIFF
--- a/ext/JumpProcessesStochasticADExt.jl
+++ b/ext/JumpProcessesStochasticADExt.jl
@@ -1,0 +1,111 @@
+module JumpProcessesStochasticADExt
+
+# Optional StochasticAD support for JumpProcesses.
+#
+# Why a separate fixed-grid simulator instead of differentiating `solve`?
+# `derivative_estimate(γ -> solve(jprob(γ), Tsit5()))` does not compose: (1) the
+# adaptive OrdinaryDiffEq solver internals assume the full `Number` interface
+# that `ForwardDiff.Dual` implements but `StochasticAD.StochasticTriple` omits
+# (so triples can't even enter the ODE state), and (2) `VR_Direct` locates jump
+# times with a `ContinuousCallback` rootfind, i.e. a boolean predicate on a
+# triple, which StochasticAD forbids by design. Both were established
+# empirically. This extension instead provides a triple-generic, fixed-grid
+# PDMP simulator that composes *by design* (pure generic arithmetic +
+# `rand(Bernoulli)`), which is what "AD on the algorithm" amounts to here.
+
+using JumpProcesses
+using StochasticAD
+using Distributions: Bernoulli
+
+# generic RK4 step over a triple-safe out-of-place drift
+@inline function _rk4(drift, u, p, t, dt)
+    k1 = drift(u, p, t)
+    k2 = drift([u[i] + 0.5dt * k1[i] for i in eachindex(u)], p, t + 0.5dt)
+    k3 = drift([u[i] + 0.5dt * k2[i] for i in eachindex(u)], p, t + 0.5dt)
+    k4 = drift([u[i] + dt * k3[i] for i in eachindex(u)], p, t + dt)
+    return [u[i] + (dt / 6) * (k1[i] + 2k2[i] + 2k3[i] + k4[i]) for i in eachindex(u)]
+end
+
+"""
+    fixedgrid_simulate(drift, channels, u0, p, tspan, dt) -> u(tspan[2])
+
+Triple-safe, fixed-grid piecewise-deterministic Markov process simulator that
+composes with StochasticAD's `derivative_estimate`/`stochastic_triple`.
+
+On each step of size `dt`: integrate the smooth flow with a generic RK4, then for
+each jump channel sample `rand(Bernoulli(rateₖ(u,p,t) * dt))` and apply the
+post-jump state of the first channel that fires (first-fire priority), all via a
+multiplicative-select blend. Using one `Bernoulli` per channel (rather than one
+`Bernoulli` for "any jump" plus a `Categorical` for "which") avoids division, so
+there is no `Categorical` 0/0 when the total rate vanishes.
+
+Arguments:
+  - `drift(u, p, t) -> du`: triple-safe, out-of-place ODE RHS.
+  - `channels`: iterable of named tuples `(rate, post)` where
+    `rate(u, p, t) -> ≥0 scalar` and `post(u, p, t) -> post-jump state vector`.
+  - `u0`, `p`, `tspan`, `dt`: initial state, parameters, `(t0, tf)`, step.
+
+Validity: this is an O(dt²)-per-step (τ-leap-style) approximation to the exact
+continuous-time PDMP, accurate when `rateₖ · dt ≪ 1`.
+"""
+function JumpProcesses.fixedgrid_simulate(drift, channels, u0, p, tspan, dt)
+    nsteps = round(Int, (tspan[2] - tspan[1]) / dt)
+    ptype  = sum(p)                                   # carries the (maybe-triple) type
+    u      = [u0i + 0 * ptype for u0i in u0]          # promote state to p's type
+    n      = length(u)
+    t0     = tspan[1]
+
+    for s in 1:nsteps
+        t        = t0 + (s - 1) * dt
+        u_smooth = _rk4(drift, u, p, t, dt)
+
+        remaining = 1 + 0 * ptype                     # triple, value 1
+        collapse  = [0 * ptype for _ in 1:n]          # triple zeros
+        for ch in channels
+            r     = ch.rate(u, p, t)
+            fired = rand(Bernoulli(r * dt))           # triple 0/1
+            take  = fired * remaining                 # 1 iff first firing channel
+            pj    = ch.post(u, p, t)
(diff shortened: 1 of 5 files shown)


isaacsas:
I think I'll let @ChrisRackauckas take a first pass at this as I don't know the ODE AD internals that well. I do have one concern though, which is that I'm not sure StochasticAD is really being maintained or developed anymore. Maybe Chris can comment on that.

isaacsas:
In either case though thanks for the contribution!

isaacsas:
More generally, if you are interested in getting sensitivity calculations into JumpProcesses and want to talk about that sometime please let me know -- I've been wanting to build that out and had it on my TODO.

RomanSahakyan03:
> More generally, if you are interested in getting sensitivity calculations into JumpProcesses and want to talk about that sometime please let me know -- I've been wanting to build that out and had it on my TODO.

Thanks @isaacsas, that makes sense.

I'd definitely be interested in talking about sensitivity calculations for JumpProcesses. I'll wait for @ChrisRackauckas's first pass here as well, since this started from his suggestion and I want to make sure I'm aligned with the intended direction.

Happy to treat this PR as a prototype and discuss what a better long-term path should look like.

ext/JumpProcessesStochasticADExt.jl
```diff
@@ -0,0 +1,111 @@
+module JumpProcessesStochasticADExt
+
+# Optional StochasticAD support for JumpProcesses.
+#
+# Why a separate fixed-grid simulator instead of differentiating `solve`?
+# `derivative_estimate(γ -> solve(jprob(γ), Tsit5()))` does not compose: (1) the
+# adaptive OrdinaryDiffEq solver internals assume the full `Number` interface
+# that `ForwardDiff.Dual` implements but `StochasticAD.StochasticTriple` omits
+# (so triples can't even enter the ODE state), and (2) `VR_Direct` locates jump
```
ChrisRackauckas, line 9: > implements but `StochasticAD.StochasticTriple` omits (so triples can't even enter the ODE state)

That doesn't seem accurate? It's a direct extension so the theory works out. What happens if you try it? Make an MWE. IIRC that already works with the package as-is.

ext/JumpProcessesStochasticADExt.jl
```diff
@@ -0,0 +1,111 @@
...
+# times with a `ContinuousCallback` rootfind, i.e. a boolean predicate on a
+# triple, which StochasticAD forbids by design. Both were established
```
ChrisRackauckas, line 11: That is required for a general variable-rate jump. So is this scoped only for fixed rate? If it's scoped for fixed rate, it can probably be simplified to just use `tstops`, which then would be a lot faster.

ChrisRackauckas:
It would be much easier to start with the jump-only case. Start with the direct method on purely constant rate jumps and SSA stepper and get that differentiating.

ext/JumpProcessesStochasticADExt.jl
```diff
@@ -0,0 +1,111 @@
...
+there is no `Categorical` 0/0 when the total rate vanishes.
+
+Arguments:
+  - `drift(u, p, t) -> du`: triple-safe, out-of-place ODE RHS.
+  - `channels`: iterable of named tuples `(rate, post)` where
+    `rate(u, p, t) -> ≥0 scalar` and `post(u, p, t) -> post-jump state vector`.
+  - `u0`, `p`, `tspan`, `dt`: initial state, parameters, `(t0, tf)`, step.
+
+Validity: this is an O(dt²)-per-step (τ-leap-style) approximation to the exact
+continuous-time PDMP, accurate when `rateₖ · dt ≪ 1`.
```
ChrisRackauckas, line 49: I'm not sure this accuracy actually holds? This is differentiating as a tau-leap method, of which none are order 2. So this is likely only O(dt)

ext/JumpProcessesStochasticADExt.jl
RomanSahakyan03: Thanks, I reduced this further. You were right that my original wording was too broad, especially because `one(γ)` in StochasticAD returns `Float64`, so my first example accidentally created a `Vector{Float64}` state.

But even with a scalar triple state, `Tsit5()` currently fails before any JumpProcesses code is involved:

```julia
using OrdinaryDiffEq
using StochasticAD

function loss_scalar_state_triple(γ)
    prob = ODEProblem(
        (u, p, t) -> -u,
        γ,
        (0.0, 1.0),
        nothing,
    )

    sol = solve(prob, Tsit5())
    return sol[end]
end

derivative_estimate(loss_scalar_state_triple, 2.0)
```

This errors during `OrdinaryDiffEqCore.__init` with:

```julia
ERROR: MethodError: no method matching StochasticAD.StochasticTriple{StochasticAD.Tag{…}, Float64, StochasticAD.PrunedFIsBackend.PrunedFIs{…}}(::Float64)
The type `StochasticAD.StochasticTriple{StochasticAD.Tag{typeof(loss_param_triple), Float64}, Float64, StochasticAD.PrunedFIsBackend.PrunedFIs{Float64}}` exists, but no method is defined for this combination of argument types when trying to construct it.

Closest candidates are:
  StochasticAD.StochasticTriple{T, V, FIs}(::V, ::V, ::FIs) where {T, V, FIs<:StochasticAD.AbstractFIs{V}}
   @ StochasticAD C:\Users\Roman\.julia\packages\StochasticAD\Gef5s\src\stochastic_triple.jl:21
  (::Type{T})(::T) where T<:Number
   @ Core boot.jl:965
  (::Type{T})(::Static.StaticInteger) where T<:Real
   @ Static C:\Users\Roman\.julia\packages\Static\d7YOk\src\Static.jl:442
  ...
```

and the stacktrace goes through:

```julia
oneunit(x::StochasticAD.StochasticTriple{...})
```

So I’ll revise the comment to avoid saying that triples cannot enter the ODE state or that OrdinaryDiffEq assumes the full `Number` interface in general. The more precise issue is that this `Tsit5` initialization path currently calls `oneunit`/scalar construction in a way that `StochasticTriple` does not support. Separately, the JumpProcesses variable-rate callback/rootfinding path has its own issue with predicates on triples.

ext/JumpProcessesStochasticADExt.jl
ChrisRackauckas: And if it's only O(dt), then just euler the other terms? In which case, this is just a tau-leap effectively done in first reaction version.

ext/JumpProcessesStochasticADExt.jl
RomanSahakyan03: You're right. Exact general variable-rate jumps need the integrated-hazard rootfind, so I should not describe this as exact `VariableRateJump` support. I’ll fix that wording.

The scope is not fixed-rate only. The target is state-dependent rates, e.g. my MCWF case where the rate depends on `|c₂(t)|²`. Since StochasticAD cannot differentiate through the rootfind predicate, this uses a fixed-grid / tau-leap-style approximation instead.

For constant-rate jumps, I agree that `tstops` is the better path. I confirmed that pre-sampling `N ~ Poisson(λT)` and placing jumps with `tstops` differentiates correctly, so I’m going to start working for the constant rate one.

RomanSahakyan03:
> It would be much easier to start with the jump-only case. Start with the direct method on purely constant rate jumps and SSA stepper and get that differentiating.

Following your suggestion, I tried the constant-rate jump-only case with `Direct` + `SSAStepper` on a pure-birth example.

It does not differentiate as-is because `DirectJumpAggregation` stores propensities in a `Vector{Float64}` around `fill_cur_rates` in `direct.jl:109`, so a `StochasticTriple` rate hits `Float64(::StochasticTriple)`.

I see two possible first steps:

1. Make the direct aggregator generic over the rate type, so triples can pass through the existing SSA path.

2. For purely constant-rate jumps, skip the SSA internals and pre-sample `N ~ Poisson(λT)`, then place jumps with `tstops`. This differentiates correctly in the pure-birth case and recovers `d/dλ E[N] = T`.

ChrisRackauckas:
You cannot presample because the rate is determined by the previous set of values. You have to iterate that. The stochastic triples are made for handling such iteration. 

> Make the direct aggregator generic over the rate type, so triples can pass through the existing SSA path.

Yes

PR #599, opened 2026-06-06
Re-enable downgrade (strict, allow_reresolve=false); intentionally RED pending upstream ForwardDiff/LogExpFunctions — red until ForwardDiff 1.3.4 (#811). saveat fix is separate (#600) — do NOT touch it — natural failure, not disabled; auto-greens on upstream release.

ChrisRackauckas-Claude:
## Summary

The `IRKGL2026a` rewrite (62e2fe2) deleted the in-place `PolInterp!` from `src/IRKCoefficients.jl` and rewrote the internal `tcoeffs`/`tcache` struct layout plus the `GaussLegendreCoefficients!`/`EstimateCoeffs!` signatures, but left `test/alloc_tests.jl` calling the old API. As a result the test suite was red at **any** compat: `UndefVarError: PolInterp! not defined` at the import line, and (once that is fixed) constructor/arity mismatches.

## Changes

- **Restore `PolInterp!`** (in-place Lagrange interpolation writing into a preallocated buffer) in `src/IRKCoefficients.jl`, ported to the current code, so the zero-allocation test's intent holds.
- **Port `alloc_tests.jl`** to the rewritten internal API:
  - add the `eta` array (`zeros(s, s)`) to the `tcoeffs` constructor (now 13 fields),
  - call `GaussLegendreCoefficients!(s, mu, c, b, eta, T)` and `EstimateCoeffs!(s, alpha, T)` (new arities),
  - add the trailing `verbose::DEVerbosity` field to the `tcache` constructor.
- **Fix the `MyNorm` allocation measurement** via a function barrier. `@allocated` on a call closing over a non-const `@testset` variable measures the boxing of that captured variable, not the callee — `MyNorm` itself allocates 0 bytes (verified by measuring inside a function). This keeps the zero-allocation assertion meaningful; it is not weakened.
- **Re-enable `Downgrade.yml` to RUN STRICT.** Removed the `if: false` gating so the downgrade job is no longer skipped. It runs **strict** (`allow_reresolve=false`) — floors are NOT lowered and `allow-reresolve` is NOT added. The strict resolve currently cannot be satisfied (the floored DiffEqBase/NonlinearSolveBase stack pins LogExpFunctions 1.0.1, which caps ForwardDiff while the test extras require ForwardDiff 1.x — disjoint windows), so the job is **intentionally RED**. This is a natural, visible failure, not a disabled job. It will **auto-green** when the upstream fix releases (ForwardDiff/LogExpFunctions — red until ForwardDiff 1.3.4, #811).
- **Bump `Reexport` floor `1.0` -> `1.2.2`.** Transitive `BracketingNonlinearSolve` requires `Reexport >= 1.2.2`; floors `1.0`/`1.1`/`1.2.0`/`1.2.1` fail to resolve.

## Testing (Julia 1.10, local)

Full suite at latest compat: **PASS**
```
Test Summary:    | Pass  Total  Time
Allocation Tests |    6      6
     Testing IRKGaussLegendre tests passed
```

Runic formatting checked clean on the changed `.jl` files.

> Ignore until reviewed by @ChrisRackauckas.


ChrisRackauckas-Claude:
Note re: the SciML/.github downgrade-workflow change (always `allow_reresolve: false`, removing the `allow-reresolve` `workflow_call` input): this PR's `Downgrade.yml` does not pass an `allow-reresolve:` input (it only references the term in the explanatory comment), so no edit is needed here — it stays compatible after `@v1` is retagged.

PR #606, opened 2026-06-13
Stochasticad constant rate

RomanSahakyan03:
## Summary

Adds an optional `StochasticAD` extension that makes `derivative_estimate` work on expectations over **jump-only `ConstantRateJump` (SSA) problems**, following @ChrisRackauckas's suggestion to start with the constant-rate case.

**New API (in `ext/JumpProcessesStochasticADExt.jl`, loaded only when StochasticAD + Distributions are present):**
- `constant_rate_ssa_final_state(jprob, p; nmax, tspan, return_saturation=false)` — the main, *exact* iterative SSA method, differentiable and supporting **state-dependent** rates.
- `poisson_count_final_state(jprob, p; tspan)` — a closed-form helper for the narrow *state-independent* (homogeneous-Poisson) case, with guards against misuse.

**Also:** `src/aggregators/ssajump.jl` makes the SSA rate cache generic over the rate type; `src/JumpProcesses.jl` adds stubs + exports (no StochasticAD in `src/`); tests run in an isolated `GROUP=StochasticAD` environment.

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated 
- [x] The new code follows the
- [x] Any new documentation only uses public API

## Additional context

**Why a reformulation (the key point).** A naive Gillespie loop differentiated with StochasticAD returns **`0`** whenever the parameter acts through the event *times* — e.g. pure death, `rate = μ·u`, where a larger `μ` simply packs in more deaths before `T`. The `while t < T` termination branches on the *primal* time, so the event count (the parameter-dependent quantity) is frozen and the derivative is silently dropped.

`constant_rate_ssa_final_state` fixes this **exactly** by rewriting each "is there a next event before `T`?" as a *tracked* `Bernoulli(1 - exp(-total_rate·(T-t)))`, run over a **fixed-length** loop, with the event time drawn from the truncated exponential, the reaction chosen by stick-breaking `Bernoulli`s, and the state updated by multiplicative-select masking (no branching on a triple, no triple-valued indexing). This has the **same trajectory distribution as the SSA** (it is not a τ-leap approximation), recovers analytic gradients, and handles state-dependent rates and multiple channels.

**Validation** (vs. closed-form analytic expectations):

| Test | StochasticAD | Analytic |
|---|---|---|
| pure death, `d/dμ E[u(T)]` | `-60.58 ± 0.43` | `-60.65` |
| birth–death, `d/dλ` | `≈ 0.81` | `0.86` |
| birth–death, `d/dμ` | `≈ -40` | `-41.1` |

**On the generic rate cache.** Per the request to let `StochasticTriple` rates flow through the existing `Direct`/`SSAStepper` path, `build_jump_aggregation` now sizes `cur_rates` by the *inferred* rate type instead of `typeof(t)` (via `Base.promote_op`, so it never executes the rate at build time; `Float64` behavior is unchanged and the existing `Constant Rate`/SSA tests pass). **Honest caveat:** this change *alone* does not produce correct gradients — the stock `while t < T` boundary still drops the event-count derivative. Correct gradients come from the reformulation above.

**Scope / limitations (this first step).** `ConstantRateJump` only (not `VariableRateJump`); jump-only problems (no continuous drift); additive affects (the net change is inferred from `affect!` and checked); a fixed `nmax` bound on the number of events (with an optional saturation flag). The earlier fixed-grid `VariableRateJump`/PDMP approach (#596) is intentionally set aside in favor of this narrower, *exact* constant-rate first step.

**Dependency note.** `StochasticAD` is an optional `[weakdeps]` extension, and its tests run in a dedicated `GROUP=StochasticAD` environment (no OrdinaryDiffEq) so they don't perturb the main test suite. (The currently-resolvable `StochasticAD` is fairly old — happy to discuss whether that's a concern for taking this as a dependency.)


RomanSahakyan03:
Hi @ChrisRackauckas and @isaacsas, please let me know when you get a chance to review the changes. Thanks!

isaacsas:
@RomanSahakyan03 thanks! 

I'd prefer if you didn't change the `Direct` code path or helper functions, and just added dispatches or separate helpers for the StochasticAD version. The `Direct` version has been extensively benchmarked, and I wouldn't want to risk its performance, so changes there would need a lot of benchmarking to be accepted.

Are you able to setup your solver to give full paths and allow evaluation of those paths at set times, or to at least support (maybe require?) `saveat` and return the solution at fixed times? This would be a lot more useful for users than just providing the solution at a fixed terminal time, and with a bias due to capping the number of steps. 

If capping the number of steps is really needed, could you instead assume a maximum total propensity bound? This would allow you to pre-sample the number of events and event firing times, and to just have a null event to account for the difference between the bound and the current total propensity (keeping everything StochasticAD compatible). While it is still not going to be possible to provide such a bound for most chemical systems, for systems where the amount of each chemical is rigorously bounded such a bound would generally hold, so it seems like a more generally applicable solver that avoids the bias in having a `nmax` parameter.

RomanSahakyan03:
Thanks @isaacsas, that makes sense.

I’ll remove the changes to the `Direct` path from this PR and keep the StochasticAD implementation separate. 

For the output, I'll make it return all path and make solver to work with `saveat`.

The bounded-propensity idea also sounds better than `nmax`. I’ll look into replacing the current capped-step approach with a thinning/uniformization version using a user-provided total propensity bound, with null events for the remaining rate. That should avoid the `nmax` bias when a valid bound is available.

So I'll update the PR in that direction: separate StochasticAD code path, `saveat` support, and bounded-propensity thinning instead of the current `nmax`-based version.

ext/JumpProcessesStochasticADExt.jl
```diff
@@ -0,0 +1,204 @@
...
+# `Bernoulli(total_rate(u)/Λ)` (else a null event), and the channel is chosen by
+# stick-breaking. All parameter dependence flows through the accept/channel
+# Bernoullis, so the gradient is captured; it is unbiased given a valid `Λ`, and
+# `saveat` is exact because the candidate times are fixed Float64.
+#
+# This code is fully SEPARATE from the `Direct` aggregator: it reads `jump.rate` /
+# `jump.affect!` only and never touches `time_to_next_jump` or the Direct rate cache.
+#
+# Scope: jump-only `DiscreteProblem`s with `ConstantRateJump`s (state-dependent
+# rates OK) and additive affects. No `MassActionJump`, no `VariableRateJump`.
```
ChrisRackauckas: This solver isn't dependent on StochasticAD though, so it shouldn't be in the extension. It can just live in the source.

ext/JumpProcessesStochasticADExt.jl
```diff
@@ -0,0 +1,204 @@
...
+    tsave = typeof(t0)[]
+    usave = typeof(u)[]
+    if ss
+        push!(tsave, t0)
+        push!(usave, copy(u))
+    end
+
+    # candidate events ~ homogeneous Poisson(Λ) on [t0, tf]. PARAMETER-FREE: Λ is a
+    # constant, so M and the times carry no derivative and never branch on a triple.
+    M = rand(Poisson(Λ * ΔT))
```
ChrisRackauckas: We should check the algorithm used here.

ext/JumpProcessesStochasticADExt.jl
```diff
@@ -0,0 +1,204 @@
...
+    t0, tf = first(tspan), last(tspan)
+    ΔT    = tf - t0
+    K     = length(jumps)
+    n     = length(u0)
+
+    saveat_times, ss, se = JumpProcesses._process_saveat(saveat, (t0, tf),
+        save_start, save_end)
+
+    Δ = [_additive_change(jumps[k], u0, p, t0) for k in 1:K]   # Float64 net change/channel
+    z = 0 * sum(p)                                # triple zero carrying p's type
```
ChrisRackauckas: This is assuming `p` is a vector

ChrisRackauckas: I guess that's a fine limitation for now, but it should be documented. The type handling in DiffEqBase would be better, but it's also a lot more to handle. We should probably at least specialize for SciMLStructures objects so that MTK/Catalyst get support for its tunables.

ChrisRackauckas:
I still don't see why you'd need a static bound on the loop for StochasticAD, but this is at least an interesting SSA to add.

isaacsas:
I suggested this because it should be simpler than getting methods  with branching termination working with StochasticAD (which I'm not sure how to do currently either).

ChrisRackauckas:
But I don't think it needs to be in the extension. This is a perfectly fine fast SSA method, with a caveat of requiring a good rate bound. It should be just implemented and tested as such a method, but then also tested and documented as a good choice to use with StochasticAD as well, but that's a secondary thing.

isaacsas:
Yes, I don't disagree. If it is written generically to work with and without StochasicAD then no extension should be needed.

src/bounded_ssa.jl
```diff
@@ -0,0 +1,183 @@
...
+    (cj === nothing || isempty(cj)) &&
+        error("BoundedSSA requires at least one ConstantRateJump.")
+    nothing
+end
+
+# Draw a {0,1} Bernoulli(p). The default (primal) path uses `rand() < p`; the
+# StochasticAD extension overloads this for `StochasticTriple` so the discrete decision
+# becomes differentiable. Isolating this one call is what keeps the solver itself free
+# of any StochasticAD dependency.
+_bounded_ssa_bernoulli(p) = (rand() < p) ? one(p) : zero(p)
```
ChrisRackauckas: This could always just be `rand(Bernoulli(p))`? That doesn't need a new dep.

src/bounded_ssa.jl
```diff
@@ -0,0 +1,183 @@
...
+# `Λ = rate_bound`): candidate event times are a homogeneous Poisson process of rate
+# `Λ` — parameter-free, so the loop never branches on a (triple-valued) time and the
+# times stay `Float64`. At each candidate the event is accepted with a
+# `Bernoulli(total_rate(u)/Λ)` (otherwise a null event absorbing the slack `Λ - a(u)`),
+# and the firing channel is chosen by stick-breaking. When a `StochasticTriple`
+# parameter flows in, the accept/channel Bernoullis become tracked, so the gradient is
+# captured; it is unbiased given a valid `Λ`, and `saveat` is exact.
+
+# minimal integrator-like object so a jump's `affect!` can be applied to a scratch
+# state to read off its net effect.
```
ChrisRackauckas:
```suggestion
```

ext/JumpProcessesStochasticADExt.jl
```diff
@@ -0,0 +1,17 @@
+module JumpProcessesStochasticADExt
+
+# BoundedSSA itself lives in JumpProcesses `src` — it is an ordinary uniformization
+# (thinning) SSA and does not depend on StochasticAD. This extension supplies the one
+# StochasticAD-specific piece: it makes BoundedSSA's discrete accept/channel decisions
+# differentiable by overloading the `_bounded_ssa_bernoulli` hook for `StochasticTriple`
+# parameters, via StochasticAD's differentiable `rand(::Bernoulli)`. When StochasticAD
+# and Distributions are loaded, `solve(jprob, BoundedSSA(; rate_bound))` /
+# `bounded_ssa_path` compose with `derivative_estimate`/`stochastic_triple`.
```
ChrisRackauckas: doesn't need necessary to have this Ext

src/JumpProcesses.jl
```diff
@@ -127,12 +127,16 @@ include("solve.jl")
...
-export SSAStepper
+export SSAStepper, BoundedSSA
 
 # leaping: 
 include("simple_regular_solve.jl")
 export SimpleTauLeaping, SimpleExplicitTauLeaping, EnsembleGPUKernel
 
+# BoundedSSA: uniformization SSA solver (differentiable via the StochasticAD extension)
+include("bounded_ssa.jl")
+export bounded_ssa_path
```
ChrisRackauckas:
```suggestion
```

only `BoundedSSA` should be public, and that should get docs.

RomanSahakyan03:
@ChrisRackauckas @isaacsas,  here’s the current state:

* The StochasticAD extension is gone. Uniformization SSA now just calls `rand(Bernoulli(p))` directly in `src`, so there’s no extra StochasticAD glue or `_bounded_ssa_bernoulli` hook anymore. If StochasticAD is loaded and `p` contains a `StochasticTriple`, it handles `rand(::Bernoulli)` itself; otherwise `BoundedSSA` stays a normal SSA with no StochasticAD dependency.

* `Distributions` is now a hard dep. That’s the main tradeoff here: `rand(Bernoulli(...))` now needs it directly. If the extra dependency weight is a concern, switching back to a weakdep plus a tiny extension hook is straightforward.

* MTK/Catalyst parameters now work. The old `0 * sum(p)` type-seeding path broke for `MTKParameters`, so it now seeds from the tunable portion via `SciMLStructures.canonicalize(Tunable(), p)`. Plain vectors still behave the same. I also verified that a Catalyst non-mass-action `ConstantRateJump` model differentiates under `BoundedSSA` and matches finite differences.

* Mass-action support is still future work. `BoundedSSA` is currently `ConstantRateJump`-only; supporting Catalyst mass-action networks will need `evalrxrate` to be triple-safe first.

RomanSahakyan03:
Update: MassActionJump now works — and it didn't need the evalrxrate changes I expected. BoundedSSA computes the propensity itself (`scaled_rate * ∏(u[spec]-j)`), so it never calls `evalrxrate`; hand-built (`param_idxs`) mass-action differentiates and matches the analytic. Tests now run through the public `solve(jprob, BoundedSSA(...))` interface.

isaacsas:
Great! I should hopefully get some time to start reviewing again in a week or so. I'm in the middle of back to back conferences this and next week, so swamped.

isaacsas:
But I'm glad the bounded method is working out ok!

src/bounded_ssa.jl
```diff
@@ -0,0 +1,300 @@
...
+    coerces rates to `Float64`, which drops the `StochasticTriple` (a documented follow-up).
+  - The differentiation parameter `prob.p` may be a plain numeric collection (e.g. a
+    `Vector`) or a SciMLStructures parameter object (MTK/Catalyst `MTKParameters`); in
+    the latter case the differentiable **tunable** portion is the target (extracted via
+    `SciMLStructures.canonicalize`), matching how the rest of JumpProcesses treats `p`.
+  - The solver itself is plain (no StochasticAD dependency): with ordinary parameters
+    `solve(jprob, BoundedSSA(; rate_bound))` is a uniformization SSA simulation. It
+    becomes differentiable when the user loads `StochasticAD` and passes a
+    `StochasticTriple` parameter — StochasticAD's own `rand(::Bernoulli)` rule makes
+    the accept/channel decisions differentiable, with no glue needed from this package.
```
ChrisRackauckas:
```suggestion
```

src/bounded_ssa.jl
```diff
@@ -0,0 +1,295 @@
...
+                 _bssa_ma_rate(u, maj, maunscaled, k - Kc) for k in 1:K]
+        total = sum(rates)
+        # thinning: real vs null event. `rand(Bernoulli(p))` handles both the primal
+        # draw and — when a StochasticTriple `p` flows in with StochasticAD loaded —
+        # the differentiable decision (StochasticAD's own `rand(::Bernoulli)` rule).
+        accept = rand(Bernoulli(total / Λ))
+
+        # which channel: stick-breaking conditional Bernoullis (last deterministic)
+        notchosen = 1 + z
+        sel = [z for _ in 1:n]
```
ChrisRackauckas, line 226: hoist this allocation out? Allocate once, grow if needed, but it seems unnecessary

src/bounded_ssa.jl
```diff
@@ -0,0 +1,295 @@
...
+        for k in 1:K
+            chose = k < K ?
+                    rand(Bernoulli(rates[k] / (sum(rates[j] for j in k:K) + 1e-300))) :
```
ChrisRackauckas, line 229: what's the magic number for?

src/bounded_ssa.jl
```diff
@@ -0,0 +1,295 @@
...
+                    (1 + z)
+            take = notchosen * chose
+            sel = [sel[i] + take * Δ[k][i] for i in 1:n]
+            notchosen = notchosen * (1 - chose)
+        end
+
+        u = [u[i] + accept * sel[i] for i in 1:n]        # apply only on a real event
```
ChrisRackauckas, line 236: Get rid of these allocations

ChrisRackauckas:
It looks correct now. But, test it with non-float64 rates, like float32, and see about that magic number handling. Float64 magic numbers are not a good sign.... and I don't see why it needs to be there? If you need to `nextfloat` then use that, not a fixed Float64 since that can promote incorrectly.

And there's a lot allocating that doesn't need to allocate. That could be a follow up PR though, I'd be happy to checkpoint based on correctness and continue to improve in a follow up.

isaacsas:
﻿Can we get some correctness tests for regular usage. i.e. can we add it into some of the existing correctness tests that sweep over all the aggregators? Many of them won’t be boundable, but perhaps some are?-SamOn Jul 8, 2026, at 2:35 PM, Christopher Rackauckas ***@***.***> wrote:﻿ChrisRackauckas left a comment (SciML/JumpProcesses.jl#606)
It looks correct now. But, test it with non-float64 rates, like float32, and see about that magic number handling. Float64 magic numbers are not a good sign.... and I don't see why it needs to be there? If you need to nextfloat then use that, not a fixed Float64 since that can promote incorrectly.
And there's a lot allocating that doesn't need to allocate. That could be a follow up PR though, I'd be happy to checkpoint based on correctness and continue to improve in a follow up.

RomanSahakyan03:
Thanks @ChrisRackauckas and @isaacsas for the feedback, and also for your patience with this PR. It took me a bit longer than expected to get back to it, but I’ve now updated BoundedSSA with numeric-type genericity, consistent `JumpProblem` RNG usage, and ordinary SSA correctness/regression tests. Both suites pass: ordinary SSA 17/17 and StochasticAD 21/21.

ChrisRackauckas:
Rebase

RomanSahakyan03:
> Rebase

Done

isaacsas:
I will try to review in the next few days. Thanks!

if we can get this finished off and put together some nice examples this could probably be a short paper for a math bio / chem journal.

RomanSahakyan03:
> I will try to review in the next few days. Thanks!
> 
> if we can get this finished off and put together some nice examples this could probably be a short paper for a math bio / chem journal.

Thanks, Sam! I really appreciate it. I’d be happy to put together some nice examples once we get the PR finalized, and I’d definitely be interested in working on a paper around this.

src/bounded_ssa.jl
RomanSahakyan03: Done!

src/bounded_ssa.jl
RomanSahakyan03: I changed the logic of counting the rates, so there isn't any magic numbers nor epsilons

src/bounded_ssa.jl
RomanSahakyan03: done

PR #607, opened 2026-06-14
Use SciMLTesting v1.2 (folder-based run_tests)

ChrisRackauckas-Claude:
Converts the JumpProcesses.jl test suite to the [SciMLTesting](https://github.com/SciML/SciMLTesting.jl) v1.2 model.

### Why explicit-args (not bare folder-discovery)
JumpProcesses has GPU-only groups (`CUDA`, `Correctness`) that are **not** declared in `test_groups.toml` and are run by a separate `GPU.yml` workflow against the `test/gpu/` sub-environment (via `activate_gpu_env`). SciMLTesting's bare folder-discovery (`run_tests()`) errors on a `GROUP` value that is not in `test_groups.toml` and cannot activate the GPU env per-group, so it cannot express these. There is also no `Core` group (`All = InterfaceI + InterfaceII`), which folder-discovery's mandatory non-empty Core would conflict with. This PR therefore uses the explicit-args `run_tests(...)` form over a folder layout (same pattern as the OrdinaryDiffEq.jl conversion).

### Folder layout
- `test/InterfaceI/` — the 21 files `GROUP=InterfaceI` ran.
- `test/InterfaceII/` — the 18 files `GROUP=InterfaceII` ran, including the former `test/spatial/*.jl` (flattened; each file is self-contained).
- `test/qa/qa.jl` — the QA body (runs in the root test env; Aqua + ExplicitImports are in the root `[extras]`).
- `test/gpu/` — unchanged (CUDA env).
- `test/shared/run_spatial_tests.jl` — the unused manual spatial driver, parked out of the group folders (never run by `runtests.jl`).

### runtests.jl
A single `run_tests(...)` call:
- `groups` = `InterfaceI` / `InterfaceII` / `CUDA` / `Correctness` thunks. The `CUDA` and `Correctness` thunks keep `activate_gpu_env()` verbatim.
- `qa = test/qa/qa.jl`.
- `all = ["InterfaceI", "InterfaceII"]` — curated, matching the old `All = InterfaceI + InterfaceII` (QA/CUDA/Correctness excluded).

Each Interface thunk preserves the original `@safetestset` labels and run order, updating only the include paths to the new folders.

### Behavior preservation
The set of tests run under each `GROUP` value (`All`, `InterfaceI`, `InterfaceII`, `QA`, `CUDA`, `Correctness`) is unchanged; `test/test_groups.toml` is untouched. The dispatch was verified to fire exactly the same group bodies as the old `if GROUP == ...` ladder. Verified end-to-end on Julia 1.11: `GROUP=QA` passes 11/11 via `Pkg.test()`.

### Deps
Add `SciMLTesting` to root `[extras]`/`[targets].test`/`[compat]`. `Pkg` is retained (`activate_gpu_env` still uses it). `SafeTestsets`/`Test` were already present.

Ignore until reviewed by @ChrisRackauckas.


isaacsas:
Please hold,off on this till the JumpProcesses 10 changes are merged. This will,be a mess to reconcile with them.

PR #649, opened 2026-09-01
Add a GPU kernel for SimpleExplicitTauLeaping

sivasathyaseeelan:
Runs explicit tau-leaping with one trajectory per GPU thread through `EnsembleGPUKernel`, alongside the existing `SimpleTauLeaping` and `SSAStepper` kernels.

```julia
using JumpProcesses, KernelAbstractions, Adapt, CUDA, Statistics

maj = MassActionJump([0.1 / 1000.0, 0.01],
                     [[1 => 1, 2 => 1], [2 => 1]],
                     [[1 => -1, 2 => 1], [2 => -1, 3 => 1]])
prob = DiscreteProblem([999.0, 10.0, 0.0], (0.0, 250.0))
jump_prob = JumpProblem(prob, PureLeaping(), maj)

sol = solve(EnsembleProblem(jump_prob), SimpleExplicitTauLeaping(),
            EnsembleGPUKernel(CUDABackend()); trajectories = 100_000, saveat = 10.0)

mean(sol.u[i].u[end][3] for i in 1:100_000)
```

## Scope

Mass action jumps only. A `RegularJump` carries arbitrary Julia functions for its rate and its `c`, which a device kernel cannot evaluate, so those problems are rejected with an error that says what to do instead. Callbacks and a `prob_func` are likewise rejected rather than silently ignored.

## How it works

* **Leap selection on the device.** `gpu_compute_gi` mirrors `compute_gi` exactly, and the per-species highest order and largest stoichiometry it needs are precomputed once on the host. `gpu_compute_tau` differs from `compute_tau` only in how the moments are gathered: instead of summing `mu_i` and `sigma_i^2` over a dense stoichiometry matrix, each propensity is evaluated once and scattered through the net stoichiometry, which visits only the nonzeros and needs no propensity cache in global memory. The two agree because the entries the host walks past are zero.
* **Saving.** The leap size adapts to the state, so the number of steps a trajectory takes is not known up front and the solution is sampled onto a fixed `saveat` grid, as the SSA kernel does. `tau` is shortened to land exactly on the next grid point when a step would overshoot it, matching the host solver, so every saved value is a state the trajectory actually occupied rather than an interpolation.
* **Rejection.** A leap that would drive a population negative is rejected and `tau` halved, with the bound carried across iterations so that a retry cannot reproduce the rejected size.
* **No scratch memory.** The state lives in registers as an `SVector`, and the propensities are evaluated twice per step, once to size the leap and once to draw the Poisson counts, so the kernel allocates no per-thread buffers in global memory. Counts come from the backend's device RNG, the same generator the `SimpleTauLeaping` kernel draws from.
* **Shared with the SSA kernel.** The stoichiometry encoding, propensity evaluation, save-grid construction and input validation are reused rather than duplicated. Two of those helpers lose their `ssa` prefix now that both kernels use them.

## Fixes `compute_hor` for zero order reactions

`compute_hor` took its element type from the first entry of the first reaction, `eltype(first(first(reactant_stoch)))`. A zero order reaction has an empty reactant stoichiometry, so that raised `BoundsError: attempt to access 0-element Vector{Pair{Int64, Int64}} at index [1]`. This affected the existing serial `SimpleExplicitTauLeaping` and `SimpleImplicitTauLeaping` as much as the kernel — neither could run a model with a constant-rate influx. The type is now taken from the container type instead, and a regression test covers both serial algorithms.

## Testing

`test/explicit_tau_kernel_tests.jl` holds the assertions and is run twice, against `CUDABackend()` in the `CUDA` group and against KernelAbstractions' `CPU()` in `InterfaceII`, so the kernel is covered by the ordinary test suite as well as by the GPU run.

The reference is the serial `SimpleExplicitTauLeaping` rather than the exact SSA, because tau-leaping carries a discretization bias that the kernel is expected to reproduce. One test pins that bias down instead of ignoring it: it checks the error against the analytic mean shrinks monotonically as `epsilon` tightens (locally 128.5 → 134.0 → 134.8 against an exact 135.3 for `epsilon` of 0.05, 0.01, 0.002). Coverage includes linear death, zero order, SIR, second and third order propensities, a conservation law, extinction, `saveat`/`save_start`/`save_end` handling, and each rejected input.

Verification notes: the CPU backend passes 80/80 locally. The new device arithmetic was also checked on real GPU hardware through Metal, where `gpu_compute_tau` reproduces the host `compute_tau` bit for bit across every `g_i` branch (orders 1 to 3, stoichiometries 1 to 3, and the zero order case). The full kernel cannot run on Metal because `pois_rand` does not compile for that backend — this is pre-existing and not specific to this PR, as master's `SimpleTauLeaping` kernel bus-errors there identically — so end-to-end GPU coverage comes from the CUDA runner.

## Checklist

### DIFF
--- a/ext/explicit_tau_leaping.jl
+++ b/ext/explicit_tau_leaping.jl
@@ -11,6 +11,8 @@ host, so the device only evaluates the closed form.
 """
 @inline function gpu_compute_gi(u, max_hor, max_stoich, i, ::Type{RT}) where {RT}
     onert = one(RT)
+    two = 2 * onert
+    three = 3 * onert
 
     @inbounds begin
         hor = max_hor[i]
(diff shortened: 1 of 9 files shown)


ext/explicit_tau_leaping.jl
```diff
@@ -0,0 +1,321 @@
...
+        stoich == 1 && return 3 * onert
+        if stoich == 2
+            # 2S_i + S_k -> products; fall back to 3 when x_i <= 1
+            return x > onert ?
+                   (3 * onert / 2) * (2 * onert + onert / (x - onert)) : 3 * onert
+        end
+        # 3S_i -> products; fall back to 3 when x_i <= 2
+        return x > 2 * onert ?
+               3 * onert + onert / (x - onert) + 2 * onert / (x - 2 * onert) : 3 * onert
+    end
```
ChrisRackauckas: `ifelse` instead of branching? Might not matter on GPU but it's probably faster anyways. There's CSE that can be done here?

sivasathyaseeelan: Done both. `gpu_compute_gi` is branch-free now, and the repeated `1 / (x_i - 1)` is computed once — the order 3, stoichiometry 2 case is exactly 3/2 of the order 2 form, so it reuses that instead of recomputing.

Verified bit-identical to the host `compute_gi` across all 600 `max_hor` / `max_stoich` / population combinations, including the `x = 1` and `x = 2` fallback boundaries, in both precisions. About 15% faster on an RTX 5080 over 200k trajectories.

PR #652, opened 2026-09-05
Preserve native mass-action data for tau-leaping solvers

ChrisRackauckas-Claude:
Preserve `MassActionJump` in `JumpProblem` and let tau-leaping solvers consume its stored rates and stoichiometry. `SimpleTauLeaping` now uses native mass-action data on CPU and the flattened mass-action representation in GPU kernels; general `RegularJump` input remains supported. This replaces the constructor-level conversion raised in review.

Add documented public propensity, stoichiometry-product, and fused drift operations for downstream solvers, with a minor version bump to 9.33. The companion StochasticDiffEq implementation uses these operations directly: https://github.com/SciML/OrdinaryDiffEq.jl/pull/4523 . This prerequisite can be released independently; its executable examples use JumpProcesses solvers. The full cross-package guide is https://github.com/SciML/JumpProcesses.jl/pull/650 .

## Verification

Julia 1.12.6 with dependencies developed in a local environment. Existing test entry points were used because `Pkg.test` encountered a relative-`[sources]` sandbox resolution error in the local monorepo dependency setup.

```sh
GROUP=InterfaceI julia --project=native-validation massaction-leaping/test/runtests.jl
GROUP=InterfaceII julia --project=native-validation massaction-leaping/test/runtests.jl
GROUP=QA julia --project=native-validation massaction-leaping/test/runtests.jl
julia --project=massaction-leaping/docs massaction-leaping/docs/make.jl
```

```text
InterfaceI: 940 passing checks across 21 testsets
InterfaceII: 1597 passing checks across 21 testsets
QA Tests | 20 20
[ Info: HTMLWriter: rendering HTML pages.
[ Info: Automatic `version="9.33.0"` for inventory from ../Project.toml
```

The new mass-action checks pass 29 assertions. KernelAbstractions CPU-backend checks include 12 fixed-step mass-action, 67 SSA, and 80 adaptive explicit assertions.

Runic on changed Julia lines, `typos` over changed files, and `git diff --check` passed. GPU kernels ran through KernelAbstractions' CPU backend; no physical GPU execution was performed. Adaptive implicit, trapezoidal, and switching methods have no GPU implementation.

The constructor regression test discriminates against the previous adapter implementation (https://github.com/SciML/JumpProcesses.jl/commit/663b28171df580638a7512a459c8e0ac51909305):

```text
Before: Test Failed: prob.regular_jump === nothing
After:  same assertion passes (exit 0)
```

The public operation tests cover scalar and vector reaction storage, zero populations, higher-order reactions, and derivative propagation. A scalar reaction test produced six errors before the scalar-data fix and passes all six checks afterward. Native/regular fixed-seed trajectory comparisons and unsupported mixed-jump rejection are included.

Ignore this PR until reviewed by @ChrisRackauckas.

🤖 Generated with Codex CLI 0.153.4 (model: gpt-6-astra; local session ID: 01a070e6-21f2-7dd3-bf52-2ddaf108ac53).

### DIFF
--- a/src/problem.jl
+++ b/src/problem.jl
@@ -350,13 +350,18 @@ function JumpProblem(prob, aggregator::PureLeaping, jumps::JumpSet;
     crjs = jumps.constant_jumps
     vrjs = jumps.variable_jumps
     
-    iip = isinplace_jump(prob, jumps.regular_jump)
+    rj = jumps.regular_jump
+    if rj === nothing && prob isa DiscreteProblem && get_num_majumps(maj) > 0
+        rj = massaction_regular_jump(maj)
+    end
+    iip = isinplace_jump(prob, rj)
     solkwargs = tstops === nothing ? make_kwarg(; callback) : make_kwarg(; callback, tstops)
 
     JumpProblem{iip, typeof(prob), typeof(aggregator), typeof(jump_cbs),
-        typeof(disc_agg), typeof(crjs), typeof(vrjs), typeof(jumps.regular_jump),
+        typeof(disc_agg), typeof(crjs), typeof(vrjs), typeof(rj),
         typeof(maj), typeof(rng), typeof(solkwargs)}(prob, aggregator, disc_agg,
-        jump_cbs, crjs, vrjs, jumps.regular_jump, maj, rng, solkwargs)
+        jump_cbs, crjs, vrjs, rj, maj, rng, solkwargs
+    )
 end
 
 aggregator(jp::JumpProblem{iip, P, A}) where {iip, P, A} = A
--- a/src/problem.jl
+++ b/src/problem.jl
@@ -351,9 +351,6 @@ function JumpProblem(prob, aggregator::PureLeaping, jumps::JumpSet;
     vrjs = jumps.variable_jumps
     
     rj = jumps.regular_jump
-    if rj === nothing && prob isa DiscreteProblem && get_num_majumps(maj) > 0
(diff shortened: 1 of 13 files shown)


src/problem.jl
isaacsas, line 1: I don’t think this is the right level to make such conversions. For some leaping algorithms it would be better to keep the mass action jumps separate, but by doing the conversion here we take away the decision on how to setup the reaction representation from the leaping method.

isaacsas (changes requested):

src/problem.jl
ChrisRackauckas: I agree, I was trying to see what would happen by normalizing the form, but this would probably take a hit. We probably instead need to update all of the StochasticDiffEq methods to be MassActionJumps-aware.

ChrisRackauckas-Claude: Replaced the constructor conversion: `JumpProblem` retains `MassActionJump`, and CPU/GPU fixed tau leaping consumes its reaction data directly. The StochasticDiffEq companion updates all four leaping methods to use stored constants and stoichiometry, including fused implicit drift: https://github.com/SciML/OrdinaryDiffEq.jl/pull/4523 . General `RegularJump` remains an additional supported input. Benchmarks in that PR show lower compilation cost in the measured cases; warm runtime is mixed.

🤖 Codex CLI 0.153.4; model: gpt-6-astra; local session ID: 01a070e6-21f2-7dd3-bf52-2ddaf108ac53.

Issue #620, opened 2026-07-31
Monotonicity assumption in RSSA/RSSACR


nifets:
I think RSSA as currently implemented is only valid when the rate function is monotonic. I didn't see this documented anywhere in the package.

isaacsas:
It is documented internally, https://github.com/SciML/JumpProcesses.jl/blob/5dc1d35a6277c4e2d7eaf4727faa2378ea87e7d2/src/aggregators/rssa.jl#L3 
but should indeed be in the public docs too.

lmshk:
If the rate function is monotonic in each argument individually, could setting `; bracket_data = BracketData(...)` on `JumpProblem` construction with negative `fluctrate`s for species with opposite monotonicity invert the bracket appropriately? Function [`delta_bracket`](https://github.com/SciML/JumpProcesses.jl/blob/5dc1d35a6277c4e2d7eaf4727faa2378ea87e7d2/src/aggregators/bracketing.jl#L64), called via `get_spec_brackets`, would simply return swapped bounds and [update_dependent_rates!](https://github.com/SciML/JumpProcesses.jl/blob/5dc1d35a6277c4e2d7eaf4727faa2378ea87e7d2/src/aggregators/rssacr.jl#L175) would then fill `ulow`/`uhigh` accordingly so that `get_jump_brackets` would see them in the correct configuration. [This condition](https://github.com/SciML/JumpProcesses.jl/blob/5dc1d35a6277c4e2d7eaf4727faa2378ea87e7d2/src/aggregators/rssacr.jl#L173) would need to be changed ([twice](https://github.com/SciML/JumpProcesses.jl/blob/5dc1d35a6277c4e2d7eaf4727faa2378ea87e7d2/src/aggregators/rssacr.jl#L199), and [twice](https://github.com/SciML/JumpProcesses.jl/blob/5dc1d35a6277c4e2d7eaf4727faa2378ea87e7d2/src/aggregators/rssa.jl#L155) [more](https://github.com/SciML/JumpProcesses.jl/blob/5dc1d35a6277c4e2d7eaf4727faa2378ea87e7d2/src/aggregators/rssa.jl#L179)) to handle the inversion.

isaacsas:
The method always reorders the bounds from the propensity function to be ordered. It shouldn’t need any reordering of species bounds. We probably don’t have a test for this, but it should work I think with a decreasing rate function.

lmshk:
[...]

PR #653, opened 2026-09-06
Add optional rate bounds to ConstantRateJump

nifets:
## Summary

Added `lrate(ulow, uhigh, p, t)` and `urate(ulow, uhigh, p, t)` to `ConstantRateJump`, allowing users to specify bounds on the rate with respect to state for the interval `[ulow,uhigh]`, which is used for bracketing aggregators `RSSA` and `RSSACR`.

This addresses #620. Previously, `get_cjump_brackets` resolved these bounds by evaluating the rate at the corners `ulow`, `uhigh` and optionally reordering the result. This is only valid when the rate is monotonic in the same direction in every species. Exposing `lrate` and `urate` means bracketing aggregators can now use crj's with arbitrary rate functions as long as the user specified bounds are correct. Behaviour remains unchanged if bounds are not supplied (but I think we should only preserve the fallback bound calculation for increasing rate functions and remove the reordering flow).

It turns out that this potential correctness issue is actually hit in this repository: `test/allocations.jl` uses `u = [1000, 10]` and `r1 = (η/K)(K - u[1] - u[2]) * u[1]`, which increases in `u[1]` but decreases in `u[2]` locally around `u`. The default fallback bounds are `(818.46, 977.46)`, whereas the rate actually ranges over `(817.74, 978.34)`. It's a small effect here but it can get much more severe at low copy numbers (e.g. for `u[1]/(1+u[2])` at `u=[5,1]`, the automatic bounds are `(1.0, 1.5)` vs the real ones `(0.17, 9.0)`). I didn't touch the test because it measures allocation rather than correctness, but ideally we would specify rate bounds for this example.

## Checklist

## Notes

`ConstantRateJump` gained two type parameters, so the two-parameter form no longer matches, which is technically breaking public API. However, nothing seems to use the type annotation in any downstream dependencies.

The bracketing requirement is now documented for the `RSSA` and `RSSACR` aggregators.

All tests passed locally.

### DIFF
--- a/docs/src/jump_types.md
+++ b/docs/src/jump_types.md
@@ -75,7 +75,7 @@ their own special time integrators.
 The constructor for a [`ConstantRateJump`](@ref) is:
 
 ```julia
-ConstantRateJump(rate, affect!)
+ConstantRateJump(rate, affect!; bounds=nothing)
 ```
 
   - `rate(u, p, t)` is a function which calculates the rate given the current
@@ -84,10 +84,11 @@ ConstantRateJump(rate, affect!)
     interface. It encodes how the state should change due to *one* occurrence of
     the jump.
 
-For aggregators that bracket rates over state intervals, such as `RSSA` and `RSSACR`, one
-can also supply `lrate` and `urate` rate bound functions. If not supplied, these bounds
-will be computed automatically by evaluating `rate` at the interval endpoints (which is
-only correct for rate functions that are increasing, such as mass action rates).
+For aggregators that bracket rates, such as `RSSA` and `RSSACR`, one can also supply a
+`bounds(ulow, uhigh, p, t)` function returning a [`RateBounds`](@ref) that bounds `rate`
+over the state bracket `[ulow, uhigh]`. If not supplied, the bounds are computed
+automatically by evaluating `rate` at the endpoints of the bracket (which is only correct
+for rate functions that are monotonic in the state, such as mass action rates).
 
 #### Defining a Mass Action Jump
 
--- a/docs/src/jump_types.md
+++ b/docs/src/jump_types.md
@@ -85,7 +85,7 @@ ConstantRateJump(rate, affect!; bounds=nothing)
     the jump.
 
 For aggregators that bracket rates, such as `RSSA` and `RSSACR`, one can also supply a
-`bounds(ulow, uhigh, p, t)` function returning a [`RateBounds`](@ref) that bounds `rate`
+`bounds(ulow, uhigh, u, p, t)` function returning a [`RateBounds`](@ref) that bounds `rate`
 over the state bracket `[ulow, uhigh]`. If not supplied, the bounds are computed
 automatically by evaluating `rate` at the endpoints of the bracket (which is only correct
 for rate functions that are monotonic in the state, such as mass action rates).
--- a/docs/src/jump_types.md
+++ b/docs/src/jump_types.md
@@ -84,6 +84,11 @@ ConstantRateJump(rate, affect!)
     interface. It encodes how the state should change due to *one* occurrence of
     the jump.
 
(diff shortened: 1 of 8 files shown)


src/jumps.jl
```diff
@@ -1,3 +1,83 @@
...
+Computed bounds on a jump's rate for a fixed state bracket `[ulow, uhigh]` at time `t`.
+
+## Fields
+
+$(FIELDS)
+
+## Notes
+
+  - The rates must satisfy `lrate <= rate(u, p, s) <= urate` for every `u` with `ulow .<= u .<= uhigh` and `s` in `[t, t + rateinterval]`.
+  - At least one of `lrate` or `urate` must be given. An omitted bound defaults to the trivially valid one: `zero` below and `typemax` above.
```
isaacsas, line 14: Are there any cases where `lrate` only is useful? Just having upper bounds is useful, but maybe we should error for now with just `lrate` being passed?

isaacsas: Nevermind, misunderstood the intention of `RateBounds`...

docs/src/jump_types.md
```diff
@@ -294,6 +300,8 @@ aggregator requires various types of dependency graphs, see the next section):
   - `RSSA`: The Rejection SSA (RSSA) method of Thanh *et al.* [^5][^6]. With `RSSACR`,
     for very large reaction networks, it often offers the best performance of all
     methods. [Dependency graph required](@ref Jump-Aggregators-Requiring-Dependency-Graphs).
+    For `ConstantRateJump`s, `RSSA` and `RSSACR` require rate bounds, see
+    [`ConstantRateJump`](@ref).
```
isaacsas: This should probably repeat that if one is not provided one is auto-generated but assumes monotonicity.

isaacsas (changes requested):
OK, this looks quite good now, thanks! Sorry it took me a bit of time to review. I left one minor review comment, and below are a few comments that Fable gave me that I agree with:

1. One-sided RateBounds through the joint bounds slot is a footgun. Nothing stops bounds = (...) -> RateBounds(lrate = …). Then urate = typemax(lrate) = Inf, sum_rate = Inf, and RSSA's rejection loop never accepts (rand()*Inf > rate always), so generate_jumps! spins forever; RSSACR dies in exponent(Inf) in the priority table. Since set_bracketing! is init-only and update_dependent_rates! is the hot path, the cheap fix is a one-time check in set_bracketing!: 0 ≤ lrate ≤ urate < Inf, plus lrate ≤ rate(u,p,t) ≤ urate at the initial state. That second check also catches sign/ordering mistakes in user bounds and grossly wrong fallback brackets, at zero steady-state cost.
2. ConstantRateJump(rate, affect!; bounds = RateBoundFunctions(...)) double-wraps, giving c.bounds.bounds::RateBoundFunctions, which passes hasbounds and then throws a MethodError on the first bracket call. Add a bounds::RateBoundFunctions dispatch in the kwarg constructor.
3. Test gap. Only bracket containment is tested. There's no end-to-end check that RSSA/RSSACR with user bounds on a non-monotone rate reproduce Direct's statistics. A small mean-comparison test in the style of linearreaction_test.jl with a u[1]/(1+u[2]) rate would close that.

Finally, could you update the HISTORY.md to summarize this change for users? 

After this I think we are good to merge.

isaacsas (approved):

isaacsas:
Looks good, thanks!

isaacsas:
@ChrisRackauckas is adding an extra type parameter to ConstantRateJumps considered breaking? I don't think it should impact most people (such as MTK or Catalyst).

ChrisRackauckas:
It's not breaking. Julia is specifically designed for dispatches leaving off trailing type parameters to automatically match Any so that this case is not breaking.

